PR #22, opened 2015-12-24
Use proper UTF-8 instead of CString.

eddyb:
Also fixes a dangling pointer bug, as `lower_str_unwrap` was returning a pointer into the `CString` allocation it had created and destroyed.

Tested on x64 Linux, Hello World works at least.

### DIFF
--- a/src/internal/value.rs
+++ b/src/internal/value.rs
@@ -1,9 +1,7 @@
 use std;
 use std::mem;
 use std::os::raw::c_void;
-use std::ffi::{CString, CStr};
-use neon_sys::raw;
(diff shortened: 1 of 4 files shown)


src/internal/value.rs
```diff
@@ -1,9 +1,7 @@
 use std;
 use std::mem;
 use std::os::raw::c_void;
-use std::ffi::{CString, CStr};
-use neon_sys::raw;
-use neon_sys::{NeonSys_NewObject, NeonSys_NewUndefined, NeonSys_NewNull, NeonSys_NewBoolean, NeonSys_NewInteger, NeonSys_NewString, NeonSys_NewNumber, NeonSys_NewArray, NeonSys_Array_Length, NeonSys_String_Utf8Length, NeonSys_String_Data, NeonSys_Value_ToString, NeonSys_Object_GetOwnPropertyNames, NeonSys_Object_Get_Index, NeonSys_Object_Set_Index, NeonSys_Object_Get, NeonSys_Object_Get_Bytes, NeonSys_Object_Set_Bytes, NeonSys_Object_Set, NeonSys_NewFunction, NeonSys_FunctionKernel, NeonSys_Call_GetIsolate, NeonSys_IsUndefined, NeonSys_IsNull, NeonSys_IsInteger, NeonSys_IsNumber, NeonSys_IsString, NeonSys_IsBoolean, NeonSys_IsObject, NeonSys_IsArray, NeonSys_IsFunction, NeonSys_TagOf, Tag};
+use neon_sys::*;
```
steveklabnik: never! Down with globs! :neutral_face:

eddyb: Go rustfmt the original line, my enter key would break if I formatted it by hand.

diorahman:
Nice!

Issue #33, opened 2016-01-05
eliminate dependence on node-gyp


dherman:
In the `neon` crate we can eliminate the dependence on node-gyp by making a macro analogous to the CPP `NODE_MODULE` macro in Node. This would be a prerequisite to fulfilling the neon-cli RFC at rustbridge/neon-cli#21. In PR #29 @eddyb showed how this could work.

(The `node-sys` crate would continue to use node-gyp, since it requires C/C++ wrappers.)

PR #35, opened 2016-01-05
Pure rust module registration (for #33)

dherman:
Based on @eddyb's original PR #29.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,11 +10,65 @@ pub mod value;
 pub mod error;
 pub mod buffer;
 
-use internal::vm::{Module, Throw};
-use internal::mem::Handle;
-use internal::value::SomeObject;
+/// The module version for Node.js 4.x is 46.
+// TODO detect this based on what we're compiling for.
+pub const NODE_MODULE_VERSION: i32 = 46;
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -10,11 +10,65 @@ pub mod value;
...
+// The module version for Node.js 4.x is 46.
+// FIXME: detect this based on what we're compiling for.
+pub const NODE_MODULE_VERSION: i32 = 46;
 
-#[no_mangle]
-pub extern "C" fn neon_init(module: Handle<SomeObject>, init: fn(Module) -> Result<(), Throw>) {
-    Module::initialize(module, init);
+#[macro_export]
+macro_rules! onload {
+    ($module:ident, $init:block) => {
```
eddyb: I prefer `:` instead of `,` but that's just nitpicking.

dherman: > I prefer `:` instead of `,`

Any particular reason why? `:` reads like a type annotation to me.

eddyb: No particular reason, just that it looks marginally better IMO.
Maybe `module => {...}`? At least it has some syntactical sense then.

PR #39, opened 2016-01-07
closes #36: Value, Object, Any renamings

### LINKED ISSUE #36: Value, Object, Any renamings
Looking at the API docs makes it clear that the traits *really* want to have the names Value and Object. And I came up with the "SomeObject" nomenclature for the root of the Object type hierarchy. So there should be a consistent nomenclature for root types in the type hierarchy, like SomeValue and SomeObject. I'm not in love with Some- as the prefix, so maybe we can come up with something better.

Also, neon::buffer should get moved under neon::value. Should we mirror the type hierarchy in a module hierarchy? Like neon::value::object::{Array, Object} etc, and neon::value::object::typed_array::Uint8Array etc?

dherman:


dherman:
I'm going to go ahead and merge this PR for now, so that we can get some experience with the feel of it. If it doesn't feel good we can iterate on it.

PR #79, opened 2016-04-26
Tests

mfpiccolo:
This gives Neon three things. 

1. Acceptance tests in Javascript
2. Compile tests
3. A place for future Rust unit tests  (by adding `cd tests/native && env NEON_NODE_ABI=`node -e 'console.log(process.versions.modules)'` cargo test` to `bin/test` and then adding Rust tests inside the native crate)

From root run `bin/test` to run the suite.


tests/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "tests",
+  "version": "0.1.0",
+  "description": "Acceptance test suite for Neon",
+  "main": "lib/index.js",
+  "author": "n <n>",
```
dherman: Can you put "The Neon Community" for this one too?

.gitignore
```diff
@@ -4,4 +4,4 @@ Cargo.lock
 **/node_modules
 **/.DS_Store
 **/build
-
```
dherman: I think we should add `**/index.node` too, so that tests/native/index.node doesn't end up in source control.

dherman:
This is so great! If you wouldn't mind fixing the couple minor nits then I'll merge right away.

mfpiccolo:
Fixed.  I removed index.node with a commit but I can squash and force if you prefer that.

dherman:
I figured out why the Travis tests are failing: they upgraded to Node.js v6, which requires an upgrade of nan and node-gyp. Can you change the `crates/neon-sys/package.json` to use `"nan": "^2.3.2"` and `"node-gyp": "^3.3.1"`?

Also, we should start specifying the exact versions of Node.js we support in `.travis.yml`. I think for now we should support 4.x and later. It's up to you if you want to add that in this PR or we can add it afterwards.

mfpiccolo:
Pushed up the neon-sys dep bumps.  I can mess with travis config later on today or tonight if you don't get to it.

PR #81, opened 2016-04-27
Fix compiler warnings on function casts

zmanian:
Fix compiler warnings from rust-lang/rust#19925

### DIFF
--- a/src/internal/vm.rs
+++ b/src/internal/vm.rs
@@ -62,7 +62,7 @@ impl IsolateInternal for Isolate {
             let b: Box<ClassMap> = Box::new(ClassMap::new());
             let raw = Box::into_raw(b);
             ptr = unsafe { mem::transmute(raw) };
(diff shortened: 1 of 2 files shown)


zmanian:
I'm gonna work on figure out why the CI is failing. May take me a few days to find the time.

zmanian:
Hmm...I'm having trouble reproducing this locally.

src/internal/vm.rs
```diff
@@ -256,7 +256,7 @@ pub trait Kernel<T: Clone + Copy + Sized>: Sized {
 
     fn pair(self) -> (*mut c_void, *mut c_void) {
         unsafe {
-            (mem::transmute(Self::callback), self.as_raw())
+            (mem::transmute(Self::callback as usize), self.as_raw())
```
eddyb: Why `usize` here?

Issue #78, opened 2016-04-26
get rid of Integer type


dherman:
It seems like `v8::Integer` isn't clearly defined (there's no documented integer range it represents, and the Cast operation seems broken), and since it doesn't correspond to the actual integer-tagging optimizations within the engine (the "smi" tag for small ints is different on different architectures and purposefully not exposed).

So I think this means there's no value in having an Integer type. Instead, there should just be overloading conveniences for coercing JS numbers to/from Rust integers.

Issue #76, opened 2016-04-24
Code coverage for initial test suite


mfpiccolo:
# Code Coverage

This issue is intended to identify and track tests that will provide a solid level of code coverage for the inital test suite.  

---
### Code Coverage
- [x] The list below describes satisfactory coverage for the intial suite

---
### Types
- [ ] successful upcasts between various types in Rust
- [ ] successful downcasts between various types in Rust
- [ ] unsuccessful downcasts between various types in Rust
### Classes
- [ ] call methods on objects of the wrong type
### Control Flow
- [ ] propagate exceptions from JS through Rust back to JS
- [ ] trigger exceptions from Rust
### JsString
- [x] create a JsString in Rust
- [ ] exotic code points properly round-tripping between Rust and JS
### JsNumber
- [x] create a JsNumber in Rust
- [ ] large integers > int32.max properly round-tripping between Rust and JS
- [ ] negative integers properly round-tripping between Rust and JS
### JsArray
- [x] create a JsArray in Rust
- [x] populate JsArray in Rust 
### JsObject
- [x] create a JsObject in Rust
- [x] populate JsObject in Rust 
### Typed Arrays
- [ ] create a typed array in Rust
- [ ] get/set data from a typed array in Rust
### Compile Test Crate
- [x] Setup and configure the crate that will handle testing compile failures
- [ ] scope lifetime errors
### CI
- [x] Setup Travis or other CI system and integrate with Github

mfpiccolo:
@dherman This is obviously not the whole test suite yet but the idea is that we can continue to add to this issue until you feel that we have covered enough for the initial test suite setup.  Then we can mark the first checkbox off to let us know what to aim for.  Then we can start knocking off the tests.

dherman:
That's probably a good initial set. It's almost impossible to have any real notion of coverage being "complete" for a project like this, but it's at least a good breadth-first start. I'm so excited to grow the test suite over time!

mfpiccolo:
Yeah agree that "complete" is a bit ambitious.  Thanks for adding to this!

kjvalencik:
Neon has pretty good coverage in the test suite since the Node-API backend.

PR #84, opened 2016-05-05
Js number tests

mfpiccolo:
Switch over to testing JsNumber because of #78

### DIFF
--- a/tests/lib/numbers.js
+++ b/tests/lib/numbers.js
@@ -2,23 +2,50 @@ var addon = require('../native');
 var assert = require('chai').assert;
 
 describe('JsNumber', function() {
-  it('should return a JsNumber built in Rust', function () {
+  it('return a JsNumber built in Rust', function () {
     assert.equal(9000, addon.return_js_number());
   });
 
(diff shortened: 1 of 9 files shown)


mfpiccolo:
This should cover the JsNumber section of #76.  Let me know if there is any other JsNumber specific tests we should have for 76.

tests/lib/numbers.js
```diff
@@ -0,0 +1,51 @@
+var addon = require('../native');
+var assert = require('chai').assert;
+
+describe('JsNumber', function() {
+  it('return a JsNumber built in Rust', function () {
+    assert.equal(9000, addon.return_js_number());
```
dherman: I think the API is `equal(actual, expected)`:

http://chaijs.com/api/assert/#method_equal

dherman:
Awesome -- just one minor issue, the expected/actual ordering is backwards in the chai tests.

tests/lib/numbers.js
mfpiccolo: Oh I will get that switched up.  👍

PR #96, opened 2016-07-01
Fix spurious improper_ctypes warning

freebroccolo:
This PR fixes seemingly spurious warnings about improper ctypes that show up when neon macros are used. Maybe there is a better way to fix this but I couldn't figure out how.


dherman:
:tada: Thank you!!

PR #107, opened 2016-08-28
Fix compile on nightly, remove warnings

winding-lines:
https://github.com/rust-lang/rust/issues/19925


dherman:
This was partially a dup of #108, which I already merged, so I went ahead and fixed the spurious imports in 520d8abeda768c5abdaca2d538e285348f2c389c -- but thanks for the PR!

PR #109, opened 2016-09-05
Take module version from node headers.

shawedwa:
If module abi version isn't specified in env, take it from the included node headers.

### DIFF
--- a/crates/neon-sys/src/module.rs
+++ b/crates/neon-sys/src/module.rs
@@ -6,4 +6,7 @@ extern "system" {
     #[link_name = "NeonSys_Module_ExecKernel"]
     pub fn exec_kernel(kernel: *mut c_void, callback: extern fn(*mut c_void, *mut c_void, *mut c_void), exports: Local, scope: *mut c_void);
 
+    #[link_name = "NeonSys_Module_GetVersion"]
+    pub fn get_version() -> i32;
+
 }
--- a/crates/neon-sys/src/neon.cc
+++ b/crates/neon-sys/src/neon.cc
@@ -250,6 +250,10 @@ extern "C" void NeonSys_Module_ExecKernel(void *kernel, NeonSys_ModuleScopeCallb
   callback(kernel, exports, scope);
 }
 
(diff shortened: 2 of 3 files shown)


crates/neon-sys/src/module.rs
```diff
@@ -6,4 +6,7 @@ extern "system" {
     #[link_name = "NeonSys_Module_ExecKernel"]
     pub fn exec_kernel(kernel: *mut c_void, callback: extern fn(*mut c_void, *mut c_void, *mut c_void), exports: Local, scope: *mut c_void);
 
+    #[link_name = "NeonSys_Module_GetVersion"]
+    pub fn get_version() -> u32;
```
shawedwa: Oops. Wrong type. Should be i32

shawedwa:
Sorry - new to Rust. Worked out the macro_internal stuff, so it all works properly now.

EdShaw:
Helps if I use the right account too. ><

KenanSulayman:
Can you also touch the invocation to not pass the environment variable anymore? This will make the `NEON_NODE_ABI` an optional override.

EdShaw:
Yup:
https://github.com/rustbridge/neon-cli/pull/31

EdShaw:
Superseded by: https://github.com/rustbridge/neon/pull/123

dherman:
r+

This is awesome, and I'm sorry for taking so long to review! I'm working with @jedireza to try to keep on top of PRs more regularly.

Because this needs to sync with neon-bindings/neon-cli#31 it's worth merging independently of #123, and I think some aspects of #123 won't be able to land as-is anyway. So I think we should merge this one.

As far as I can tell, this improvement has the same behavior as the way it originally works, but is simply less brittle. So it seems like a clear win.

Thanks so much!

dherman:
(I think @jedireza wanted to just test this out in a few versions of Node before merging, so I'll wait to merge till he's ready.)

jedireza:
+r

I tested using the latest node `4`, `6`, `7` with mac and linux.

dherman:
(Next I'll merge the commit that eliminates the env variable entirely.)

PR #121, opened 2016-10-19
Expose JS strings' UCS2 representation to rust

maxbrunsfeld:
Wow, this is an awesome library.

I'm working on a module that performs string manipulation in rust using string indices that are provided by javascript. I'd like to avoid transcoding the JS strings back and forth from UTF8 and converting UCS2 indices to UTF8 indices.

This adds the following methods on `JsString` class, analogous to the existing utf8-based methods:
- `new_from_ucs2(scope, data: &[u16])` (analogous to `new`)
- `ucs2_value() -> Vec<u16>` (analogous to `value`)
- `ucs2_size() -> isize` (analogous to `size`)

I added a test in a similar style to your existing tests.

I wasn't totally sure about the naming; V8 doesn't refer to strings internal encoding in any consistent way, so `ucs2` seemed like the most precise qualifier to use.

### DIFF
--- a/crates/neon-sys/src/neon.cc
+++ b/crates/neon-sys/src/neon.cc
@@ -177,6 +177,19 @@ extern "C" size_t NeonSys_String_Data(char *out, size_t len, v8::Local<v8::Value
   return Nan::DecodeWrite(out, len, str, Nan::UTF8);
 }
 
+extern "C" bool NeonSys_String_NewFromUCS2(v8::Local<v8::String> *out, v8::Isolate *isolate, const uint16_t *data, int32_t len) {
+  Nan::MaybeLocal<v8::String> maybe = v8::String::NewFromTwoByte(isolate, data, v8::NewStringType::kNormal, len);
+  return maybe.ToLocal(out);
+}
+
+extern "C" int32_t NeonSys_String_UCS2Length(v8::Local<v8::String> str) {
+  return str->Length();
+}
+
+extern "C" size_t NeonSys_String_UCS2Data(uint16_t *out, size_t len, v8::Local<v8::String> str) {
+  return str->Write(out, 0, -1);
+}
+
 extern "C" bool NeonSys_Convert_ToString(v8::Local<v8::String> *out, v8::Local<v8::Value> value) {
   Nan::MaybeLocal<v8::String> maybe = Nan::To<v8::String>(value);
   return maybe.ToLocal(out);
--- a/src/internal/js/mod.rs
+++ b/src/internal/js/mod.rs
@@ -296,6 +296,12 @@ impl JsString {
         }
     }
 
+    pub fn ucs2_size(self) -> isize {
+        unsafe {
+            neon_sys::string::ucs2_len(self.to_raw())
+        }
+    }
+
     pub fn value(self) -> String {
         unsafe {
             let capacity = neon_sys::string::utf8_len(self.to_raw());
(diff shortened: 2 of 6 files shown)


src/internal/js/mod.rs
```diff
@@ -307,10 +313,26 @@ impl JsString {
         }
     }
 
+    pub fn ucs2_value(self) -> Vec<u16> {
```
RReverser: I think it might be worth to expose slice directly from pointer & length provided by V8 - that way you need only one method instead of two, and can avoid unnecessary allocations if you just need to read the data.

RReverser: More specifically, I'm talking about using https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

crates/neon-sys/src/neon.cc
```diff
@@ -177,6 +177,19 @@ extern "C" size_t NeonSys_String_Data(char *out, size_t len, v8::Local<v8::Value
...
+  Nan::MaybeLocal<v8::String> maybe = v8::String::NewFromTwoByte(isolate, data, v8::NewStringType::kNormal, len);
+  return maybe.ToLocal(out);
+}
+
+extern "C" int32_t NeonSys_String_UCS2Length(v8::Local<v8::String> str) {
+  return str->Length();
+}
+
+extern "C" size_t NeonSys_String_UCS2Data(uint16_t *out, size_t len, v8::Local<v8::String> str) {
+  return str->Write(out, 0, -1);
```
bnoordhuis: Since you have `len`, I'd pass that in instead of `-1`.  You might want to pass in `NO_NULL_TERMINATION` as well.

src/internal/js/mod.rs
bnoordhuis: There is no (stable) pointer though, V8 only lets you copy out the character data.

src/internal/js/mod.rs
```diff
@@ -307,10 +313,26 @@ impl JsString {
...
+        unsafe {
+            let capacity = neon_sys::string::ucs2_len(self.to_raw());
+            let mut buffer: Vec<u16> = Vec::with_capacity(capacity as usize);
+            let p = buffer.as_mut_ptr();
+            let len = neon_sys::string::ucs2_data(p, capacity, self.to_raw());
+            buffer.set_len(len as usize);
```
bnoordhuis: This should not be necessary.  If the logic is correct, then `len == capacity`.

src/internal/js/mod.rs
RReverser: Ah true. But then, what's the benefit of doings this - it still doesn't guarantee that data is not being transformed on the fly, because, as far as I remember at least, V8 can use also UTF-8 and Latin-1 as internal representations for the string, which would mean that sometimes this will introduce transforms to UCS-2 which weren't done previously.

Can't we instead construct strings in V8 directly from UTF-8 representation that Rust provides? (and then cases like Latin-1 and UTF-8 -> Rust usually won't require any conversion, just copies)

bnoordhuis: > V8 can use also UTF-8 and Latin-1 as internal representations for the string, which would mean that sometimes this will introduce transforms to UCS-2

That's true but it can be avoided with the IsOneByte() and ContainsOnlyOneByte() methods.

> Can't we instead construct strings in V8 directly from UTF-8 representation that Rust provides?

Yes, if I understand you correctly.  V8 has the concept of external strings where the storage for the string lives outside the JS heap.  Node uses them to avoid copying its built-in scripts.

They have some overhead involved compared to regular strings though, so it's only worth it for larger strings.  We established empirically that the turning point is around the 1 MB mark.

RReverser: I guess the overhead mostly comes from processing UTF-8 strings against JS API which works on UCS-2 codepoints (this is why V8 uses UTF-8 internally rarely). Using such strings just for storing & console.log-ging should be fine though, even for small strings.

Also, we could try and allocate one-byte-kind strings for anything that is latin-1 string on Rust side, in which case it's also just a copy in the worst case with no need for transform.

bnoordhuis: I suspect (but didn't verify) it's because external strings are modeled as weak references that need special handling by the garbage collector.  Normal strings are simply collected when unreachable, external strings need to have their destructor invoked.

maxbrunsfeld:
@bnoordhuis @RReverser Thanks for the review. I'm no longer using for this module, so I probably will not get around to landing this PR. I'm cool with closing it out or leaving it open in case anyone else wants this functionality.

dherman:
I'd left this around because I was never sure what I wanted to do with it, specifically because I wasn't sure how universal the idea of string API's for different string representations would be. (Neon tries to avoid V8-specificity.) But now that the N API, which involved stakeholders from multiple JS engines, includes multiple string representations, I feel better about committing to something like this.

I think I'd like to write up a mini-proposal for the API first. So I'll go ahead and close this now, but it'll be there in the closed PR list as a reference for the eventual implementation.

(Max, thanks again for all your help with Neon last year!)

PR #123, opened 2016-10-25
Fix neon on new node versions, and make it work on Electron as well

as-cii:
This pull request changes neon so that it works on newer node versions. In addition, this pull request will enable using neon from Electron applications as well. ✨ It does so by:
1. Using the `NODE_MODULE_VERSION` constant assigned by node-gyp to set the ABI version during `register_module!`. Using an environment variable should be unnecessary, as node-gyp already takes care of all the work to figure out the modules version for the current node instance. 
2. Fixing rustbridge/neon#119, which was causing hard crashes on Node >= 6.3.0. In particular, using `Function` to add methods to a class prototype is now deprecated, and therefore we have changed neon to use `FunctionTemplate`s instead.

@dherman: what do you think? 💭

### DIFF
--- a/crates/neon-sys/src/fun.rs
+++ b/crates/neon-sys/src/fun.rs
@@ -6,6 +6,9 @@ extern "system" {
     #[link_name = "NeonSys_Fun_New"]
     pub fn new(out: &mut Local, isolate: *mut c_void, callback: *mut c_void, kernel: *mut c_void) -> bool;
 
+    #[link_name = "NeonSys_Fun_Template_New"]
+    pub fn new_template(out: &mut Local, isolate: *mut c_void, callback: *mut c_void, kernel: *mut c_void) -> bool;
+
     #[link_name = "NeonSys_Fun_ExecKernel"]
     pub fn exec_kernel(kernel: *mut c_void, callback: extern fn(*mut c_void, *mut c_void, *mut c_void), info: &FunctionCallbackInfo, scope: *mut c_void);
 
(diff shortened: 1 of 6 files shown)


EdShaw:
IMO, it's probably better to get rid of the version override - of you want to target a specific version, it's going to be safer to acquire the right headers. As such, happy for this to go in instead of my pr.

If there a corresponding change in neon CLI to not set the env var?

as-cii:
> IMO, it's probably better to get rid of the version override - of you want to target a specific version, it's going to be safer to acquire the right headers. As such, happy for this to go in instead of my pr.

Agreed, that's why we decided to remove it in this pull request. The idea is that `node-gyp` will automatically take care of that and there are a lots of ways to force a build against a specific version of node (i.e. environment variables, command line arguments, etc.).

> If there a corresponding change in neon CLI to not set the env var?

Yes, we have submitted another pull request (https://github.com/rustbridge/neon-cli/pull/32) which fixes some Windows problems and takes care of removing the environment variable as well.

Phrohdoh:
What about this fixes usage with Electron?

EdShaw:
Not sure if latest electron requires the changes for node compatibility.

The node Abi changes "fixes" electron by removing the assumption that your building for the node that you're running neon-cli with.

dherman:
OK, this PR gave us lots and lots of helpful info and we're making good progress. Windows support is in pretty good shape now, and #194 is tracking the remaining steps we need for Electron support. So I'm closing this one with much gratitude!

PR #124, opened 2016-10-28
Allow JS class instances to be created in Rust

nathansobo:
Previously, you could only create instances of Rust classes via JS. This PR adds a static `::new` method to every declared JS class on the Rust side. This method  takes a scope and a raw backing struct and creates an instance of the JS class wrapping the struct without running the constructor code defined in `init`.

/cc @maxbrunsfeld @as-cii

### DIFF
--- a/crates/neon-sys/src/neon.cc
+++ b/crates/neon-sys/src/neon.cc
@@ -369,6 +369,12 @@ extern "C" void NeonSys_Class_MetadataToClass(v8::Local<v8::FunctionTemplate> *o
   *out = static_cast<neon::ClassMetadata *>(metadata)->GetTemplate(isolate);
 }
 
+extern "C" void NeonSys_Class_MetadataToInstance(v8::Local<v8::Object> *out, v8::Isolate *isolate, void *metadata_pointer, void *internals) {
+  neon::BaseClassMetadata *metadata = static_cast<neon::BaseClassMetadata *>(metadata_pointer);
+  v8::Local<v8::Object> object = metadata->BuildInstance(isolate, internals);
+  *out = object;
+}
+
 extern "C" void *NeonSys_Class_GetInstanceInternals(v8::Local<v8::Object> obj) {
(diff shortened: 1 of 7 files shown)


crates/neon-sys/src/neon_class_metadata.h
```diff
@@ -202,6 +202,13 @@ class BaseClassMetadata: public ClassMetadata {
     }
   }
 
+  virtual v8::Local<v8::Object> BuildInstance(v8::Isolate *isolate, void *internals) {
+    v8::Local<v8::Object> object = GetTemplate(isolate)->InstanceTemplate()->NewInstance(isolate->GetCurrentContext()).ToLocalChecked();
```
KenanSulayman: Wouldn't nan::MaybeLocal be more appropriate here?

crates/neon-sys/src/neon.cc
```diff
@@ -369,6 +369,12 @@ extern "C" void NeonSys_Class_MetadataToClass(v8::Local<v8::FunctionTemplate> *o
   *out = static_cast<neon::ClassMetadata *>(metadata)->GetTemplate(isolate);
 }
 
+extern "C" void NeonSys_Class_MetadataToInstance(v8::Local<v8::Object> *out, v8::Isolate *isolate, void *metadata_pointer, void *internals) {
+  neon::BaseClassMetadata *metadata = static_cast<neon::BaseClassMetadata *>(metadata_pointer);
+  v8::Local<v8::Object> object = metadata->BuildInstance(isolate, internals);
```
KenanSulayman: And here, nan::MaybeLocal?

nathansobo:
I'm going to abandon this PR because we've decided not to use Neon for now, but I'll leave it open in case anybody wants to pick it up and push it forward.

dherman:
Thanks, I will definitely come back to this—right now my top priority is the Windows PR, but after that I will get back into the class system.

dherman:
I'll close this since it won't land as-is, but I will for sure cite you when we land the final version. :D

PR #125, opened 2016-11-11
fix link to v8 embedder's guide

jedireza:


jedireza:
Closing since #127 obsoletes this change.

PR #126, opened 2016-11-11
use npm start script

jedireza:
Depends on https://github.com/rustbridge/neon-cli/pull/33


dherman:
I really like the intuitions behind this -- I agree both that `npm -e require(".")` is an icky thing to recommend, and that `npm install` semantically feels off when you're telling people how to rebuild. But maybe we can open an issue to try to capture what different workflows we might want, so we can think through what the UI should look like for each.

jedireza:
Closing for now. See https://github.com/neon-bindings/neon-cli/issues/35

PR #127, opened 2016-11-15
expand hacking on neon docs

jedireza:
I've expanded the on the _Hacking on Neon_ docs a bit and added some resource links.

I think we could merge this now and it would be an improvement but I'm also fine making it a WIP and to continue expanding on it.


dherman:
r+

PR #128, opened 2016-11-21
expand getting acquainted with rust docs

jedireza:
I've expanded the on the _Getting Acquainted With Rust_ guide a bit.


dherman:
r+

PR #130, opened 2016-11-23
expand neon_sys::raw mod documentation

jedireza:


crates/neon-sys/src/raw.rs
```diff
@@ -1,18 +1,35 @@
...
 
+/// Represents an instance of the V8 runtime.
 pub type Isolate = c_void;
 
 const HANDLE_SCOPE_SIZE: usize = 24;
 
+/// A V8 `HandleScope`.
+///
+/// A `HandleScope` contains `Local` handles. `HandleScope`s are used by V8 to help the garbage
+/// collector do it's bookkeeping. Once a new `HandleScope` is created all subsequently created
```
dherman: its

crates/neon-sys/src/raw.rs
```diff
@@ -26,6 +43,11 @@ impl HandleScope {
 
 const ESCAPABLE_HANDLE_SCOPE_SIZE: usize = 32;
 
+/// A V8 `EscapableHandleScope`.
+///
+/// A `EscapableHandleScope` is like `HandleScope` but also allows us to push `Local` handles out
+/// to the previous `HandleScope`, preventing the `Local` value from being garbage collected when
```
dherman: "preventing...from being garbage collected" isn't _quite_ the right way to say it, I think. Maybe something like, "allowing the `Local` value to remain rooted longer than the `HandleScope` where it was initially rooted." If you think that's too confusing and nit-picky, though, don't worry about it.

jedireza: Not too nit-picky, I like that more.

PR #133, opened 2016-11-28
expand neon_sys::scope mod documentation

jedireza:


jedireza:
Consolidated into # 142

PR #136, opened 2016-11-29
expand neon_sys::string mod documentation

jedireza:


jedireza:
Consolidated into # 142

PR #137, opened 2016-11-29
expand neon_sys::primitive mod documentation and re-order definitions to align

jedireza:


crates/neon-sys/src/primitive.rs
```diff
@@ -1,34 +1,46 @@
+//! Facilities for working with primitive values.
+
 use raw::{Local, Isolate};
 
 extern "system" {
 
+    /// Mutates the `out` argument provided to refer to the `v8::Undefined` object.
```
jedireza: This is the first time mentioning `v8::`, I wonder how specific we should be about this throughout the `neon_sys` docs.

jedireza:
Consolidated into # 142

PR #147, opened 2016-12-12
expand neon-sys docs

jedireza:


dherman:
r+

**So** great to start to have more complete docs.

PR #151, opened 2016-12-23
Consistent calling convention

dherman:
Consistently use the "C" calling convention for neon-sys.

(This PR was separated out from #122.)


jedireza:
r+

PR #152, opened 2016-12-23
Eliminate ABI environment variable

dherman:
This brings in the commit that eliminates the ABI environment variable entirely, as discussed in the thread in #123.


jedireza:
r+ :shipit:

PR #153, opened 2016-12-24
Update the travis test matrix

### LINKED ISSUE #149: Only run travis builds for node versions 4, 6 and 7
Currently we're running builds for rust `stable`, `beta`, `nightly` and node `4.2`, `5.11`, `5`, `6` (which becomes 12 jobs on travis).

I suggest that we change the node versions to `4`, `6`, `7`. By only specifying the major version (ex: `6` not `6.9`) travis will install the latest stable sub-version of the major specified (ex: `6.9.2` for `6`).

I also suggest we only run builds for the supported versions of node and current. Based on the [LTS schedule](https://github.com/nodejs/LTS#lts-schedule), this is `4`, `6`, `7`.

This would also speed-up the build times a bit since it would result in 9 jobs instead of the 12 we have now.

jedireza:
This resolves #149.

👍 Looks like builds are faster. If we could remove `sudo: true` one day, we'd be able to use Travis' container-based infrastructure, and save more time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ Although Neon is only tested on the latest stable Rust, it will likely work on a
 
 Support for [LTS versions of Node](https://github.com/nodejs/LTS#lts-schedule) and current are expected. If you're using a differnt version of Node and believe it should be supported, let us know.
 
-* [Download Node](http://nodejs.org)
-* [Download Rust](http://rust-lang.org)
+* [Download Node](https://nodejs.org)
+* [Download Rust](https://www.rust-lang.org)
 
 # A Taste...
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Windows support is on the way. Follow [#122](https://github.com/neon-bindings/ne
 | Rust beta    | x      | x      | x      |
 | Rust nightly |        |        |        |
 
-Support for Rust stable and beta are expected. We do run builds against nightly, but allow them to fail.
+Although Neon is only tested on the latest stable Rust, it will likely work on a range of stable versions of Rust. It is known to require at least Rust 1.13.
 
 Support for [LTS versions of Node](https://github.com/nodejs/LTS#lts-schedule) and current are expected. If you're using a differnt version of Node and believe it should be supported, let us know.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,17 +27,34 @@ This will ask you a few questions and then generate a project skeleton for you.
...
+
+### Rust and Node
+
+|              | Node 4 | Node 6 | Node 7 |
+| ------------ | ------ | ------ | ------ |
+| Rust stable  | x      | x      | x      |
+| Rust beta    | x      | x      | x      |
+| Rust nightly |        |        |        |
+
+Support for Rust stable and beta are expected. We do run builds against nightly, but allow them to fail.
```
dherman: This doesn't state a minimum Rust stable version. We talked in #149 about our Node version policy but not our Rust version policy.

Rust seems stricter about backwards compatibility than Node, so maybe we'll be able to get away with just stating a minimum Rust stable version. I'm sure we can't realistically test on all Rust versions > _min_, but we should at least **state** a minimum. We could say something like:

> Although Neon is only tested on the latest stable Rust, it will likely work on a range of stable versions of Rust. It is known to require at least Rust 1.13.

(I think we should bump our lower bound from 1.7 to 1.13 so that our programming style can rely on `?` now.)

Let's go with the above verbiage for now, unless you think there's more we could do (maybe a small smoke test for the minimum Rust version in our CI so we can at least do a sanity check whether our recommended minimum is still plausible?) and merits another RFC bug.

README.md
```diff
@@ -27,17 +27,34 @@ This will ask you a few questions and then generate a project skeleton for you.
...
+
+Support for [LTS versions of Node](https://github.com/nodejs/LTS#lts-schedule) and current are expected. If you're using a differnt version of Node and believe it should be supported, let us know.
+
+* [Download Node](http://nodejs.org)
```
dherman: use an https link

README.md
```diff
@@ -27,17 +27,34 @@ This will ask you a few questions and then generate a project skeleton for you.
...
+* [Download Rust](http://rust-lang.org)
```
dherman: use https://www.rust-lang.org since that's canonical

jedireza:
@dherman new commits pushed. We should be good to go.

dherman:
:rocket:

PR #158, opened 2016-12-24
Remove reference to multirust.

winding-lines:
The recommended installer is now rustup. It is linked from the main rust page.
Keep things simple and have just one rust link in the README.


dherman:
Thanks, @winding-lines - this will actually already be removed in #153 as soon as we fix a couple nits there. But thank you!

dherman:
On a related note, what I was about to look into was fixing neon-cli to work with rustup instead of multirust.

PR #159, opened 2016-12-25
Windows npm issues

dherman:
(Pulled out from #122.)

### DIFF
--- a/crates/neon-sys/build.rs
+++ b/crates/neon-sys/build.rs
@@ -32,14 +32,14 @@ const NPM_COMMAND : &'static str = "npm.cmd";
 
 fn build_object_file() {
     // Ensure that all package.json dependencies and dev dependencies are installed.
-    Command::new(NPM_COMMAND).arg("install").status().ok()
+    Command::new(NPM_COMMAND).args(&["install", "--silent"]).status().ok()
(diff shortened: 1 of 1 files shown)


jedireza:
r+ to the changes, they're simple enough.

I wonder what we lose by adding `--silent`? At first glance this sounds like a bug that should be filed up-streamed. If we get around this using `--silent` do we want to create an issue that tracks this and removes the argument when it's not required?

jedireza:
I see now that the output of cargo build scripts are suppressed anyways, unless an error occurs.

 - https://github.com/rust-lang/cargo/issues/1106

There is a `-vv` argument (possibly undocumented) we can pass to cargo to see the output of the build script while it's running.

 - https://github.com/rust-lang/cargo/pull/2630

This can be helpful in the case of windows builds hanging since an error hasn't actually occurred yet (hanging process). With this I'm able to get more insight into what's causing the hang (possibly something related to `$ node-gyp rebuild`).

dherman:
OK, it took a while but the bug has been identified in Rust's `Command` API, and a PR has been filed:

https://github.com/rust-lang/rust/pull/38835

So we should just go ahead with the workaround but leave a comment to remove the workaround once the Rust fix has been in long enough. We may want to wait at least a couple of Rust release cycles so people can use older Rusts.

PR #165, opened 2017-01-21
Windows linker issues

dherman:
(Pulled out from #122.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,11 @@ homepage = "https://www.neon-bindings.com"
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
 exclude = ["neon.jpg"]
+build = "build.rs"
+
+# This name is arbitrary, but allows us to re-export the location of node.lib
+# to dependent packages so they can link on Windows.
+links = "neon-sys"
 
 [dependencies]
 cslice = { version = "=0.1.1", path = "crates/cslice" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,12 @@ license = "MIT/Apache-2.0"
 exclude = ["neon.jpg"]
 build = "build.rs"
 
-# This name is arbitrary, but allows us to re-export the location of node.lib
-# to dependent packages so they can link on Windows.
-links = "neon-sys"
+# Sharing the name of the library produced by neon-sys (which will be
+# renamed to `neon-runtime` -- see issue #169) allows the `build.rs`
+# script for Neon projects to find the path to `node.lib` in Windows.
+#
+# See also: http://doc.crates.io/build-script.html#the-links-manifest-key
+links = "neon-runtime"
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -8,6 +8,11 @@ homepage = "https://www.neon-bindings.com"
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
 exclude = ["neon.jpg"]
+build = "build.rs"
+
+# This name is arbitrary, but allows us to re-export the location of node.lib
+# to dependent packages so they can link on Windows.
+links = "neon-sys"
```
jedireza: If this name is arbitrary, and we're hapy with the one chosen, maybe we could re-phrase this comment and maybe a link to the crates.io docs about this key.

```toml
# This name allows us to re-export the location of node.lib
# to dependent packages so they can link on Windows.
# http://doc.crates.io/build-script.html#the-links-manifest-key
```

crates/neon-sys/package.json
```diff
@@ -1,14 +1,10 @@
...
   "scripts": {
-    "configure-release": "node-gyp configure",
-    "build-release": "node-gyp build",
-    "configure-debug": "node-gyp configure -d",
-    "build-debug": "node-gyp build -d"
+    "preinstall": "echo 'Skipping node-gyp installation as part of npm install.'"
   },
   "devDependencies": {
-    "nan": "^2.3.2",
-    "node-gyp": "^3.3.1"
```
jedireza: Do we expect `node-gyp` to be installed globally?

tests/native/Cargo.toml
```diff
@@ -3,10 +3,14 @@ name = "tests"
...
 license = "MIT/Apache-2.0"
+build = "build.rs"
 
 [lib]
 name = "tests"
 crate-type = ["dylib"]
 
+[dev-dependencies]
+neon-build = "0.1.11"
+
```
jedireza: So the `neon-build` crate isn't required in production? Will production builds still succeed?

crates/neon-sys/package.json
bnoordhuis: It's installed with node (technically, npm) so yes.  Depending on a specific node-gyp version is an anti-pattern in most cases.

I get bug reports from time to time where people complain node-gyp stopped working for them and then it turns out they've pinned it to a five years old release.

Cargo.toml
dherman: I'll tweak the comment but separately I've filed #169 to rename `neon-sys`. Thanks for calling this out.

tests/native/Cargo.toml
dherman: Good catch. Will fix.

dherman: Also filed neon-bindings/neon-cli#46 and committed da0160757b233f1fcd9389041be61f74245076b4 to fix the template in the CLI.

PR #172, opened 2017-01-24
Windows node gyp issues

dherman:
(Pulled out from #122.)


dherman:
I successfully ran all the tests in my Windows VM, so this should be good to go. After this, all that's left to merge from #122 is the commits that set up AppVeyor CI.

PR #183, opened 2017-02-02
rename neon-sys to neon-runtime

### LINKED ISSUE #169: rename neon-sys
@jedireza's review of #165 caused me to go learn about the [`links` manifest key](http://doc.crates.io/build-script.html#the-links-manifest-key) and [*-sys packages](http://doc.crates.io/build-script.html#-sys-packages), and I realized I'd misunderstood the *-sys convention. `neon-sys` should be the name of a crate that provides Rust bindings for the native [libneon](http://webdav.org/neon/) library.

So we should rename (yank + create new) the `neon-sys` crate. I think `neon-runtime` is a pretty good intuition, since you can think of neon as providing a kind of custom runtime system for mediating Node and Rust.

dherman:
Resolves #169.


jedireza:
r+ :shipit:

PR #189, opened 2017-02-18
run node-gyp from a local dependency

dherman:


dherman:
@jedireza What do you think of bumping the Rust version requirement to the more recent 1.15, at least in Windows? With this PR I'm pretty sure it's required; with 1.13 it was hanging, I think because of rust-lang/rust#38811.

jedireza:
I'm fine with bumping the minimum version of Rust to 1.15 for everything. It seems like a maintenance headache to require different versions of Rust for different environments.

Changes look good. We should update the README with the minimum version for Rust though.

PR #191, opened 2017-02-28
Fail the build if a test fails

jedireza:
This issue may be relevant: https://github.com/travis-ci/travis-ci/issues/1066

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,10 +38,10 @@ matrix:
   allow_failures:
     - rust: nightly
 
-script:
-- |
-  (cd tests/ && npm test)
+script: |
+  cd tests/ &&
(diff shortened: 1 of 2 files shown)


jedireza:
@dherman there's a failure with `compile_tests` and it looks like it's been failing for a while now (for the last month, if not longer). Should we open another issue for that or try to fix it in this PR?

Also it doesn't look like AppVeyor is running `compile_tests` for Rust nightly on Windows, is that something we should address here, in another issue or not worry about it?

dherman:
@jedireza Sorry for the delay. Let's merge this and open another issue for the `compile_tests` failure.

dherman:
r+

PR #214, opened 2017-06-29
Task API for running background tasks in the libuv thread pool

dherman:


crates/neon-runtime/src/neon_task.h
```diff
@@ -0,0 +1,93 @@
...
+  ~Task() {
+    callback_.Reset();
+  }
+
+  void execute() {
+    result_ = perform_(rust_task_);
+  }
+
+  void complete() {
+    Nan::HandleScope scope;
```
bnoordhuis: You should enter an Isolate::Scope and Context::Scope here:
```c++
v8::Isolate::Scope isolate_scope(isolate_);
v8::HandleScope handle_scope(isolate_);  // Or Nan::HandleScope
v8::Local<v8::Context> context = v8::Local<v8::Context>::New(isolate_, context_);
v8::Context::Scope context_scope(context);
// ...
```
Store the context when the task is created because it can change.

Using `v8::Isolate::GetCurrentContext()` in this method is not sound because it's possible no context has been entered or it's a different one from the one that created the task.

crates/neon-runtime/src/neon_task.h
```diff
@@ -0,0 +1,93 @@
...
+    if (trycatch.HasCaught()) {
+      argv[0] = trycatch.Exception();
+      argv[1] = v8::Undefined(isolate_);
+    } else {
+      argv[0] = v8::Null(isolate_);
+      argv[1] = completion;
+    }
+
+    v8::Local<v8::Function> callback = v8::Local<v8::Function>::New(isolate_, callback_);
+    callback->Call(isolate_->GetCurrentContext(), v8::Null(isolate_), 2, argv);
```
bnoordhuis: If `try_catch.HasCaught()` is true, there is still an exception pending here and you can't call into the VM.  If you scope the TryCatch in curly braces, it will have been cleared.

Whether it's a good idea to catch exceptions and continue is up for debate, though.

crates/neon-runtime/src/neon_task.h
```diff
@@ -0,0 +1,93 @@
...
+      perform_(perform),
+      complete_(complete)
+  {
+    request_.data = this;
+    result_ = nullptr;
+    callback_.Reset(isolate, callback);
+  }
+
+  ~Task() {
+    callback_.Reset();
```
bnoordhuis: Not strictly necessary, `~Persistent()` does that by default.

crates/neon-runtime/src/neon_task.h
```diff
@@ -0,0 +1,93 @@
...
+  v8::Isolate *isolate_;
+  void *rust_task_;
+  Neon_TaskPerformCallback perform_;
+  Neon_TaskCompleteCallback complete_;
+  void *result_;
+  v8::Persistent<v8::Function> callback_;
+};
+
+void execute_task(uv_work_t *request) {
+  Task *task = static_cast<Task*>(request->data);
```
bnoordhuis: FWIW, we use a container_of-like macro for that in libuv and node.js.  Saves one level of pointer indirection.

jedireza:
Nice! When this lands... we'll really be cooking with 🔥.

A couple things that crossed my mind:

 - Should we include "async" in any of the naming so it's crystal clear that this is indeed off the main thread? Maybe docs will suffice.
 - A JavaScript test/example would be great to see.

dherman:
@bnoordhuis @jedireza Thank you both so much! I'll make sure to incorporate all your feedback before merging.

crates/neon-runtime/src/neon_task.h
```diff
@@ -0,0 +1,101 @@
...
+    v8::Local<v8::Context> context = v8::Local<v8::Context>::New(isolate_, context_);
+    v8::Context::Scope context_scope(context);
+
+    v8::Local<v8::Value> argv[2];
+
+    argv[0] = v8::Null(isolate_);
+    argv[1] = v8::Undefined(isolate_);
+
+    {
+      v8::TryCatch trycatch(isolate_);
```
bnoordhuis: try/catch no longer deoptimizes in recent versions of V8.  But even if it did, it wouldn't matter because this is C++ code, not JS code - the performance pitfall only applied to JS functions with a try/catch statement.

barrotsteindev: Sorry my bad,didn't notice the try catch was only before the callback, thanks for the answer. Anything I could do to help get this pull request accepted asap?
I really need async support and would gladly contribute if needed

barrotsteindev: Sorry my bad,didn't notice the try catch was only before the callback, thanks for the answer. Anything I could do to help get this pull request accepted asap?
I really need async support and would gladly contribute if needed

dherman:
Thanks to a bunch of debugging, we should be back in a state where this can build. Should be able to merge soon!

I'm re-running the CI builds now to see what's left to fix.

dherman:
AppVeyor is being very suspiciously flakey lately but I'm pretty confident the builds hanging are AppVeyor's fault since we haven't seen them happen at all on multiple machines. And there were no runtime test failures.

So once I reran the CI build a few times and got all green I felt confident enough to merge. I'll reach out to AppVeyor to see what's going wrong.

Issue #168, opened 2017-01-22
Cannot create class for generic Rust type


phrohdoh:
[`BigArchive`](https://github.com/Phrohdoh/libbig/blob/7f2328f066bb677b98f47fb5daf6912b2ede780a/src/lib.rs#L14) is `<T: Read + Seek>`

```rust
declare_types! {
    pub class JsBigArchive for BigArchive {
        init(call) {
            let scope = call.scope;
            let path = call.args.require(call.scope, 0)?.check::<JsString>()?.value();
            Ok(try!(BigArchive::new_from_path(&path)))
        }
    }
}

/*
error[E0243]: wrong number of type arguments: expected 1, found 0
  --> src/feature_neon/mod.rs:28:43
   |
28 |     pub class JsBigArchive for BigArchive {
   |                                           ^ expected 1 type argument

error[E0243]: wrong number of type arguments: expected 1, found 0
  --> src/feature_neon/mod.rs:28:43
   |
28 |     pub class JsBigArchive for BigArchive {
   |                                           ^ expected 1 type argument

error: aborting due to 2 previous errors

error: Could not compile `libbig`.
*/
```

```rust
declare_types! {
    pub class JsBigArchive for BigArchive<T> {
        init(call) {
            let scope = call.scope;
            let path = call.args.require(call.scope, 0)?.check::<JsString>()?.value();
            Ok(try!(BigArchive::new_from_path(&path)))
        }
    }
}

/*
error: no rules expected the token `<`
  --> src/feature_neon/mod.rs:28:42
   |
28 |     pub class JsBigArchive for BigArchive<T> {
   |                                          ^

error: Could not compile `libbig`.
*/
```

dherman:
This one is a trickier issue: you're wanting to define a JS class for a generic type. Depending on what you're trying to do, you could just define the JS class for a specific instantiation of the type, e.g. `BigArchive<i32>` (and annoyingly, because of issue #167 you have to do a type alias like:

```rust
type BigArchiveOfInt = BigArchive<i32>;
declare_types! {
    pub class JsBigArchive for BigArchiveOfInt {
        ... 
    }
}
```

But I suspect that's going to be horrible as soon as you're looking for more than one instantiation of `BigArchive`. You _could_ define a macro to generalize this but it'll be painful.

The more general solution will be for us to find a way to define JS classes for generic types. One answer might be that you could define a generic class declaration like:

```rust
[...]

Issue #207, opened 2017-05-19
Using declare_types with explicit lifetime


jugglingcats:
I believe this is a direct consequence of https://github.com/rust-lang/rust/issues/34303 which hopefully will be fixed, but am adding here for awareness.

It doesn't seem possible to have a lifetime on the underlying struct for a Javascript class, eg:

```
pub struct TestMe<'a> {
    x: &'a mut i32
}

declare_types! {
    pub class JsTestMe for TestMe {
        init(call) {
            Ok(TestMe {x: &5})
        }
    }
}
```
This fails with compile error:
```
error[E0107]: wrong number of lifetime parameters: expected 1, found 0
  --> src\shm/mod.rs:99:35
   |
99 |     pub class JsTestMe for TestMe {
   |                                   ^ expected 1 lifetime parameter
```

Vurich:
Workaround: use `pub class JsTestMe as TestMe for TestMe<'static>`. There is no way to do this with a generic parameter right now, but you can make it concrete.

ryantate13:
Is it possible to get this to work with lifetimes other than `'static`? I get a `use of undeclared lifetime name` error when I try.

kjvalencik:
@ryantate13 It is not because the lifetime _must_ be `'static`. The object is being handed over to v8 for tracking and garbage collection. Therefore, it must be bounded by the `'static` lifetime. I.e., the struct must be capable of being held for the entirety of the program.

If the struct were valid for a shorter time period, it might become invalid while it is still being held by Javascript. Dereferencing this struct would be a use-after-free bug.

Referencing counting (e.g., `Arc`) is a common way to upgrade something to a `'static` lifetime.

I'm going to close this issue because it sounds like `'static` is already supported and lifetimes other than `'static` cannot be supported.

Issue #167, opened 2017-01-22
declare_types `for X` should accept path (including ident)


phrohdoh:
Currently the following isn't possible:

```rust
declare_types! {
    pub class JsBigArchive for ::BigArchive {
        init(call) {
            let scope = call.scope;
            let path = call.args.require(call.scope, 0)?.check::<JsString>()?.value();
            Ok(::BigArchive::new_from_path(&path)?)
        }
    }
}

/*
error: expected ident, found ::
  --> src/feature_neon/mod.rs:26:32
   |
26 |     pub class JsBigArchive for ::BigArchive {
   |                                ^^

error: Could not compile `libbig`.
*/
```

dherman:
Yeah, this is an annoying papercut! Thanks for the report. I see in #168 you figured out the workaround but we should definitely make it nicer.

I'm not sure if allowing `ty` instead of `ident` will cause grammar ambiguities but we can try that first. Failing that, we could allow something like `ident` or `(ty)` (that is, if you want a more general type syntax you have to parenthesize it). That wouldn't be ideal but it'd be better than the status quo.

Vurich:
Workaround: use `pub class JsBigArchive as BigArchive for ::BigArchive`. This macro only takes an `ident` instead of `ty` because it desugars to that.

kjvalencik:
`declare_types` is deprecated.

PR #220, opened 2017-08-10
Allow any type in `declare_types`

Vurich:
Closes https://github.com/neon-bindings/neon/issues/168, https://github.com/neon-bindings/neon/issues/167, https://github.com/neon-bindings/neon/issues/207. This appears to be an oversight, as using the `as $name` form does allow any type.


Vurich:
I realise that this would fail because it passes the `ty` to a pattern expecting a `ident`.

PR #221, opened 2017-08-11
Windows Compilation Fixes

hone:
This fixes linking neon on Windows to account for rust-lang/rust#38850.

It also replaces the gyp `<(target_arch)` var, so `rustc` can find the `node.lib` directory.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,21 +1,23 @@
 os: Visual Studio 2015
 
-platform:
-  - x64
-  - x86
-
 environment:
-  NODEJS_VERSION: "6"
   RUST_BACKTRACE: 1
   matrix:
-    - NODE_ARCHITECTURE: x64
+    - PLATFORM: x64
+      NODEJS_VERSION: "6"
       RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
-
-    - NODE_ARCHITECTURE: x86
+    - PLATFORM: x86
+      NODEJS_VERSION: "6"
+      RUST_TOOLCHAIN: stable-i686-pc-windows-msvc
+    - PLATFORM: x64
+      NODEJS_VERSION: "8"
+      RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
+    - PLATFORM: x86
+      NODEJS_VERSION: "8"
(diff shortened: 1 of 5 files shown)


ctaggart:
@hone, need help getting this to work on AppVeyor?

dherman:
@ctaggart Thanks for that offer! I just tested this on my Windows machine and it's still failing so I don't think it's specific to AppVeyor. We're just dealing with some finicky issues satisfying the linker in Windows with the right incantations. (If you have suggestions about that let me know!) I'm gonna dig into this and see what I can figure out.

dherman:
OK, as it turns out, it was only failing on my machine because I had made a mistake with my environment. Now it's building fine.

@ctaggart So this is indeed an AppVeyor mystery. Any help you can offer would be awesome!

ctaggart:
@dherman What was the mistake in your environment? I opened PR #224 with what I had attempted last night, but it doesn't work yet.

appveyor.yml
```diff
@@ -1,21 +1,23 @@
...
-
-    - NODE_ARCHITECTURE: x86
+    - PLATFORM: x86
+      NODEJS_VERSION: "6"
+      RUST_TOOLCHAIN: stable-i686-pc-windows-msvc
+    - PLATFORM: x64
+      NODEJS_VERSION: "8"
+      RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
+    - PLATFORM: x86
+      NODEJS_VERSION: "8"
```
ctaggart: This matrix looks good & looks like AppVeyor is green!
:four_leaf_clover: :four_leaf_clover: :four_leaf_clover: :four_leaf_clover:

hone: Thanks, yeah I wanted to test both LTS + latest and the old configuration was doing a 2x2 matrix where you'd test x64 toolchain on a x86 machine.

dherman:
@ctaggart I had installed something with chocolatey that I thought was the same as the Windows Build Tools, but it wasn't. So the build was picking up GNU `link` in the path instead of the Visual Studio linker.

dherman:
This is fantastic. @hone thank you so much for your persistence. I'm planning to add some comments and clarifications to the build logic, which has definitely gotten intricate enough to be hard to maintain, but that's on me. :)

PR #231, opened 2017-08-29
refactoring: use `pub(crate)` to drastically simplify the file layout

dherman:
I've been wanting to make the project structure simpler for a long time, and [Rust 1.18's](https://blog.rust-lang.org/2017/06/08/Rust-1.18.html) `pub(crate)` feature makes that possible. This PR restructures the repository just to have one single copy of the module hierarchy in the filesystem, instead of the previous internal hierarchy + facade hierarchy.

I think it's a big win, although there are a couple of extra consequences. First, we have to bump the minimum Rust version required from 1.15 to 1.18. I think that shouldn't be a problem, since Rust's compatibility is strong and upgrades are pretty easy. But I want to make sure users have a chance to weigh in on that.

Second, because of the [somewhat strange privacy lint](https://twitter.com/littlecalculist/status/902548109984538628), it's not possible for `pub` traits or types to refer to `pub(crate)` traits or types. So the PR still has to use a variation on the facade module pattern. For example, in order for the public `Scope` trait to extend `ScopeInternal`, I define `ScopeInternal` as a `pub` item inside of a `pub(crate)` internal module. This still allows us to nest the internal modules instead of having to have two parallel hierarchies, but it's a little unfortunate that it requires the extra inner module.

Anyway, I still think the benefits are worth those (IMO manageable) costs.


src/js/error.rs
```diff
@@ -1 +1,103 @@
-pub use internal::js::error::{throw, JsError, Kind};
+// pub use internal::js::error::{throw, JsError, Kind};
```
jedireza: Are we going to leave these in here commented out?

jedireza: Maybe these caused git not to recognize the file name changes.

dherman: D'oh, yeah, I forgot to delete those -- will do!

dherman:
> Too bad git didn't recognize the file renaming in most of this.

@jedireza I actually should have done this with `git mv`, you're totally right. I just copy-pasta'ed because I was thinking it was a merge of multiple files, but the original facade modules were basically empty and not interesting from a file history standpoint. I'll try to re-do this patch to preserve more of the history, by starting from the internal modules and moving them. Thanks for the idea!

dherman:
@jedireza I redid the history with `git mv`. At the command-line a `git diff` seemed to preserve the history but for some reason GitHub isn't showing it.

jedireza:
Thanks for giving it a shot.

dherman:
@ffflorian @bennyn Is it OK with you guys if, starting with the next release, Neon requires at least Rust 1.18?

@hone Are you OK with this?

I'm going to go ahead and merge this, but I won't publish a release until I know it's not going to break anyone. We can easily revert in the next couple days.

hone:
I have no issues with this.

ffflorian:
No problem for us :)

dherman:
Awesome, thanks!

PR #233, opened 2017-08-30
Update README.md with new rust version requirement

kjvalencik:
Neon requires Rust 1.18 since https://github.com/neon-bindings/neon/pull/231


dherman:
Doh, good catch, thank you :)

PR #236, opened 2017-09-15
Remove JsInteger closes #78

### LINKED ISSUE #78: get rid of Integer type

corbinu:
Remove JsInteger closes #78

(Part of the SDRust Libz Blitzz)


dherman:
Thanks, it's great to get this done. I think I'll try to bundle up a couple of minor incompatible changes with this and do a new release with a new minor version (0.2).

Issue #238, opened 2017-09-18
Prompted responses in `neon new` not escaped in generated package.json


dherman:
_From @bantic on August 11, 2016 4:28_

The response to the prompts in the `neon new` flow are not escaped when they are interpolated into the resulting `package.json`, which can lead to an invalid JSON structure. For instance, entering the value `my "cool" project` for the "description" prompt results in a package.json that looks like:

``` javascript
// ....
"description": "my "cool" project",
// ...
```

_Copied from original issue: neon-bindings/neon-cli#30_

robbiepitts:
Looking into this.

dherman:
Fixed in #238. Thanks, @robbiepitts!

PR #251, opened 2017-09-26
Escape quotes in certain `neon new` input values

robbiepitts:
Addresses https://github.com/neon-bindings/neon/issues/238. The description, author, and git fields could potentially have quotes in them, and we only need to escape double quotes, so this escapes double quotes in those fields only. For the git url though, I simply `encodeURI`.


dherman:
Looks good to me 😃

dherman:
Closes #238

Issue #249, opened 2017-09-22
API for getting the global object


dherman:
There needs to be an API for getting at the global object. In the V8 API you can get this from [v8::Context::Global()](https://v8docs.nodesource.com/node-8.0/df/d69/classv8_1_1_context.html). In Neon it should probably be a method of `Scope`, something like:

```rust
fn global(&mut self) -> Handle<JsObject>;
```

reSHARMA:
I am new to neon , but I really want to contribute on this issue . Can you guide me what should I do next ?

dherman:
@reSHARMA So sorry, @GabrielCastro got to this one first with #249. I'm planning to spend some time tomorrow finding more good tasks for first-time contributors.

dherman:
Fixed in #249.

PR #253, opened 2017-10-08
Create an API for accessing the global object #249

GabrielCastro:
I've implemented the api for getting a reference to the global object as per #249. 
I've made it a method on scope such that the following works
```rust
pub fn call_console_log(call: Call) -> JsResult<JsObject> {
    let scope = call.scope;
    
    let global = scope.global();
    let console = global.get(scope, "console")?.check::<JsObject>()?;
    let log = console.get(scope, "log")?.check::<JsFunction>()?;
    let args: Vec<Handle<JsString>> = vec![JsString::new(scope, "hello world").unwrap()];
    log.call(scope, console, args.into_iter())?;
...
```

### DIFF
--- a/crates/neon-runtime/src/neon.cc
+++ b/crates/neon-runtime/src/neon.cc
@@ -248,6 +248,11 @@ extern "C" size_t Neon_Scope_AlignofEscapable() {
   return alignof(v8::EscapableHandleScope);
 }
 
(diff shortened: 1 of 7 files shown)


dherman:
@GabrielCastro Beautiful work! :D

dherman:
It looks like the Travis failure was a transient npm issue, and our appveyor config is busted right now, so I'm going to merge. (This is definitely not an acceptable state of affairs for the CI--we should get this fixed soon.)

PR #254, opened 2017-10-12
Update README.md to fix wording

KamuelaFranco:
Unless I am misunderstanding the context here, I believe the word choice was meant to be “routing.” Feel free to close immediately if I’m off base here.


RReverser:
No, it is "rooting" (the term related to GC).

Issue #246, opened 2017-09-20
change top-level test workflow to use cargo


dherman:
Running all the Neon tests should be kicked off with `cargo test`, which is the idiomatic way to run tests for a Rust project. This can be done by shelling out to `npm` etc where necessary.

- [x] one test called `cli`:
  - in the `cli` directory: shells out to `npm install` and `npm run transpile`
  - in the `test/cli` directory: shells out to `npm install` and `npm run transpile` and `npm test`
- [x] one test called `static`, only enabled if we're in Rust nightly:
  - in `test/static` directory: shells out to `cargo test`
  - or maybe it can be called into through a compiletest API instead of shelling from cargo to cargo?
- [x] one test called `dynamic`:
  - in `test/dynamic` directory: shells out to `npm install` and `npm test`

Then the CI scripts can be changed to just do a single `cargo test` instead of all of the above.

dherman:
@humphd That would be amazing! I'd be happy to help mentor any students who want to help with this, too. (I have so far been having trouble keeping to a regular schedule monitoring these issue threads, but if you ever need my attention, please feel free -- you or your students -- to ping me on slack or twitter to check a thread, I'll make sure to respond as quickly as possible.)

dherman:
@humphd Looks like @MilanLoveless got to this one already in #255, at least partially. Another infrastructure issue we could really use help with is getting #252 over the finish line -- do you think your students might be interested/able to help with that one?

humphd:
OK, thanks for letting me know.  Reading that PR, it also looks like "make this work fully on Windows" might be something for them to do in future as well.  RE: #252, I know that @SeanPrashad hit this recently, and I was suggesting he look into it.  Maybe he'd like to try this?

I'll circle back to this stuff later and do some bugs with my students as it lines up with topics we're covering.

SeanPrashad:
@humphd @dherman and I spoke briefly about this on Slack last night! I'd love to try and tackle #252 in the near future if that's ok with everyone 😬

humphd:
@SeanPrashad I think it would be a good fit for you, and give you some nice exposure to build automation, which is something I remember you expressing interest in at the start of the term.

dherman:
I believe this was fixed by @MilanLoveless in #255 (with a couple follow-up commits from me for Windows)!

PR #255, opened 2017-10-14
Consolidated tests for NIX systems

MilanLoveless:
# What does this PR do?
This PR allows all tests to be called by running `cargo test` from the root of the project directory on NIX systems. Static tests run conditionally: only when using the nightly toolchain. Further work needs to be done to get Windows tests running properly.

# Background Context
This feature will allow the CI to run a single command to complete all tests. Also, it appears that the static tests are currently failing (they are green when using the nightly toolchain).

# How should it be manually tested?
run `cargo test` from the root of the project directory

# Corresponding Issue
#246 change top-level test workflow to use cargo

# Screenshot(s)
screen shot 2017-10-13 at 8 44 06 pm

# GIF of how this PR makes you feel
giphy 5


MilanLoveless:
I got these failing static tests to pass by switching to the nightly build of Rust.
screen shot 2017-10-14 at 1 28 42 pm

dherman:
This is awesome, and this PR absolutely makes me feel like an animated gif kitteh. 😻😻😻

I'll add a couple minor notes inline.

src/lib.rs
```diff
@@ -266,3 +266,71 @@ macro_rules! declare_types {
...
+    use std::process::Command;
+    let cli_output = if cfg!(target_os = "windows") {
+        Command::new("cmd")
+                .args(&["/C", "Implement this later"])
+                .output()
+                .expect("failed to execute process")
+    } else {
+        Command::new("sh")
+                .arg("-c")
+                .arg("cd cli; npm install; npm run transpile")
```
dherman: I think it'd be better to chain these with `&&` instead of `;` so that the whole command fails if any of the individual steps fails.

src/lib.rs
```diff
@@ -266,3 +266,71 @@ macro_rules! declare_types {
...
+
+    let cli_test_output = if cfg!(target_os = "windows") {
+        Command::new("cmd")
+                .args(&["/C", "Implement this later"])
+                .output()
+                .expect("failed to execute process")
+    } else {
+        Command::new("sh")
+                .arg("-c")
+                .arg("cd test/cli; npm install; npm run transpile && npm test")
```
dherman: Again I think `&&` would be better.

src/lib.rs
```diff
@@ -266,3 +266,71 @@ macro_rules! declare_types {
...
+    use std::process::Command;
+    let static_output = if cfg!(target_os = "windows") {
+        Command::new("cmd")
+                .args(&["/C", "Implement this later"])
+                .output()
+                .expect("failed to execute process")
+    } else {
+        Command::new("sh")
+                .arg("-c")
+                .arg("cd test/static; cargo test")
```
dherman: Another spot for `&&`

src/lib.rs
```diff
@@ -266,3 +266,71 @@ macro_rules! declare_types {
...
+    use std::process::Command;
+    let dynamic_output = if cfg!(target_os = "windows") {
+        Command::new("cmd")
+                .args(&["/C", "Implement this later"])
+                .output()
+                .expect("failed to execute process")
+    } else {
+        Command::new("sh")
+                .arg("-c")
+                .arg("cd test/dynamic; npm install; npm test")
```
dherman: And one last instance where we should use `&&`

src/lib.rs
```diff
@@ -266,3 +266,71 @@ macro_rules! declare_types {
...
+                .output()
+                .expect("failed to execute process")
+    };
+    assert!(cli_test_output.status.success());
+}
+
+#[test]
+fn static_test() {
+    use std::process::Command;
+    let static_output = if cfg!(target_os = "windows") {
```
dherman: Let's have this function skip out early if the current Rust version isn't on the nightly channel. Here's a plan for how to do it:

- Add [rustc_version](https://crates.io/crates/rustc_version) as a [dev dependency](http://doc.crates.io/specifying-dependencies.html#development-dependencies) to this crate.
- Add an `extern crate rustc_version;` to the top level, and I think we need to wrap that with `#[cfg(test)]` so it's only included when we're testing.
- Check if [`version_meta().unwrap().channel`](http://kimundi.github.io/rustc-version-rs/rustc_version/index.html) is [`Channel::Nightly`](http://kimundi.github.io/rustc-version-rs/rustc_version/enum.Channel.html), and if not, return early.

(I wish there was a `#[cfg(nightly)]` attribute but [that hasn't happened yet](https://internals.rust-lang.org/t/setting-cfg-nightly-on-nightly-by-default/1893). If it ever does we can simplify this!)

MilanLoveless:
review comments and suggestions for improvement have been addressed 😄

dherman:
Looks great! Now we can just finish up the windows commands when we talk on Wednesday and this will be ready to go. =)

MilanLoveless:
Awesome!

dherman:
I realized the reason it was failing to test in Windows was that I was using `cargo test` instead of `cargo test --release`. I'll merge this and then afterwards I'll submit a fix that gives you a clearer compiler error if you build with the wrong profile.

PR #258, opened 2017-10-16
Add error message for missing Rust executable

### LINKED ISSUE #216: Better handling of missing rustc executable
Somehow on my build machine I lost rustc from my path, and neon build fails with the following error:

```
TypeError: Cannot read property 'toString' of null
    at Function.BuildSettings.current (/usr/local/lib/node_modules/neon-cli/dist/neon-cli/build-settings.js:29:13)
```

A minor issue but more helpful message would have prevented me from suspecting a node version problem (again!).

ffflorian:
This adds an error message (instead of complaining about `Cannot read property 'toString' of null`) if Rust (particularly `rustc`) is not found.

Fixes #216.

### DIFF
--- a/cli/src/build-settings.ts
+++ b/cli/src/build-settings.ts
@@ -28,8 +28,13 @@ export default class BuildSettings {
   }
 
   static current(toolchain: rust.Toolchain) {
-    let rustc = rust.spawnSync("rustc", ["--version"], toolchain)
-      .stdout
+    const stdout = rust.spawnSync("rustc", ["--version"], toolchain).stdout;
+
+    if (!stdout) {
+      throw new Error('Rust is not installed or rustc is not in your path.');
+    }
+
+    let rustc = stdout
       .toString('utf8')
       .trim();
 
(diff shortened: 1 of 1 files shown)


cli/src/build-settings.ts
```diff
@@ -28,8 +28,13 @@ export default class BuildSettings {
   }
 
   static current(toolchain: rust.Toolchain) {
-    let rustc = rust.spawnSync("rustc", ["--version"], toolchain)
-      .stdout
+    const stdout = rust.spawnSync("rustc", ["--version"], toolchain).stdout;
+
+    if (!stdout) {
+      throw new Error('Rust is not installed or rustc is not in your path.');
```
RReverser: Maybe better to store and print original message from `stderr`? (or is it useless?)

ffflorian: Good idea, let me check.

ffflorian: The original message just contains "spawnSync rustc ENOENT" so that isn't too helpful. Nevertheless, if another error beside `ENOENT` appears, it will also be thrown.

dherman:
I _really_ appreciate these kinds of quality improvements. Thank you, Florian!

Issue #248, opened 2017-09-21
get appveyor back to green


dherman:
The directory reorg this week broke CI; Travis is fixed but AppVeyor is still red. Copy what `.travis.yml` is doing and make that happen in `appveyor.yml`.

PR #252, opened 2017-09-26
WIP: Fix appveyor config

robbiepitts:
Addresses https://github.com/neon-bindings/neon/issues/248. Updated config to reference the new directory structure and execute the necessary npm scripts to run the tests.

PR #259, opened 2017-10-18
Fix Appveyor build

andrewstucki:
Fixes: https://github.com/neon-bindings/neon/issues/248

There are currently a few issues with the broken appveyor build.

1) The newest versions of chalk (2.2.0 at this time) export typescript definitions that conflict with those used from DefinitelyTyped
2) The directory structure appears to have changed since the appveyor build was last passing
3) The cli tests fail (and will continue to do so) due to nexpect not passing `shell` or `windowsVerbatimArguments` to `spawn`

This PR fixes all three of those and updates the appveyor script to more closely mirror the build matrix of travis.

Regarding number 3, since nexpect hasn't been updated in ~3 years I think to get the desired support for windows sub-process spawning you'll either have to fork nexpect or move the test suite away from using it. Until then, the cli tests aren't running on appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,23 +4,37 @@ environment:
   RUST_BACKTRACE: 1
   matrix:
     - PLATFORM: x64
+      NODEJS_VERSION: "4"
+      RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
+    - PLATFORM: x64
       NODEJS_VERSION: "6"
       RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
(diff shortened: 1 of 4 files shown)


SeanPrashad:
I believe this will also close #252!

andrewstucki:
So, I forked nexpect [here](https://github.com/andrewstucki/node-suspect), re-wrote it in Typescript and allowed for passing down all `child_process.spawn` options, which now allows re-enabling the cli tests on windows.

dherman:
Amazing work, especially the persistence to fork nexpect. :D

Issue #179, opened 2017-01-31
[Windows] Build doesn't work


JoeyAcc:
I saw that very recently there were a number of commits to enable building on windows, so I decided to try it out on my Windows 10 workstation, in combination with `Node v4.5.0`. When I try to build, I get an error (see the log below).

I have installed:
* `neon-cli` globally with `npm install -g neon-cli`
* the Windows build tools as outlined by [Option 1 for Windows](https://github.com/nodejs/node-gyp#installation).

I'm also using this in the `Cargo.toml` of the native sub-project:
```
[dependencies]
# neon = "0.1"   # I would normally use this, but there's no release supporting building on Windows yet
neon = { git = "https://github.com/neon-bindings/neon.git", rev = "82632b7" }
```
The project itself was written using a version of neon before the Windows build patches landed, I don't know if that matters or not.

I then run `npm run-script install`, which in turn just calls `neon build`. I then get the error log below as output. 
I'm not sure why linking fails, but I'd appreciate any help I can get on the issue.

If additional information is useful, please let me know and I'll add it.

Error log:
````
c:\Users\Joey\WebstormProjects\quip-npm>npm run-script install

> quip-npm@0.1.0 install c:\Users\Joey\WebstormProjects\quip-npm
> neon build

neon info running cargo
neon info cargo build --release
   Compiling quip-npm v0.1.0 (file:///C:/Users/Joey/WebstormProjects/quip-npm/native)
warning: unused `#[macro_use]` import, #[warn(unused_imports)] on by default
 --> src\lib.rs:2:1
  |
2 | #[macro_use] extern crate quip;
  | ^^^^^^^^^^^^

error: linking with `link.exe` failed: exit code: 1120
  |
[...]

PR #261, opened 2017-10-22
report a clear error for windows debug builds

dherman:
Windows builds don't currently work with a `--debug` profile. They fail with incredibly obscure link errors (such as the one reported in #179), which is really painful for users to try to troubleshoot. It would be great to support Windows debug builds, but until we can, this PR fails fast with a clearer compiler error.


MilanLoveless:
Looks great! I pulled it down and verified that it still runs as intended on my mac.
Is there a case where we would have rust tests anyplace outside of `tests/static`? We may want to move the code to check release vs. debug build arguments and associated constants into `/src/lib.rs`

dherman:
> We may want to move the code to check release vs. debug build arguments and associated constants into `/src/lib.rs`.

That’s a pretty nice idea—we could provide some metadata about the build that would occasionally be useful for introspection, similar to how Node provided this info via `process.versions`. (This would also have the nice side benefit of eliminating the duplicated logic and extra boilerplate I just added to `test/static`.)

dherman:
OK, we're finally passing! I'll merge this and then finish the Windows test logic.

Issue #264, opened 2017-10-25
Allow closures for JsFunction initialization


abrodersen:
Currently, the constructor for  `JsFunction` values only permits passing static function pointers. This makes continuation passing APIs impossible to write. 

For example, say I want to design a function `doWork` that accepts a callback, and this callback must provide a value asynchronously. I would design the function such that I would pass a continuation into the callback like so:

````
doWork(someData, (continue) => {
  doSomethingAsynchronously(otherData, (result) => { continue(result) });
});
````

Within Rust, I would like to pass some context into the continuation:

````
fn do_work(call: Call) -> JsResult<JsValue> {
    let callback = call.arguments.require(call.scope, 0)?.check::<JsFunction>()?;
    let some_other_state = ();
    let continuation = JsFunction::new(scope, move |call| {
        let extra_data = call.require(call.scope, 0)?;
        // do something with some_other_state
    });
    
    // I might want to pass the context into a task
    let task = MyTask { some_other_state };
    task.schedule(callback);

    Ok(JsUndefined::new().upcast())
}
````

This style of API is very common in Node.js, especially when doing things like asynchronous IO.

Looking at the source code, it appears that the function pointer is merely turned into a raw pointer. Couldn't the code be modified to use a `Box<FnMut(Call) -> JsResult<JsValue>>` when creating a `JsFunction`? `Box` even has built in methods for converting too and from raw pointers.

kjvalencik:
These use cases are covered under the Classes 2.0 RFC and the async task API.

PR #267, opened 2017-10-30
Modify API to allow boxed closures for JsFunctions.

abrodersen:
Fix for issue #264. Feedback is welcome.

### DIFF
--- a/src/js/mod.rs
+++ b/src/js/mod.rs
@@ -20,6 +20,7 @@ use self::internal::{ValueInternal, SuperType, FunctionKernel};
 pub(crate) mod internal {
     use std::mem;
     use std::os::raw::c_void;
+    use std::panic::AssertUnwindSafe;
     use neon_runtime;
     use neon_runtime::raw;
     use mem::{Handle, Managed};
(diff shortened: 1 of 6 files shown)


nixpulvis:
Is there a use case for unboxed JsFunctions, other than the tiny memory overhead.

src/js/mod.rs
```diff
@@ -48,26 +49,37 @@ pub(crate) mod internal {
     }
 
     #[repr(C)]
-    pub struct FunctionKernel<T: Value>(pub fn(Call) -> JsResult<T>);
+    pub struct FunctionKernel<T: Value>(pub Box<FnMut(Call) -> JsResult<T>>);
```
RReverser: Might be worth having type parameter `F: FnMut(Call) -> JsResult<T>` to avoid dynamic dispatch.

dherman:
Thanks for this contribution! What I'd like to do over time is have new and/or changed APIs go through RFCs, since they affect the overall design. I don't want to throw process at you since you already did this work but it's definitely worth a good discussion.

It's up to you if you want to write an RFC or not but would you mind at least fleshing out some of the motivation and rationale (basically the kinds of stuff that goes in an RFC)?

abrodersen:
I have created an RFC for this pull request: neon-bindings/rfcs#12

rebo:
Hi, I've tried this pull request, when I call twice a jsFunction defined in this way the runtime crashes with a malloc pointer being freed was not allocated error.

code:

            let foo = String::from("foo");
        let bar_func = JsFunction::new(scope, Box::new(move |call| {
        let msg = foo.clone() + "bar";
        
        Ok(JsString::new(call.scope, &msg).unwrap())
        }))?;
        
        
        let args: Vec<Handle<JsValue>> = vec![];
        bar_func.call(scope,JsUndefined::new(),args.clone());
        bar_func.call(scope,JsUndefined::new(),args.clone());  // crashes on second call, first is fine though

abrodersen:
@rebo Thank you for the feedback! I'll look into it today.

abrodersen:
@rebo Can you try out the updated code? I think it should fix the issue you identified.

abrodersen:
@dherman I've updated this PR with feedback from you and others. Would you consider merging it?

rebo:
Hi @abrodersen , seems to work fine now thanks for fixing.  I have it working with your persistent handles and a neon async task.

kjvalencik:
Closing this until we have a solid RFC and have resolved https://github.com/neon-bindings/rfcs/pull/12#issuecomment-535526184.

PR #276, opened 2017-11-12
let's get some lockfile discipline around here!

dherman:
We keep getting pwned by TypeScript upgrades. Turns out there's a known solution for this... 😊

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ Cargo.lock
 **/build
 **/index.node
 **/artifacts.json
-**/package-lock.json
 cli/lib
 test/cli/lib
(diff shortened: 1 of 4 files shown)


dherman:
Oh, ha, our Node version matrix includes versions that have npm < 5. So to get across-the-board lockfiles I guess we'd need to either use Yarn or wait until we don't have to support node < 8.

RReverser:
Well you can also use old good shrinkwraps, and new npm should even generate one for you out of the lockfile with `npm shrinkwrap`.

Another option would be also to use `~` instead of `^` for version ranges of often-breaking packages.

PR #282, opened 2017-11-30
Replace RefCell<bool> with Cell<bool>

maciejhirsz:
- ~I've went through and `#[inline]`d a lot of the stuff. I'd say at least 80% of those are totally fine (there is a lot of functions that just do some transmutes which could and probably should be transparent to the compiler), some might be unnecessary. Do point out places where you feel like it's too much.~
- There was a `RefCell<bool>` in `scope.rs` that I swapped for `Cell<bool>`, yay no runtime cost.

~Overall I had some good results with this, was able to boost a Node.js benchmark from \~386k to \~424k ops/sec. The cost of interacting with V8 might not be the highest priority for very time consuming tasks (going into milliseconds), but the lower that cost is, the broader the number of problems neon can solve.~

~I've also found that catching panics is pretty costly, I reckon having a feature flag to disable it (the "I-know-what-I-am-doing-my-code-doesnt-panic-anyway-and-if-it-does-then-v8-can-die-with-it" flag) could be handy, thoughts?~

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -3,7 +3,7 @@
 use std::mem;
 use std::os::raw::c_void;
 use std::marker::PhantomData;
-use std::cell::RefCell;
+use std::cell::Cell;
 use neon_runtime;
 use neon_runtime::raw;
 use mem::Handle;
@@ -13,11 +13,11 @@ use self::internal::ScopeInternal;
 
 pub(crate) mod internal {
     use vm::internal::Isolate;
-    use std::cell::RefCell;
 
     pub trait ScopeInternal: Sized {
         fn isolate(&self) -> Isolate;
-        fn active_cell(&self) -> &RefCell<bool>;
+        fn active(&self) -> bool;
+        fn set_active(&self, bool);
     }
 }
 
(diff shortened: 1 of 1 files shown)


RReverser:
Did you try compiling with LTO instead of manually marking each function? Unless I'm missing something, you're basically enforcing LTO manually (inter-crate inlining is the only effect of `#[inline]` on already small functions), and there is a flag for that.

maciejhirsz:
@RReverser I wasn't even aware that's a thing (you learn every day, eh), will give it a shot and report back.

LifeIsStrange:
Isn't LTO superced by thinLTO ?

RReverser:
@LifeIsStrange LTO is a common name for link-time optimisation, while ThinLTO is just one of implementations.

RReverser:
@maciejhirsz How did it go, by the way?

maciejhirsz:
@RReverser sorry for leaving this hanging. I did some quick tests but it wasn't super conclusive so I left it on the backburner and got swarmed with other things. IIRC the inlining did have some effect, but there wasn't much to it. I'm on holidays soon, but I'm not sure if I'll have time to pick this up.

I'll rebase my PR to include only this change:
> There was a `RefCell<bool>` in `scope.rs` that I swapped for `Cell<bool>`.

maciejhirsz:
@RReverser can you restarted the failed CI? I don't really see how it could have been affected by this PR.

maciejhirsz:
@dherman ^ :)

dherman:
@maciejhirsz Restarted the builds and everything passed. Thanks!

I'm not an expert at Cell vs RefCell -- can you help me understand why this is both equivalent and faster?

One tiny style change I'd make is to replace `set_style(bool)` with `activate()` and `deactivate()` since it never seems to actually need a computed boolean and the names read a little more nicely. But I can make this change after merging this PR.

Last question: I've never used LTO; how do you enable it?

RReverser:
@dherman Difference in performance comes from the fact that Cell doesn't do dynamic checks to ensure that data is not mutably borrowed elsewhere. It can get away with that because it can hold only Copy types, which allows to 1) return a copy of value on `.get()` and 2) simply put the new value in place on `.set()` without worrying that previous value might still be dynamically borrowed.

RReverser:
@dherman As for LTO, you can enable it with something like

```toml
[profile.release]
lto = true
```

in Cargo.toml.

dherman:
@maciejhirsz Thank you for this PR <3

PR #283, opened 2017-12-01
Allow mutable borrow of self for Tasks.

abrodersen:
Some `Task` implementations might need mutable access to the struct data during `perform`. Nobody else will have access to the struct after calling `schedule` since it takes ownership of `self`, so we can safely allow mutable access inside of `perform`.

### DIFF
--- a/src/task.rs
+++ b/src/task.rs
@@ -25,7 +25,7 @@ pub trait Task: Send + Sized {
     type JsEvent: Value;
 
     /// Perform the task, producing either a successful `Output` or an unsuccessful `Error`. This method is executed in a background thread as part of libuv's built-in thread pool.
-    fn perform(&self) -> Result<Self::Output, Self::Error>;
+    fn perform(&mut self) -> Result<Self::Output, Self::Error>;
 
     /// Convert the result of the task to a JavaScript value to be passed to the asynchronous callback. This method is executed on the main thread at some point after the background task is completed.
     fn complete<'a, T: Scope<'a>>(self, scope: &'a mut T, result: Result<Self::Output, Self::Error>) -> JsResult<Self::JsEvent>;
@@ -51,7 +51,7 @@ pub trait Task: Send + Sized {
 }
 
 unsafe extern "C" fn perform_task<T: Task>(task: *mut c_void) -> *mut c_void {
-    let task: Box<T> = Box::from_raw(mem::transmute(task));
+    let mut task: Box<T> = Box::from_raw(mem::transmute(task));
     let result = task.perform();
     Box::into_raw(task);
     mem::transmute(Box::into_raw(Box::new(result)))
--- a/test/dynamic/native/src/js/tasks.rs
+++ b/test/dynamic/native/src/js/tasks.rs
@@ -11,7 +11,7 @@ impl Task for SuccessTask {
     type Error = String;
     type JsEvent = JsNumber;
 
-    fn perform(&self) -> Result<Self::Output, Self::Error> {
+    fn perform(&mut self) -> Result<Self::Output, Self::Error> {
         Ok(17)
     }
 
@@ -33,7 +33,7 @@ impl Task for FailureTask {
     type Error = String;
     type JsEvent = JsNumber;
 
-    fn perform(&self) -> Result<Self::Output, Self::Error> {
+    fn perform(&mut self) -> Result<Self::Output, Self::Error> {
         Err(format!("I am a failing task"))
     }
 


kjvalencik:
It would be a more substantial breaking change, but I think the most flexibility would be offered by having `perform` take ownership of `self` and modify `complete` to be a static method.

This makes tasks that must consume data as part of perform less error prone and more simple. For example, today if you need to consume a portion of data, your `Task` struct must use `Cell<Option<T>>`. Even with mutability added, you would still need `Option<T>`.

The downside is that code that would like to pass data along to the `complete` method that isn't used by `perform` would need to thread it through `perform`. I think this is trade-off is acceptable.

If carrying extra data is something we would like to support, I think it would be better as a separate object. That way we could also relax some of Send/Sync constraints.

qm3ster:
@kjvalencik Was this comment to any extent motivated by our discussion in the Slack? :stuck_out_tongue:
Also, I have since learned that calling them "static method"s is bad manners. In Rust, they are "associated functions" :stuck_out_tongue:

kjvalencik:
It absolutely was! It was a great suggestion. I gave you credit in the slack dev channel, but failed to here. Sorry about that!

qm3ster:
Oh :O
I didn't know there's such a thing.
Can I come in?

qm3ster:
Additionally, it would probably be more flexible to just pass the handle to the callback into `complete`.
The current way makes it clearer what to do with errors, but it would be more flexible to be able to conditionally call the JsFunction 0, 1, or more times.
:+1: for the separate `extra data` struct to be available in the `complete`, especially if it can have references to values from the Js world by not being sent.

qm3ster:
@kjvalencik, I just thought of another killer use for passing the handle to the callback into `complete`.
If this is done, the `Task` can now reschedule itself or a different `Task` on this callback, in addition to calling it!
This would significantly improve the (`event-emitter`)[https://github.com/neon-bindings/examples/tree/master/event-emitter] usecase, removing the `setImmediate` polling trickery. It will look like a real subscription on the JS side.

kjvalencik:
@qm3ster In my opinion, that it is a significant enough departure from the `Task` API that it should be considered a distinct use case. I think there is room for multiple async APIs. `Task` can be the low boilerplate simple one-off use case, with others that could support more complex workflows.

qm3ster:
Totally. I can see ergonomic value for people coming from JS of having `complete` return a value.

dmitry-zakablukov:
Immutable `Task::perform()` doesn't allow to move variables into a closure, see https://github.com/neon-bindings/neon/pull/520

kjvalencik:
@dmitry-zakablukov The workaround is to _also_ wrap the `Option` with a `Cell` for interior mutability.

This is obviously very poor ergonomics. This is one of the top items on the neon teams backlog for discussion. Currently there are two proposals:

1. Change `&self` to `&mut self` as proposed in this PR and your own.
    * **Pro:** Simple, less significant change to the API.
    * **Con:** Taking ownership of a value requires runtime correctness checks instead of compile time. E.g., checking for `Option::None`.
2. Change `&self` to `self` so that `perform` takes ownership. `complete` becomes an associated function instead of an instance method.
    * **Pro:** Allows for APIs that require moving variables / taking ownership.
    * **Con:** Breaking change. Also, requires developers to thread `self` in the return value if they need access to it in `complete`.

Another option would be to have two different APIs. A `&mut self` version of the API could be trivially built on top of one that takes ownership. However, this would result in API bloat.

@dmitry-zakablukov Do you have any thoughts on the issue?

dmitry-zakablukov:
@kjvalencik I have found that the current API is sufficient for my case if the objects that I move into the closure are cloneable. In other words, I have to clone all arguments to make closure implement Fn instead of FnOnce.
But if the objects are not cloneable...

In my opinion, the second approach is better. And it should be done asap to affect less number of projects dependent on neon.

kjvalencik:
@dmitry-zakablukov Changing `&` to `&mut`, even though more minor, is *also* a breaking change. Perhaps unlikely, but, consider a scenario where a user manually calls `perform` multiple times. This would no longer be allowed if the method also returned a reference to `self`.

qm3ster:
Well, you know what they say: the more breaking the more urgent :D

dmitry-zakablukov:
@kjvalencik , does neon API allow to call `perform()` several times? Which result of this calls will be passed to `complete()`?
It looks like a callback: to produce several objects via a single call to neon module.

kjvalencik:
Neon won't ever call `perform` multiple times (`schedule` takes ownership). It would need to be a developer manually calling `perform` from their application code.

kjvalencik:
Hello! I have created an RFC for a new `Task` API to resolve these ownership issues. The new API is significantly different from the existing one and I'd love to have feedback on the RFC PR.

tl;dr-- I am proposing a closure based API to replace the existing `Task` trait`

```rust
pub fn add_one(mut cx: FunctionContext) -> JsResult<JsUndefined> {
    let n = cx.argument::<JsNumber>(0)?.value();
    let cb = cx.argument::<JsFunction>(1)?;

    cx.task(move || {
            let result = n + 1.0;

            move |mut cx| Ok(cx.number(result))
        })
        .schedule(cb)
}
```

https://github.com/neon-bindings/rfcs/pull/30

dherman:
I'm going to go ahead and close this since we have active work now on a newer approach with @kjvalencik's TaskBuilder RFC (neon-bindings/rfcs#30). But this PR was an important first step towards helping us understand the use cases. ❤️

PR #284, opened 2017-12-08
Implement a safe wrapper for persistent handles.

abrodersen:
I plan to write up an RFC, but I thought I would put the code up to solicit feedback since I've already implemented it.

Create a `PersistentHandle` type that can hold references to managed v8 objects. The `PersistentHandle` stores a pointer to a `Nan::Persistent<v8::Value>` object. Since the underlying pointer references heap memory, the pointer can be safely passed between threads. A `Drop` trait implementation ensures that the v8 reference is cleared using `Reset()` and the memory is deallocated. There's also a `Clone` implementation that allows cloning the underlying reference. **Note:** The `Clone` trait only borrows `&self`, but I am unsure if `Reset(Nan::Persistent)` is thread-safe or not.

### DIFF
--- a/crates/neon-runtime/src/neon.cc
+++ b/crates/neon-runtime/src/neon.cc
@@ -538,7 +538,7 @@ extern "C" bool Neon_Mem_SameHandle(v8::Local<v8::Value> v1, v8::Local<v8::Value
 }
 
 extern "C" void *Neon_Mem_NewPersistent(v8::Local<v8::Value> val) {
-  return static_cast<void*>(new Nan::Persistent<v8::Value>(val));
+  return new Nan::Persistent<v8::Value>(val);
 }
 
 extern "C" void Neon_Mem_New(v8::Local<v8::Value> *out, void *persistent) {
@@ -550,12 +550,11 @@ extern "C" void *Neon_Mem_ClonePersistent(void *persistent) {
   Nan::Persistent<v8::Value>* original = static_cast<Nan::Persistent<v8::Value>*>(persistent);
   Nan::Persistent<v8::Value>* copy = new Nan::Persistent<v8::Value>();
   copy->Reset(*original);
-  return static_cast<void*>(copy);
+  return copy;
 }
 
 extern "C" void Neon_Mem_DeletePersistent(void *persistent) {
   Nan::Persistent<v8::Value>* p = static_cast<Nan::Persistent<v8::Value>*>(persistent);
-  p->Reset();
   delete p;
 }
 
--- a/crates/neon-runtime/src/neon.cc
+++ b/crates/neon-runtime/src/neon.cc
@@ -537,6 +537,15 @@ extern "C" bool Neon_Mem_SameHandle(v8::Local<v8::Value> v1, v8::Local<v8::Value
   return v1 == v2;
 }
 
+extern "C" void *Neon_Mem_NewPersistent(v8::Local<v8::Value> val) {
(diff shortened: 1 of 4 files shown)


crates/neon-runtime/src/neon.cc
```diff
@@ -537,6 +537,28 @@ extern "C" bool Neon_Mem_SameHandle(v8::Local<v8::Value> v1, v8::Local<v8::Value
...
+  Nan::Persistent<v8::Value>* original = static_cast<Nan::Persistent<v8::Value>*>(persistent);
+  Nan::Persistent<v8::Value>* copy = new Nan::Persistent<v8::Value>();
+  copy->Reset(*original);
+  return static_cast<void*>(copy);
+}
+
+extern "C" void Neon_Mem_DeletePersistent(void *persistent) {
+  Nan::Persistent<v8::Value>* p = static_cast<Nan::Persistent<v8::Value>*>(persistent);
+  p->Reset();
+  delete p;
```
bnoordhuis: The destructor also calls Reset().  Not harmful, just redundant.

crates/neon-runtime/src/neon.cc
```diff
@@ -537,6 +537,28 @@ extern "C" bool Neon_Mem_SameHandle(v8::Local<v8::Value> v1, v8::Local<v8::Value
   return v1 == v2;
 }
 
+extern "C" void *Neon_Mem_NewPersistent(v8::Local<v8::Value> val) {
+  return static_cast<void*>(new Nan::Persistent<v8::Value>(val));
```
bnoordhuis: The static_cast isn't necessary here.

crates/neon-runtime/src/neon.cc
```diff
@@ -537,6 +537,28 @@ extern "C" bool Neon_Mem_SameHandle(v8::Local<v8::Value> v1, v8::Local<v8::Value
...
+extern "C" void Neon_Mem_New(v8::Local<v8::Value> *out, void *persistent) {
+  Nan::Persistent<v8::Value>* p = static_cast<Nan::Persistent<v8::Value>*>(persistent);
+  *out = Nan::New(*p);
+}
+
+extern "C" void *Neon_Mem_ClonePersistent(void *persistent) {
+  Nan::Persistent<v8::Value>* original = static_cast<Nan::Persistent<v8::Value>*>(persistent);
+  Nan::Persistent<v8::Value>* copy = new Nan::Persistent<v8::Value>();
+  copy->Reset(*original);
+  return static_cast<void*>(copy);
```
bnoordhuis: Ditto.

crates/neon-runtime/src/neon.cc
abrodersen: Thanks, the documentation does not make this clear.

abrodersen:
I have created an RFC for this pull request: neon-bindings/rfcs#13

rebo:
Hi,

I have created an example app showing how this pull request's PersistentHandles can be used to reference Node.js objects in callbacks from a Neon Task.

A MainBackgroundTask signals a DispatchTask to return data from Node.js when it needs it.

https://github.com/rebo/async-neon-example

dherman:
This was a great start to the exploration of the design space for doing background work in arbitrary Rust threads! It looks like neon-bindings/rfcs#25 ended up being the sweet spot in the design space (thanks to N-API paving the way with their "thread-safe callbacks" design). Thank you for this contribution and for getting the conversation going!

PR #291, opened 2018-01-02
Use `node::MakeCallback` instead of calling a method directly on Task callbacks

### LINKED ISSUE #289: Unable to resolve promise from a Task callback
## Description

When attempting to convert a Task callback to a promise, calling `resolve` or `reject` does not fulfill the promise until the process is killed.

## Example

```js
// This promise never resolves even though `resolve` is called
new Promise((resolve, reject) => (
  addon.perform_async_task((err, res) => {
    if (err) {
      reject(err);
    } else {
      resolve(res);
    }
  })
)).then(() => console.log('Never Called'));

// If you leave the context of the callback, it works
new Promise((resolve, reject) => (
  addon.perform_async_task((err, res) => {
    if (err) {
      setImmediate(reject, err);
    } else {
      setImmediate(resolve, res);
    }
  })
)).then(() => console.log('Gets called'));
```

## Expected

Promise fulfills regardless.

## Actual

Promise only fulfills if resolved or rejected from another context.

## Questions

1. Is this only impacting promises?
2. Does it impact userland promises or only native promises?
3. Is this more generally related to scheduling async work from a callback?

kjvalencik:
Also, fixes memory leak in Task API.

* Ensures `nextTick` is processed
* Captures the error domain
* Fixes #289

This is how handles callbacks on async tasks.

1. [`Nan::Callback` created to queue task](https://github.com/nodejs/nan/blob/b2455ac52e97568456089c5658bbd64a016112b8/examples/async_pi_estimate/async.cc#L61)
1. [`Nan::Callback` tracked by `AsyncWorker`](https://github.com/nodejs/nan/blob/b2455ac52e97568456089c5658bbd64a016112b8/nan.h#L1506)
1. [`node::MakeCallback` used to execute `nan::Callback`](https://github.com/nodejs/nan/blob/b2455ac52e97568456089c5658bbd64a016112b8/nan.h#L1478-L1484)

### DIFF
--- a/crates/neon-runtime/src/neon_task.h
+++ b/crates/neon-runtime/src/neon_task.h
@@ -59,7 +59,9 @@ public:
     }
 
     v8::Local<v8::Function> callback = v8::Local<v8::Function>::New(isolate_, callback_);
-    callback->Call(context, v8::Null(isolate_), 2, argv);
+    node::MakeCallback(isolate_, context->Global(), callback, 2, argv);
+    callback_.Reset();
+    context_.Reset();
   }
 
   void *get_result() {
--- a/test/dynamic/lib/tasks.js
+++ b/test/dynamic/lib/tasks.js
@@ -27,4 +27,16 @@ describe('Task', function() {
       }
     });
   });
+
+  it('executes microtasks after callback', function () {
+    return new Promise((resolve, reject) => {
+      addon.perform_async_task((err, res) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve(res);
+        }
+      });
+    });
+  });
 });


crates/neon-runtime/src/neon_task.h
```diff
@@ -59,7 +59,9 @@ class Task {
     }
 
     v8::Local<v8::Function> callback = v8::Local<v8::Function>::New(isolate_, callback_);
-    callback->Call(context, v8::Null(isolate_), 2, argv);
+    node::MakeCallback(isolate_, context->Global(), callback, 2, argv);
+    callback_.Reset();
+    context_.Reset();
```
kjvalencik: These missing `Reset` on the `Persistent` objects were a memory leak in the existing code.

bnoordhuis: The changes look okay to me and MakeCallback() is the right method to call but apropos those resource leaks, their destructors should clean out the references.

Are you saying that isn't happening or is there something that is holding onto the Task longer than necessary?

kjvalencik: I'm not sure their destructors clean-out the references. They are wrapped in in `Nan` with a destructor added to call this.

You can observe the leak by taking a heap dump, calling an async API a bunch, forcing a GC, and then taking another heapdump.

kjvalencik: https://github.com/nodejs/nan/blob/b2455ac52e97568456089c5658bbd64a016112b8/nan.h#L1383-L1385

bnoordhuis: Oh, I think I see the issue now - those `v8::Persistent` members should have been of type `v8::Global` or `v8::Persistent<T,M>` with `M::kResetInDestructor = true`.  Might be good to check if the code base has more of these issues.

kjvalencik:
@dherman Is there anything I can provide to make this easier to merge?

We are currently using my fork in production successfully, but it would be nice to switch to upstream.

Cheers!

PR #303, opened 2018-04-12
Updated Task trait and new Worker trait for async work

tptee:
See RFC: https://github.com/neon-bindings/rfcs/pull/15

TODO:
- [ ] Add tests for worker failure cases
- [ ] Fix build script (`node-gyp` isn't always installed globally)


tptee:
Welp, `bindgen`-generated bindings are platform and architecture specific...don't know what I expected. We have two options here:
- Add `bindgen` as a build step. This requires users to have `libclang` installed on their system–annoying and probably not already on CI images
- Pre-bake and conditionally compile bindings for all of Node's supported architectures.

Thoughts @dherman ?

tptee:
Another idea–put this PR behind an `async` Cargo feature and only run the `bindgen` build step if the user opts in!

dherman:
Thanks for exploring this design space, @tptee! It looks like neon-bindings/rfcs#25 is the direction we're heading, but we wouldn't have gotten there without all we'd learned from this exploration.

Issue #305, opened 2018-05-02
neon build fails with node.js 10


hyperfekt:
See nodejs/nan#766.

mhsjlw:
I'm guessing that if https://github.com/neon-bindings/neon/blob/4e2273c14de6820d81cb377a44e5634b88f69174/crates/neon-runtime/package.json#L11 is bumped to `2.10.0` It'll work. I'll give it a try.

mhsjlw:
Alright, it appears that 

https://github.com/neon-bindings/neon/blob/4e2273c14de6820d81cb377a44e5634b88f69174/crates/neon-runtime/src/neon.cc#L30-L32

this function fails to build due to 

```
../src/neon.cc:32:16: error: no member named 'Callee' in 'v8::FunctionCallbackInfo<v8::Value>'
  *out = info->Callee();
```

So it looks like it's been removed from V8? I'll keep investigating, but once this builds I'm guessing it will work.

mhsjlw:
It does appear that it was marked as deprecated:

https://codereview.chromium.org/1582793004/patch/1/10001

But has since been removed:

https://chromium-review.googlesource.com/c/v8/v8/+/741600

I'll look at the possibility of moving to `->Data()`

mhsjlw:
A PR has been opened with my changes

PR #310, opened 2018-05-13
Support Node 10

mhsjlw:
Alright, a few things:

1. Pretty much no idea what I'm doing, I tried to look for developer documentation but found none
2. Haven't tested this with any other version than Node v10.1.0
3. I ran the test suite and everything passed

On paper, it looks like it works. I'm not entirely sure if there is a case where you are `::Cast()`ing a non-function.

Please let me know if I need to and how to implement backwards compatibility with this, also see #305 for a little more background information and how I got to this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rust:
 
 env:
   matrix:
+    - TRAVIS_NODE_VERSION="10"
     - TRAVIS_NODE_VERSION="8"
     - TRAVIS_NODE_VERSION="6"
     - TRAVIS_NODE_VERSION="4"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,9 @@ environment:
       NODEJS_VERSION: "8"
       RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
     - PLATFORM: x64
+      NODEJS_VERSION: "10"
+      RUST_TOOLCHAIN: stable-x86_64-pc-windows-msvc
+    - PLATFORM: x64
       NODEJS_VERSION: "4"
       RUST_TOOLCHAIN: beta-x86_64-pc-windows-msvc
     - PLATFORM: x64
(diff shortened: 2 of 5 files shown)


mhsjlw:
@pitaj Good idea, it has been added. This is a quick way for me to ensure that my changes work across versions as well (I doubt it will)

pitaj:
@dherman any chance of looking at this? It doesn't require an RFC does it?

kjvalencik:
I don't think this is safe because the data might not actually be a callee. Should this code be removed or at least feature flagged for future removal?

mhsjlw:
Based on this documentation: https://github.com/neon-bindings/neon/blob/b83ab66a442cc2b26fae7dc9227919be1d11a705/crates/neon-runtime/src/call.rs#L28, when would it ever not be the callee?

Additionally, I do not believe that this is the optimal solution, it just just the solution that I immediately know of that fixes this issue. Suggestions are appreciated, removal might not be needed.

dherman:
@kjvalencik I've been googling around and haven't found a ton of recent discussions of this removal from the V8 API. In 2015 there was [this conversation](https://groups.google.com/forum/#!topic/v8-users/n5UFkh26dX4), but then they seemed like they were backtracking because of Node plugins, so I'm not sure what led them to decide it was OK to remove after all. Looking at the [V8 API changes doc](https://docs.google.com/document/d/1g8JFi8T_oAE_7uAri7Njtig7fKaPDfotU6huOa1alds/edit) just lists it as a removal but without links to any discussion, and the [commit](https://chromium-review.googlesource.com/c/v8/v8/+/741600) just has a one-line mention of why it's helpful for the V8 implementation.

I wonder if anyone knows of a discussion that led people to feel this was not a commonly-used feature. I'm leaning towards just removing it from Neon since that's what V8 is doing anyway, and it's not supported in N-API.

dherman:
Thank you for the informative comment, as always!

> I doubt it's necessary, though. Node's citgm project has pretty good coverage of native modules and none that I know of broke due the removal of the `Callee()` method.

Yeah, I think we should avoid providing primitives beyond what V8 and N-API provide, so I'm going to propose we just cut `callee()` for 0.2.

dherman:
@mhsjlw Thank you so much for getting the ball rolling with this PR! I ended up submitting the separate PR #314, but I wouldn't have gotten there without your contribution here!

mhsjlw:
Great to have helped! Is there a release on the horizon?

Issue #260, opened 2017-10-19
Working examples, where to put these?


rm-rf-etc:
Hi, I'd like to submit some basic examples to help people who are trying out this module and Rust for the first time. I hit quite a few frustrating things when trying to get each of these working. I'd like to submit a PR somewhere but don't know where you'd like this to go. Please advise.

```
#[macro_use]
extern crate neon;

use neon::vm::{Call, JsResult};
use neon::js::{
	Object, // <- https://github.com/neon-bindings/neon/issues/57
	JsFunction,
	JsObject,
	JsArray,
	JsBoolean,
	JsInteger,
	JsNull,
	JsNumber,
	JsString,
	JsUndefined };


/*
	// If you need to inspect the value of an expression:
	let () = JsBoolean::new();
*/

// Use _ to squelch warnings, or any name starting with _
fn get_null_sync(_: Call) -> JsResult<JsNull> {
	Ok(JsNull::new())
}

// Use _ to squelch warnings, or any name starting with _
fn get_undefined_sync(_: Call) -> JsResult<JsUndefined> {
	Ok(JsUndefined::new())
}

fn get_number_sync(call: Call) -> JsResult<JsNumber> {
	Ok(JsNumber::new(call.scope, 5f64))
}

fn get_integer_sync(call: Call) -> JsResult<JsInteger> {
	Ok(JsInteger::new(call.scope, 5i32))
}

fn get_string_sync(call: Call) -> JsResult<JsString> {
	Ok(JsString::new(call.scope, "my string").unwrap())
}

fn get_boolean_sync(call: Call) -> JsResult<JsBoolean> {
	Ok(JsBoolean::new(call.scope, false))
}

fn get_array_sync(call: Call) -> JsResult<JsArray> {

	let scope = call.scope;
	let array = JsArray::new(scope, 3);

	try!(array.set(0, JsInteger::new(scope, 1)));
	try!(array.set(1, JsInteger::new(scope, 2)));
	try!(array.set(2, JsInteger::new(scope, 3)));

	Ok(array)
}

fn get_object_sync(call: Call) -> JsResult<JsObject> {

	let scope = call.scope;
	let object = JsObject::new(scope);

	try!(object.set("prop1", JsInteger::new(scope, 1)));
	try!(object.set("prop2", JsInteger::new(scope, 2)));
	try!(object.set("prop3", JsInteger::new(scope, 3)));

	Ok(object)
}

[...]

PR #333, opened 2018-07-12
added some examples

TomPridham:
I was banging my head against the wall for a few days trying to get some pretty basic stuff working because there isn't very much documentation right now. I added a bunch of examples for some common use cases.

I took some examples from: https://github.com/neon-bindings/neon/issues/260, and some helper things from: https://github.com/dfrankland/node-eyeliner/blob/master/native/src/js_object_utils.rs

I'm not sure if I structured those examples right, but `cargo test` doesn't complain, so I feel like I did.

I also have a question. I spent a couple hours trying to figure out what I was doing wrong with the `interacting_with_js_arrays` example. it turned out that it was because I the array I was passing in was `[1, 2, 3]`. All I changed was adding some decimal bits after and it suddenly started working. Is `Variant::Number` not the right thing for integers? I know there was a `JsInteger` type, but I thought that was removed.


kjvalencik:
@TomPridham did you happen to see the dynamic tests? It's a bit dense, but provides examples of most things that you could do. Could it be worthwhile to clean those up with better documentation and a bit more segmentation to act as examples?

https://github.com/neon-bindings/neon/tree/master/test/dynamic

TomPridham:
I had looked at them briefly. I'm coming from a JS background and they didn't seem super accessible or relevant. I also wouldn't have thought to look there first off. I only ended up looking in there after having done a bunch of other searches and not finding anything.

I also feel like having tests act as examples could get sort of messy. I could add some documentation, but I think having the examples is more valuable.

dherman:
@TomPridham Thank you for these! This is really good feedback about the kinds of things a JS programmer can get tripped on coming to Neon and Rust for the first time. And I totally agree more examples can be a great way to help newcomers explore basic use cases.

If I make a few suggestions, would you be up for making a few edits before we merge?

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
+#[macro_use]
+extern crate neon;
+
+// thanks to @rm-rf-etc for these: https://github.com/neon-bindings/neon/issues/260
```
dherman: (Yeah, this is such an awful foot-gun in Neon 0.1. We're fixing this in 0.2, which will be shipping very soon. In the meantime this is good, and we can update this example and get rid of the comment once it's no longer necessary. That change won't invalidate the rest of this example project, which will still be useful!)

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+  JsBoolean,
+  JsInteger,
+  JsNull,
+  JsNumber,
+  JsString,
+  JsUndefined};
+
+/*
+	// If you need to inspect the value of an expression:
+	let () = JsBoolean::new();
```
dherman: This comment is a little confusing: did you mean `()` to be a stand-in for "whatever pattern you want to match"? As is, `()` is real Rust syntax, and this statement as you've written it would parse properly but cause a type error. That seems like it could trip someone up who's trying to understand the API. Maybe something like `let <<put your pattern here>> = JsBoolean::new();`? I guess I'm not completely sure what you're trying to get across in this comment?

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+*/
+
+// Use _ to squelch warnings, or any name starting with _
```
dherman: Nice! This kind of contextual tip is such a great way to teach newcomers as they go along.

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+
+fn get_boolean_sync(call: Call) -> JsResult<JsBoolean> {
+  Ok(JsBoolean::new(call.scope, false))
+}
+
+fn get_array_sync(call: Call) -> JsResult<JsArray> {
+  let scope = call.scope;
+  let array = JsArray::new(scope, 3);
+
+  try!(array.set(0, JsInteger::new(scope, 1)));
```
dherman: `try!` is kind of "old-fashioned Rust" style, since the `?` syntax was introduced in [Rust 1.13](https://blog.rust-lang.org/2016/11/10/Rust-1.13.html). The more idiomatic way to write these lines would be:

```rust
array.set(0, JsInteger::new(scope, 1))?;
array.set(1, JsInteger::new(scope, 2))?;
array.set(2, JsInteger::new(scope, 3))?;
```

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+  try!(array.set(2, JsInteger::new(scope, 3)));
+
+  Ok(array)
+}
+
+fn get_object_sync(call: Call) -> JsResult<JsObject> {
+  let scope = call.scope;
+  let object = JsObject::new(scope);
+
+  try!(object.set("prop1", JsInteger::new(scope, 1)));
```
dherman: Again, `?` would be better style than `try!`.

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+  try!(object.set("prop3", JsInteger::new(scope, 3)));
+
+  Ok(object)
+}
+
+fn get_function_sync(call: Call) -> JsResult<JsFunction> {
+  fn func(call: Call) -> JsResult<JsInteger> {
+    Ok(JsInteger::new(call.scope, 5))
+  }
+  Ok(JsFunction::new(call.scope, func).unwrap())
```
dherman: It's not great style to use `.unwrap()`, and this is unnecessarily taking apart and then reassembling a `Result`. You can make this line simpler by just writing:

```rust
fn get_function_sync(call: Call) -> JsResult<JsFunction> {
    fn func(call: Call) -> JsResult<JsInteger> {
        Ok(JsInteger::new(call.scope, 5))
    }
    JsFunction::new(call.scope, func)
}
```

examples/basics/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+register_module!(m, {
+	try!(m.export("getNullSync", get_null_sync));
```
dherman: Also here I'd replace `try!` with `?`.

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+#[macro_use]
+extern crate neon;
+
+use neon::vm::{Call, JsResult};
+use neon::js::{Variant, JsNumber, JsArray};
+
+// https://github.com/dfrankland/node-eyeliner/blob/master/native/src/js_object_utils.rs
+pub fn js_value_to_float(js_number: Variant) -> f64 {
```
dherman: I wouldn't recommend this way of doing a type check, and the `Variant` type will be going away. I'll show the better style below...

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+  match js_number {
+    Variant::Number(number) => number.value(),
+    _ => panic!("heck"),
+  }
+}
+
+fn accepts_js_arrays(call: Call) -> JsResult<JsNumber> {
+  let js_arr_handle = call.arguments.get(call.scope, 0).unwrap();
```
dherman: Again, let's not recommend `.unwrap()`. It's better style and more concise to use `?`:

```rust
let js_arr_handle = call.arguments.get(call.scope, 0)?;
```

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+  let js_arr = js_arr_handle.downcast::<JsArray>().unwrap().to_vec(call.scope)?;
```
dherman: Use `?` instead of `.unwrap()` here too.

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+  let vec_of_numbers: Vec<_> = js_arr.iter().map(move|js_value| {
```
dherman: This is a great use case! Again let's try to avoid `Variant` and panics. The idiomatic Rust way to do this would be:

```rust
let vec_of_numbers: Vec<_> = js_arr.iter().map(|js_value| {
    let js_number = js_value::check::<JsNumber>()?;
    Ok(js_number.value())
}).collect()?;
```

Don't freak out if that seemed magical! It might be worth a comment. This is a good Rust idiom to learn but it's definitely a little dauntingly implicit at first. The short version of the story is, you can map with a function that returns a `Result` type and get a `Result<Vec<...>>` with nothing more than `.map(...).collect()`. Rust's iteration protocol is powerful enough to do the right thing.

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+    let js_number = js_value.variant();
+    js_value_to_float(js_number)
+  }).collect();
+  let sum = vec_of_numbers.iter().sum();
+  Ok(JsNumber::new(call.scope, sum))
+}
+
+register_module!(m, {
+  try!(m.export("accepts_js_arrays", accepts_js_arrays));
```
dherman: One more `try!` to replace with `?`. :)

examples/interacting_with_js_arrays/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+	Ok(())
+});
+
+// in lib/index.js
```
dherman: Is there a reason to write this comment here, instead of just having that code in the JS file?

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#[macro_use]
+extern crate neon;
+
+use neon::vm::{Call, JsResult};
+use neon::js::{Variant, JsString, JsObject};
+
+// https://github.com/dfrankland/node-eyeliner/blob/master/native/src/js_object_utils.rs
+pub fn js_value_to_string(js_string: Variant) -> String {
```
dherman: Similar to the other example, let's get rid of this function and just use `check()` below.

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+  match js_string {
+    Variant::String(string) => string.value(),
+    _ => panic!("heck"),
+  }
+}
+
+fn accepts_js_objs(call: Call) -> JsResult<JsString> {
+  let js_obj_handle = call.arguments.get(call.scope, 0).unwrap();
+  let js_obj = js_obj_handle.downcast::<JsObject>().unwrap();
```
dherman:
```rust
let js_obj = js_obj_handle.check::<JsObject>()?;
```

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+// https://github.com/dfrankland/node-eyeliner/blob/master/native/src/js_object_utils.rs
+pub fn js_value_to_string(js_string: Variant) -> String {
+  match js_string {
+    Variant::String(string) => string.value(),
+    _ => panic!("heck"),
+  }
+}
+
+fn accepts_js_objs(call: Call) -> JsResult<JsString> {
+  let js_obj_handle = call.arguments.get(call.scope, 0).unwrap();
```
dherman: Replace `.unwrap()` with `?`.

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+  let js_obj = js_obj_handle.downcast::<JsObject>().unwrap();
+
+  let prop_js = js_obj.get(call.scope, "myProp").unwrap().variant();
+  let prop_text: String = js_value_to_string(prop_js);
```
dherman:
```rust
let prop_js = js_obj.get(call.scope, "myProp")?;
let prop_text: String = prop_js::check::<JsString>()?.value();
```

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+  Ok(JsString::new(call.scope, prop_text.as_str()).unwrap())
```
dherman: This is another case of unnecessarily taking apart and reassembling a Result. Instead you can just write:

```rust
JsString::new(call.scope, prop_text.as_str())
```

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+register_module!(m, {
+  try!(m.export("accepts_js_objs", accepts_js_objs));
```
dherman: One more `try!` to replace with `?`.

examples/interacting_with_js_objects/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+	Ok(())
+});
+
+// in lib/index.js
```
dherman: Also here, should we drop this comment and just have the JS in the JS source file?

examples/basics/src/lib.rs
TomPridham: I'm not sure what was meant by that. I copied most(all?) of that from #260. I can delete it

examples/interacting_with_js_arrays/src/lib.rs
TomPridham: I originally did it this way because I wasn't sure what structure would work in the examples and was trying to keep it minimal. I moved everything to mirror the output of `neon new` since I think that makes more sense.

TomPridham:
I think I addressed everything. I moved the examples to mirror the output of `neon new`. let me know if anything else needs work

amilajack:
Examples for neon have been congregated into the [amilajack/neon-examples](https://github.com/amilajack/neon-examples). I am working with @dherman on moving those examples to the @neon-bindings org. Now that we have a place for examples this PR should be closed. Please PR the [neon-examples repo](https://github.com/amilajack/neon-examples) if you have any examples you would like to add

TomPridham:
👍

PR #335, opened 2018-07-16
Task trait should be marked 'static. Fixes #288.

### LINKED ISSUE #288: Task API allows unsafe borrows
## Description

The Task API allows unsafe references enabling bugs like use after free.

## Example

```rust
struct BorrowedTask<'b>(&'b str);
```

https://github.com/kjvalencik/neon/tree/task-lifetime-issue
https://github.com/kjvalencik/neon/commit/00216ece7f1e6d7d63153ced608e2e08b429fa42

## Expected

Compile error.
> borrowed value does not live long enough

## Actual

Reference to string points at already freed memory.

```
  Task
    ✓ completes a successful task
    ✓ completes a failing task
    ✓ completes with owned task
    1) completes with borrowed task
    ✓ completes with static borrowed task
    2) completes with static shortened borrowed task
    ✓ completes with forgot borrowed task


  5 passing (9ms)
  2 failing

  1) Task completes with borrowed task:
     AssertionError: expected 'Hello, \u0000\u0000\u0000\u0000\u0000!' to equal 'Hello, World!'
      at addonAsync.perform_borrowed_task.then.res (lib/tasks.js:58:27)
      at <anonymous>

  2) Task completes with static shortened borrowed task:
     AssertionError: expected 'Hello, `��\u0003\u0000!' to equal 'Hello, World!'
      at addonAsync.perform_borrowed_task_static_short.then.res (lib/tasks.js:68:27)
      at <anonymous>
```

kjvalencik:
The same applies here.

> The 'static constraint means that the closure and its return value must have a lifetime of the whole program execution. The reason for this is that threads can detach and outlive the lifetime they have been created in. Indeed if the thread, and by extension its return value, can outlive their caller, we need to make sure that they will be valid afterwards, and since we can't know when it will return we need to have them valid as long as possible, that is until the end of the program, hence the 'static lifetime.

https://doc.rust-lang.org/std/thread/fn.spawn.html

### DIFF
--- a/src/thread/mod.rs
+++ b/src/thread/mod.rs
@@ -12,12 +12,12 @@ use neon_runtime;
 use neon_runtime::raw;
 
 /// A Rust task that can be executed in a background thread.
-pub trait Task: Send + Sized {
+pub trait Task: Send + Sized + 'static {
     /// The task's result type, which is sent back to the main thread to communicate a successful result back to JavaScript.
-    type Output: Send;
+    type Output: Send + 'static;
 
     /// The task's error type, which is sent back to the main thread to communicate a task failure back to JavaScript.
-    type Error: Send;
+    type Error: Send + 'static;
 
     /// The type of JavaScript value that gets produced to the asynchronous callback on the main thread after the task is completed.
     type JsEvent: Value;


src/thread/mod.rs
```diff
@@ -12,7 +12,7 @@ use neon_runtime;
 use neon_runtime::raw;
 
 /// A Rust task that can be executed in a background thread.
-pub trait Task: Send + Sized {
+pub trait Task: Send + Sized + 'static {
     /// The task's result type, which is sent back to the main thread to communicate a successful result back to JavaScript.
     type Output: Send;
```
kjvalencik: Should this be `'static` as well?

dherman: I think so too. Take a look at `std::thread::spawn`'s signature:

```rust
pub fn spawn<F, T>(f: F) -> JoinHandle<T> where
    F: FnOnce() -> T,
    F: Send + 'static,
    T: Send + 'static, 
```
The return type of the function is required to be `Send + 'static`, and that's the analogous type to `Output`.

kjvalencik:
@dherman Updated.

dherman:
This is fantastic, thank you so much.

PR #342, opened 2018-07-23
Skip writing artifacts.json if it hasn't changed to prevent rebuilds

kjvalencik:


kjvalencik:
This should be ignored, but it's not. This change worked, but I'm not sure if it's best. Still testing.

PR #343, opened 2018-07-23
Add artifacts.json and index.node to exclude

kjvalencik:
This can be removed if workspaces are implemented and the .gitignore file is accurately picked up (https://github.com/neon-bindings/rfcs/pull/4).


dherman:
Excellent, thank you!

PR #347, opened 2018-07-31
Experimental version of API that replaces `Handle<'a, V>` with `&V`

dherman:
This experimental PR gets rid of the somewhat unpleasant `Handle<'a, V>` smart pointer type and replaces it with a simple `&'a V`. Every context carries an internal arena for allocating handles so that the APIs can return references to those arena-allocated handles.

As a simple example:

```rust
fn forty_two(mut cx: FunctionContext) -> NeonResult<&JsNumber> {
    Ok(cx.number(42))
}
```


src/context/internal.rs
```diff
@@ -52,7 +55,68 @@ pub struct ScopeMetadata {
...
+    pub handle_scope: &'a mut R,
+    pub handle_arena: HandleArena<'a>
+}
+
+const CHUNK_SIZE: usize = 16;
+
+type Chunk = [raw::Local; CHUNK_SIZE];
+
+pub struct HandleArena<'a> {
+    chunks: VecDeque<Box<Chunk>>,
```
matklad: `Vec` should be enough here I think?

dherman: I think you're right: since these are arenas they only deallocate all at once, so `pop_front` will never be used.

amilajack:
🎉 🎉 🎉 Awesome work Dave!

src/context/internal.rs
dherman: I ended up using an arena crate from crates.io

dherman:
Looks like something is wrong specifically with Node 6. Gotta go to bed but I'll investigate that next.

kjvalencik:
FYI, Node 6 is end of life in April.

dherman:
Ugh, and now I can't even build it on my mac to investigate, because [Node 6 has been left in an uncompilable state on macOS Mojave](https://github.com/nodejs/node/issues/24648#issuecomment-442392756).

dherman:
> Ugh, and now I can't even build it on my mac to investigate

OK but I can still build on Linux. Investigating…

dherman:
OK I believe this is the issue: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32897 — I'm talking with the handlebars maintainers and I'll track their fix as it lands.

dherman:
I merged #396 to fix the typescript issue, and I'm rebasing and trying again…

dherman:
@kjvalencik This PR is ready for experimentation! If you have any time and are still interested, it'd be super interesting to see an experiment or two with real code, both for comparing readability and to ensure there aren't noticeable performance issues.

kjvalencik:
@dherman Yes, I can do that. I'll take notes along the way similar to what I did for the `v0.2`. I do have a decent amount of clean-up I would like to do prior to trying it out (upgrade dependencies, switch to Rust 2018).

dherman:
This is out of date, and @kjvalencik's experience experimenting with this change was pretty negative. I would love to know more about it at some point, but this isn't very high priority right now so I'm going to close the PR. We can always look at it for reference in future if we're interested in reviving the idea.

kjvalencik:
Unfortunately, I deleted the branch that tested this API. If I recall correctly, this change forced users to define lifetimes in many places that it was previously not necessary. It made it very difficult for a new Rust user to try out Neon without immediately being thrown in the deep end.

Issue #197, opened 2017-03-14
Ability to call Javascript methods from Rust threads


jugglingcats:
As discussed on Slack: https://rust-bindings.slack.com/archives/neon/p1489508576491688.

The basic idea is that a Rust addon can create a continuous background thread that notifies Javascript when things happen (event style).

Psuedo Javascript:
```
addin.run(function() { 
  console.log("got the event!")
}
```

Psuedo Rust:
```
fn run(call: Call) -> JsResult<JsString> {
    let scope = call.scope;
    let f = call.arguments.require(scope, 0)?.check::<JsFunction>()?;

    thread::spawn(|| {
        while (true) {
            let args: Vec<Handle<JsNumber>> = vec![];
            match f.call(scope, JsNull::new(), args) {
                _ => { println!("Call to javascript done!"); }
            }
            thread::sleep(1000);
        }
    });
    Ok(JsString::new(scope, "hello node").unwrap())
}
```

If this was achievable it would be a _lot_ nicer than the C equivalent, an example of which is here: https://gist.github.com/matzoe/11082417.

I imagine there would need to be some kind of wrapper around the call, so maybe not completely transparent as shown above.

jugglingcats:
I've made a bit of progress with this but am stuck!

I've managed to invoke a Rust function using `uv_async_send` (ie. on the libuv default thread). In this method/context it's safe to get the current isolate scope and invoke the Javascript callback, but I don't really understand how to encapsulate this information from the Neon `Call` object and make use of it.

Am struggling both with Rust basics and the idiomatic Neon way this should work. Any advice/pointers gratefully received. I can't even tell if I'm close or not!!

My working Rust project (so far) is here:
https://github.com/jugglingcats/neon-uv/tree/iter1

My working C++ project (which does invoke the js callback) is here:
https://github.com/jugglingcats/node-uv/tree/iter1

The C++ project shows the basic approach to converting a `Local` to `Persistent` in v8 in order to preserve between the main addon method and the callback, although I don't claim to understand this fully. We need to do something similar in Rust-land. C++ snippet:

```
	v8::Local<v8::Function> func = v8::Local<v8::Function>::Cast(args[0]);
	auto savedFunc = new v8::Persistent<v8::Function>();
	savedFunc->Reset(isolate, func);
	handle->data = savedFunc;
```
and (in callback):
```
	v8::Isolate *isolate = Isolate::GetCurrent();
	HandleScope scope(isolate);

[...]

PR #375, opened 2018-11-29
Add ThreadSafeCallback helper to schedule callbacks on the main thread

geovie:
Hi,
this adds a little helper to call a JavaScript function from any rust thread as requested in #197.
~~I'm not sure if this will need a RFC, but I can write one if requested.~~
[RFC](https://github.com/neon-bindings/rfcs/pull/25)

Feedback welcome, especially about the v8 scope setup.

Inspired by [mika-fischer/napi-thread-safe-callback](https://github.com/mika-fischer/napi-thread-safe-callback)


maackle:
I don't have the depth to comment on the viability of this, but I would love to have this functionality! I'm currently building something kind of complicated using `Task` because callbacks in threads are not possible.

amilajack:
@geovie sorry for the delay on this. We'll review this soon

qm3ster:
Oh wow, this is exactly what I need.
I've been trying to get "emitting" to work, even with help on the JS side, but I can neither call a callback asynchronously without terminating a `Task`, nor pass a `tokio::sync::mpsc::Receiver` on the "class" struct into a new `Task` because of lifetimes.

jrd-rocks:
Thanks for this. I needed some sane way to get callbacks working. This has been a life saver!

x1957:
any updates?

dherman:
@x1957 I'm discussing with @aep on Slack about updating this PR, and over on the RFC PR I've asked @geovie if I can help finish up the RFC. Would love to get this landed!

dignifiedquire:
I tried the new branch, and it breaks unfortunately, using `setTimeout` + a loop in the thread, seems to results in the callback from `setTimeout` to never be called.

dignifiedquire:
you can see the issue here: https://github.com/deltachat/deltachat-neon/blob/master/test.js#L14

dignifiedquire:
sorry, it was my mistake, I was putting too much work into the `cb.schedule` closure, the PR in #429 does work nicely for me

crates/neon-runtime/native/src/neon.cc
```diff
@@ -8,6 +8,7 @@
 #include "neon_string.h"
 #include "neon_class_metadata.h"
 #include "neon_task.h"
+#include "neon_threadsafe_cb.h"
```
geovie: thread safe callback is used for the internal API as it's not event handler specific

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct EventHandler(Arc<EventHandlerInner>);
+
+impl EventHandler {
+    pub fn new(callback: Handle<JsFunction>) -> Self {
+        TaskContext::with(|mut cx: TaskContext| {
```
geovie: Is TaskContext correct here?

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub fn bind<T: Value>(this: Handle<T>, callback: Handle<JsFunction>) -> Self {
+        let cb = unsafe {
+            neon_runtime::threadsafecb::new(this.to_raw(), callback.to_raw())
+        };
+        EventHandler(Arc::new(EventHandlerInner(cb)))
+    }
+
+    pub fn schedule<T, F>(&self, arg_cb: F)
+        where T: Value,
+              F: for<'a> FnOnce(&mut TaskContext<'a>) -> Vec<Handle<'a, T>>,
```
geovie: I'd love to have `IntoIterator<...>` instead of `Vec<..>`, but I couldn't get it to work with the lifetimes....

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+unsafe extern "C" fn handle_callback<T, F>(this: raw::Local, func: raw::Local, callback: *mut c_void)
+    where T: Value, F: for<'a> FnOnce(&mut TaskContext<'a>) -> Vec<Handle<'a, T>>, F: Send + 'static {
+    TaskContext::with(|mut cx: TaskContext| {
+        let this = JsValue::new_internal(this);
+        let func: Handle<JsFunction> = Handle::new_internal(JsFunction::from_raw(func));
+        let callback: Box<F> = Box::from_raw(mem::transmute(callback));
+        let args = callback(&mut cx);
+        let _result = func.call(&mut cx, this, args);
```
geovie: Here the result (return value or exception) is discarded, I'm not sure if this is okay.   
`EventHandler::schedule_with` was added to allow to use the result.

geovie:
@dherman @kjvalencik I've updated the PR with the changes from the RFC

kjvalencik:
@geovie Are you able to update this with the latest changes from the RFC? @aep also appears to be working on this. Can you coordinate? Thanks! I would :heart: to get this merged.

geovie:
@kjvalencik The PR is already up to date with the RFC, I will rebase it later. Any review is welcome

kjvalencik:
@geovie Ah, sorry I missed that update! Thanks! I'll start reviewing.

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct EventHandler(Arc<EventHandlerInner>);
+
+impl EventHandler {
+    pub fn new(callback: Handle<JsFunction>) -> Self {
+        TaskContext::with(|mut cx: TaskContext| {
+            let cb = unsafe {
```
kjvalencik: To reduce the amount of `unsafe`, could this method delegate to `bind` instead?

```rust
pub fn new(callback: Handle<JsFunction>) -> Self {
    TaskContext::with(|mut cx| Self::bind(cx.global(), callback))
}
```

src/eventhandler/mod.rs
kjvalencik: This could be a future improvement. I think to benefit from this, it would require a more holistic change to arguments in Neon.

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        EventHandler(Arc::new(EventHandlerInner(cb)))
+    }
+
+    pub fn schedule<T, F>(&self, arg_cb: F)
+        where T: Value,
+              F: for<'a> FnOnce(&mut TaskContext<'a>) -> Vec<Handle<'a, T>>,
+              F: Send + 'static {
+        let callback = Box::into_raw(Box::new(arg_cb)) as *mut c_void;
+        unsafe {
+            neon_runtime::threadsafecb::call((*self.0).0, callback, handle_callback::<T, F>);
```
kjvalencik: Similar to above, can this method delegate to `schedule_with`?

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        where F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>),
+              F: Send + 'static {
+        let callback = Box::into_raw(Box::new(arg_cb)) as *mut c_void;
+        unsafe {
+            neon_runtime::threadsafecb::call((*self.0).0, callback, handle_callback_with::<F>);
+        }
+    }
+}
+
+unsafe extern "C" fn handle_callback<T, F>(this: raw::Local, func: raw::Local, callback: *mut c_void)
```
kjvalencik: Is this necessary? Since it's a specialized version of `handle_callback_with`, could it be implemented with wrapping the `FnOnce` in another lambda?

src/eventhandler/mod.rs
geovie: Absolutely, thanks for pointing out.

kjvalencik:
@geovie Unfortunately, files got moved around in #440 to prep for N-API. Are you able to resolve the conflicts? Let me know if you would like me to help.

geovie:
@kjvalencik rebased on master, should be good once CI passes

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn bind<T: Value>(this: Handle<T>, callback: Handle<JsFunction>) -> Self {
+        let cb = unsafe {
+            neon_runtime::threadsafecb::new(this.to_raw(), callback.to_raw())
+        };
+        EventHandler(Arc::new(EventHandlerInner(cb)))
+    }
+
+    pub fn schedule<T, F>(&self, arg_cb: F)
+        where T: Value,
+              F: for<'a> FnOnce(&mut TaskContext<'a>) -> Vec<Handle<'a, T>>,
```
dherman: I love that you reused `TaskContext`! I think this suggests a rename + a backwards compatibility hack. I'm thinking we should rename `TaskContext` to `EventContext`, since both the `Task` and `EventHandler` API are using the same concept here.

Then for compatibility, we can add `type TaskContext<'a> = EventContext<'a>;`. What do you think?

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+            let _result = callback.call(cx, this, args);
+        })
+    }
+
+    pub fn schedule_with<F>(&self, arg_cb: F)
+        where F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>),
+              F: Send + 'static {
+        let callback = Box::into_raw(Box::new(arg_cb)) as *mut c_void;
+        unsafe {
+            neon_runtime::threadsafecb::call((*self.0).0, callback, handle_callback::<F>);
```
dherman: For clarity and consistency, I think we should call the runtime operation `schedule` instead of `call`.

crates/neon-sys/native/src/neon_threadsafe_cb.h
```diff
@@ -0,0 +1,99 @@
...
+
+        void close() {
+            // close is called when the rust struct is dropped
+            // this guarantees that it's called only once and
+            // that no other method (call) will be called after it.
+            close_ = true;
+            uv_async_send(&async_);
+        }
+
+        void async_call() {
```
dherman: We could maybe call this `complete` and the static method could be `complete_callback`, for consistency with the Task API.

crates/neon-sys/native/src/neon_threadsafe_cb.h
```diff
@@ -0,0 +1,99 @@
...
+            async_.data = this;
+            uv_async_init(uv_default_loop(), &async_, async_callback);
+            // Save the this argument and the callback to be invoked.
+            self_.Reset(isolate, self);
+            callback_.Reset(isolate, callback);
+            // Save the context (aka realm) to be used when invoking the callback.
+            context_.Reset(isolate, isolate->GetCurrentContext());
+        }
+
+        void call(void *rust_callback, Neon_ThreadSafeCallbackHandler handler) {
```
dherman: I would recommend we call this `schedule` for clarity and consistency with the public API we're implementing.

crates/neon-sys/native/src/neon_threadsafe_cb.h
```diff
@@ -0,0 +1,99 @@
...
+            {
+                std::lock_guard<std::mutex> lock(mutex_);
```
dherman: I think I follow the logic here and it looks pretty tight. My one suggestion would be that we could probably implement more of this logic in pure Rust, which would be more maintainable and easier to understand. That said, I don't think that blocks landing this. We could file a followup issue for migrating the thread-safe handler queue to Rust.

crates/neon-sys/native/src/neon_threadsafe_cb.h
```diff
@@ -0,0 +1,99 @@
...
+            // invoking the callback, and use the context (i.e. realm) we saved with the task.
+            v8::Isolate::Scope isolate_scope(isolate_);
+            v8::HandleScope handle_scope(isolate_);
+            v8::Local<v8::Context> context = v8::Local<v8::Context>::New(isolate_, context_);
+            v8::Context::Scope context_scope(context);
+
+            v8::Local<v8::Value> self = v8::Local<v8::Value>::New(isolate_, self_);
+            v8::Local<v8::Function> callback = v8::Local<v8::Function>::New(isolate_, callback_);
+
+            while (true) {
```
dherman: This loop is the hardest part of the logic to understand. If you could possibly comment it with some explanation, that would help. Also, if you derived this from someone else's code (whether it's the N-API thread-safe callback implementation, or some textbook algorithm), it'd be good to cite the source just so maintainers know where to go in order to understand it better, fix bugs in the future, etc.

src/eventhandler/mod.rs
kjvalencik: I don't think it's impossible, but the reference to the global context would need to be lazily created as part of the callback and not _here_.

With that said, I agree that it follows the existing patterns in neon better to add it to the `Context` trait.

I also support deferring the convenience method to a separate PR.

src/eventhandler/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+}
+
+unsafe extern "C" fn handle_callback<F>(this: raw::Local, func: raw::Local, callback: *mut c_void)
+    where F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>), F: Send + 'static {
+    TaskContext::with(|mut cx: TaskContext| {
```
geovie: @dherman is this usage of `TaskContext::with`correct? If yes I think I can rewrite it so we can still provide both `new` and `bind`

dherman: Yes, it looks correct. Great point! Instead of grabbing the global object beforehand, we can grab it when the callback is invoked. 👍

geovie:
@dherman Addressed your comments, should be good to go now 👍

bochaco:
One more to the list of persons awaiting for this! and happy to see it almost there! awesome! I'll be giving it a try myself from this branch

bochaco:
@geovie , one question, is there a way to return value from the callback? it seems this PR doesn't support that, is that right? if so, is this something that could potentially be added?

geovie:
@bochaco You can return something from the javascript callback and get it on the rust side.
Take a look at the tests especially https://github.com/neon-bindings/neon/pull/375/files#diff-5b76efb8dfb111ab1ba34426ecf0abfeR115

bochaco:
Thanks a lot @geovie , I guess I cannot get the value out of the `cb.subscribe_with` call right? that's async and no way to wait for a result from it, unless perhaps using a mpsc channel? or what would you suggest for such an scenario?

kdy1:
When will be this feature merged?

Acanguven:
Any updates on this? It will probably make Neon lot more useful and amazing. Because the current state of this solution limits us we always have to use node as the data source.

crates/neon-sys/native/src/neon.cc
```diff
@@ -522,17 +522,23 @@ extern "C" void Neon_Task_Schedule(void *task, Neon_TaskPerformCallback perform,
   neon::queue_task(internal_task);
 }
 
-extern "C" void* Neon_ThreadSafeCallback_New(v8::Local<v8::Value> self, v8::Local<v8::Function> callback) {
+extern "C" void* Neon_EventHandler_New(v8::Local<v8::Function> callback) {
```
dherman: This was a nice idea, but unfortunately it's not future-proof for N-API, which doesn't have an equivalent of `isolate::GetCurrent()` and `isolate->GetCurrentContext()`. And in fact, this means the `bind` API is also not future-proof for N-API.

What this means is, we really need the `EventHandler` to take a Neon context as an explicit argument to its constructor(s).

So the best path forward is going to be: let's just drop this down to a single constructor, which takes three arguments: `cx`, `self`, and `callback`. It's a noticeably less convenient API, but I think that's OK to get started. Down the road, we can experiment with convenience methods of `cx` (just like the other convenience constructors such as `cx.string(...)` etc), like `cx.event_handler(...)`.

But let's wait on the convenience forms and for now just have one method in the API:

```rust
EventHandler::new(cx, self, callback)
```

dherman:
> Any updates on this? It will probably make Neon lot more useful and amazing. Because the current state of this solution limits us as we always have to use node as the data source.

@Acanguven Thanks for keeping us on task (no pun intended) :P I've just done a review on the latest revisions. @geovie Hopefully the next round should be the last needed revision!

crates/neon-sys/native/src/neon.cc
geovie: @dherman Should be good now, please take a final look! 😄

dherman:
Ah, I see we're failing some Windows builds. I reran and it's failing again. Investigating...

Others:
I was curious about this test failure. Here are the two interesting bits of the CI build that failed:
[compile fail stuff](https://gist.github.com/Others/c8b70b6a1db5ee7c73e9bd5a8f598600)
[electron/gyp fail stuff](https://gist.github.com/Others/5024b0cc51dd61b82259b0576663a1f0)

dherman:
So the compile-fail bug is probably just the usual flakiness of those tests (they're extremely sensitive to tiny changes in the Rust compiler's output formatting). I'm going to need to just disable those tests until we can come up with a better strategy for them.

From searching around, I think the electron failure is also likely unrelated to this PR: it seems like [this bug](https://github.com/nodejs/node-gyp/issues/1933). I'm looking into what we can do to fix that issue.

dherman:
OK I can confirm that the workaround from that issue (`npm i -g npm@6.11.3`) does indeed make the Electron tests pass.

This is a bit annoying! We'll need to track that node-gyp issue until they release a fix. For the time being, maybe we can get CI working by applying the workaround in the Appveyor config. I'll try a little PR to do that, which should rebase fine with this PR.

dherman:
Testing out a workaround: https://github.com/neon-bindings/neon/pull/483

dherman:
I got CI working again on master with #483 so I went ahead and rebased this PR -- let's see how rerunning CI goes! 🤞

dherman:
Doh! It looks like the changes in master did break this PR (my fault, since it took me so long to diagnose these CI issues).

@geovie Would you like me to try to fix this in your PR? (GitHub gives me permission to update your PR, but I don't want to push non-rebase changes without your permission.) I'm happy to do it unless you'd rather.

geovie:
@dherman Go for it, I won't manage to do it this week...

dherman:
Next step to get CI working should be to conditionally replace the implementation with `unimplemented!()` for the N-API backend for now. I'll do that next...

anymos:
Looking forward to see that part going live

Issue #384, opened 2019-01-11
neon build should rebuild if nodejs version is changed


kdy1:
See https://travis-ci.com/swc-project/node-swc/jobs/169705074

amilajack:
Good catch. I'll start working on a fix for this

dherman:
Great catch! I just talked with @amilajack about how to tackle this one. Here are the steps I think should work:

- [ ] BuildSettings.current() should detect the Node version (via `child_process.spawnSync("rustc", ["--version"])`
- [ ] add a `private node: string | null` property to BuildSettings class
- [ ] BuildSettings.construct() should take the Node version as an extra (`string | null`) parameter
- [ ] BuildSettings.fromJSON() and BuildSettings.toJSON() should also serialize/deserialize the Node version
- [ ] BuildSettings.match() should also compare the `.node` fields of the two instances

This could definitely use some tests, too. The caching logic has been brittle. If that turns out to be too hard, still best to implement this fix first and we can add tests separately, but see what you think you can do!

PR #388, opened 2019-01-18
cache bust neon build if node version changes

amilajack:
Closes https://github.com/neon-bindings/neon/issues/384

This PR introduces the following changes:

## Changes

- Busts cache if `nodeVersion` in `./native/artifacts.json` !== current node version
- Changes to `rust.ToolChain` type declarations
- Adds `@types/node` so TS can have type declarations for node
- Changes `BuildSettings` to take a `nodeVersion` argument

## How to test this PR

```bash
git clone https://github.com/amilajack/neon
cd neon
git checkout node-version-cli-cache-bust-fix
cd cli && npm install && npm run transpile && npm link
neon new foobar
cd foobar
npm link neon-cli

# add the following line to the package.json:
# "build": "neon build"

# trigger a build, generates artifacts.json
npm run build

# Change nodeVersion in artifacts.json
# Then build again, should build without cache
npm run build
```

You can also test this by changing your node version (but this is a bit more work). This can be done by linking too a different node version with `brew link node@8` (for example).


cli/src/rust.ts
```diff
@@ -1,23 +1,23 @@
 import * as async from './async/child_process';
 import * as child_process from 'child_process';
 
-export type Toolchain = string;
+export type Toolchain = string | null;
```
dherman: You can remove the `| null` from this definition once you make the change I suggested above.

cli/src/build-settings.ts
```diff
@@ -54,25 +65,32 @@ export default class BuildSettings {
...
+      throw new TypeError("value must be an object");
     }
-    let rustc = value.rustc;
-    let env = value.env;
+    let { rustc, env, nodeVersion } = value;
     if (typeof rustc !== 'string') {
-      throw new TypeError("value.rustc is not a string");
+      throw new TypeError("value.rustc must be a string");
+    }
+    if (!('nodeVersion' in value)) {
```
dherman: I think this is not the safest logic: the case where there's no `'nodeVersion'` key is when they used a previous version of Neon, in which case we don't know what Node version they were using, so we should leave it `null` to allow ourselves to cache-bust.

cli/src/build-settings.ts
```diff
@@ -13,35 +13,46 @@ function isStringDict(x: JSON.Object): x is Dict<string | null> {
 
 export default class BuildSettings {
   private rustc: string;
+  private nodeVersion: string;
```
dherman: Based on my comment about defaulting below, this field will need to be `string | null`.

cli/src/build-settings.ts
```diff
@@ -13,35 +13,46 @@ function isStringDict(x: JSON.Object): x is Dict<string | null> {
...
   private env: Dict<string | null>;
 
-  constructor(rustc: string, env: Dict<string | null>) {
+  constructor(rustc: string, nodeVersion: string, env: Dict<string | null>) {
```
dherman: Forgot to mention in the review: the type of this parameter should also by `string | null`.

cli/src/build-settings.ts
dherman: It's worth adding a comment in the code here, explaining that it can be `null` because older versions of Neon didn't have this key. That will help the null case be understandable/maintainable.

amilajack:
Hey @dherman @kjvalencik! Would love to know what I can do to push this across the finish line

amilajack:
@dherman I realized that the PR actually does work. The issue was actually my steps to reproduce. When running the install command, the previously linked dependency is override and therefore the test repo will end up using the version in the registry.

```bash
git clone https://github.com/amilajack/neon
cd neon
git checkout node-version-cli-cache-bust-fix
cd cli && npm install && npm run transpile && npm link
neon new foobar
cd foobar
npm link neon-cli

# add the following line to the package.json:
# "build": "neon build"

# trigger a build, generates artifacts.json
npm run build
# Change nodeVersion in artifacts.json
# Then build again, should build without cache
npm run build
```

cli/src/build-settings.ts
```diff
@@ -13,35 +14,46 @@ function isStringDict(x: JSON.Object): x is Dict<string | null> {
...
     return Object.keys(this.env).every(key => {
       return (!this.env[key] && !other.env[key]) ||
              (this.env[key] === other.env[key]);
     });
   }
 
-  static current(toolchain: rust.Toolchain) {
-    const spawnResult = rust.spawnSync("rustc", ["--version"], toolchain);
+  static getNodeVersion(): string {
+    const nodeVersionResult = child_process.spawnSync("node", ["--version"]);
```
kjvalencik: Can you check `process.version` or `process.versions`?

amilajack: Are you talking about like the following:

```js
const nodeVersionsResult =
  child_process.spawnSync("node", ["-e 'console.log(process.versions)'"]);
```

cli/src/build-settings.ts
```diff
@@ -13,35 +14,46 @@ function isStringDict(x: JSON.Object): x is Dict<string | null> {
 
 export default class BuildSettings {
   private rustc: string;
+  private nodeVersion: string | null;
   private env: Dict<string | null>;
 
-  constructor(rustc: string, env: Dict<string | null>) {
+  constructor(rustc: string, nodeVersion: string | null, env: Dict<string | null>) {
     this.rustc = rustc;
+    this.nodeVersion = nodeVersion
```
dherman: Missing semicolon. (We should adopt a linter and autoformatter! Not in this PR though.)

cli/src/build-settings.ts
```diff
@@ -13,35 +14,46 @@ function isStringDict(x: JSON.Object): x is Dict<string | null> {
...
     this.env = env;
   }
 
   match(other: BuildSettings) {
+    if (other.nodeVersion !== this.nodeVersion) return false;
```
dherman: Make this multiline with braces:

```js
if (other.nodeVersion !== this.nodeVersion) {
  return false;
}
```

cli/src/build-settings.ts
dherman: I'm personally not in favor of the `const`-everywhere style, but mostly I think we should be consistent. Right now, we've only use `const` for constants -- i.e., global variables bound to immutable values. Would you be OK with just using `let` to be consistent with the house style?

test/cli/package.json
```diff
@@ -49,7 +49,7 @@
     "typescript": "^2.2.2"
   },
   "scripts": {
-    "test": "mocha lib/acceptance",
+    "test": "npm run transpile && mocha lib/**/*.js",
```
dherman: Nice catch! I'm constantly forgetting to run this manual step.

cli/src/build-settings.ts
amilajack: I'd love to! Whatever it takes to keep the style consistent. My bad for not catching this in advance.

amilajack:
@kjvalencik Updated the code to reflect your comment

dherman:
Merged! 🎉

Issue #362, opened 2018-09-29
Problem with loading native modules which built with --release


katyo:
The `neon build --release` produces broken NodeJS native modules:

```
Error: Module did not self-register.
    at Object.Module._extensions..node (module.js:681:18)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
...
```

I understand, this is possibly a Cargo/Rust problem related to inapproppriate config.
Experimentally I found that it actual for all opt-level's > 0
But I don't know a proper solution.

katyo:
Link to rust issue https://github.com/rust-lang/rust/issues/54691

katyo:
I found temporary workaround.

1. Add init function which calls `__LOAD_NEON_MODULE` itself:

```rust
#[no_mangle]
pub extern "C" fn node_addon_init() {
    __LOAD_NEON_MODULE();
}

register_module!(mut cx, {
  // your module code
}
```

The `no_mangle` directive is important.

2. Fix "install" script in _package.json_:

```json
"scripts": {
    "install": "RUSTFLAGS='-C link-arg=-Wl,-init=node_addon_init' neon build --release",
  },
```

This solution uses ".init" section instead of ".ctors"/".init_array".
Of course, I tested this on linux only.

kjvalencik:
This was fixed in #392.

PR #392, opened 2019-01-28
Fix for release builds (issue #362)

katyo:
During link-time optimization (particularly when you build with --release)
linker removes unused static symbol `__LOAD_NEON_MODULE`.

I forced it to be used to avoid such incorrect behavior.


katyo:
See also https://github.com/rust-lang/rust/issues/54691

dherman:
Thanks @katyo !

Issue #357, opened 2018-09-15
error: Neon only builds with --release. For tests, try `cargo test --release`.


mlengse:
## environment
- windows 10 pro 1083 OS build 17134.191
- system type 64-bit OS, x64-based processor
- rustc 1.29.0 (aa3ca1994 2018-09-11)
- cargo 1.29.0 (524a578d7 2018-08-05)
- node v8.11.3
- npm 6.4.1
- yarn v1.9.4
- msvs_version 2015
- python 2.7.14
- neon 0.2.0

## step
- npm install --global --production windows-build-tools
- npm install --global neon-cli
- neon new threading-hint
- cd threading-hint
- yarn

## error message

```
yarn install v1.9.4
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
$ neon build
[##############################################################------------------------------------------------] 96/171neon info forcing rebuild for new build settings
neon info running cargo
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling version_check v0.1.4
   Compiling libc v0.2.43
   Compiling regex v0.2.11
   Compiling ucd-util v0.1.1
   Compiling utf8-ranges v1.0.1
   Compiling gcc v0.3.54
   Compiling neon-build v0.2.0
   Compiling semver-parser v0.7.0
   Compiling cslice v0.2.0
   Compiling memchr v2.0.2
   Compiling regex-syntax v0.5.6
   Compiling lazy_static v1.1.0
   Compiling neon v0.2.0
   Compiling threading-hint v0.1.0 (file:///E:/proj/threading-hint/native)
   Compiling aho-corasick v0.6.8
   Compiling semver v0.9.0
   Compiling thread_local v0.3.6
   Compiling neon-runtime v0.2.0
error: Neon only builds with --release. For tests, try `cargo test --release`.
   --> C:\Users\USER\.cargo\registry\src\github.com-1ecc6299db9ec823\neon-0.2.0\src\lib.rs:288:1
    |
288 | compile_error!("Neon only builds with --release. For tests, try `cargo test --release`.");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `neon`.

[...]

PR #400, opened 2019-04-03
On Windows, build the C code in release mode always.

retep998:
Rust always uses the release CRT, even in debug mode, so the C code should do the same.

Fixes https://github.com/neon-bindings/neon/issues/357

### DIFF
--- a/crates/neon-sys/native/binding.gyp
+++ b/crates/neon-sys/native/binding.gyp
@@ -13,6 +13,15 @@
                         'LinkTimeCodeGeneration': 0
                     }
                 }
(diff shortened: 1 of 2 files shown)


retep998:
I updated this PR to now still build the C code in debug mode, but I've tried my best to convince gyp to tell the C compiler to use the release CRT. I still need to test whether this new change does the right thing.

dherman:
Thanks as always for helping us with the intricacies of building on Windows, @retep998, and again, my bad for overlooking this PR before. 🙏

Issue #389, opened 2019-01-23
"Error: Module did not self-register." in Electron 4.0.1 on Windows(but works fine in 3.1.1)


Syndim:
OS: Windows 10 1809
Rust version: 1.32.0 x86_64 MSVC
Electron version: 4.0.1
Neon binding version: 0.2.0
Node.js version: 10.15.0

How to reproduce:
1. Clone the neon-binding example repo: `https://github.com/neon-bindings/examples`
2. Change the electron version from `3.0.10` to `4.0.1` in packages.json under `electron-app`
3. `yarn` and then `yarn start`

Got the following log in the electron window:

```
Error: Module did not self-register.
    at process.module.(anonymous function) [as dlopen] (ELECTRON_ASAR.js:160:31)
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:722:18)
    at Object.module.(anonymous function) [as .node] (ELECTRON_ASAR.js:160:31)
    at Module.load (internal/modules/cjs/loader.js:602:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:541:12)
    at Function.Module._load (internal/modules/cjs/loader.js:533:3)
    at Module.require (internal/modules/cjs/loader.js:640:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object. (C:\examples\electron-app\node_modules\neon-hello\lib\index.js:1:168)
    at Object. (C:\examples\electron-app\node_modules\neon-hello\lib\index.js:5:3)
```

It doesn't reproduce under electron v3, is there any way to debug what's happening?

amilajack:
Considering that `4.0.1` is a pretty recent major server of electron, it is likely that this might be an electron issue.

itserg2008:
I have the same error even in 3.1.1. Tried to downgrade to this version but got the same error.

itserg2008:
As I made a build using --release flag, I found the solution: here it is https://github.com/katyo/neon/commit/3319c6ce54ba67451b236901ca0669397d68c659

itserg2008:
I see exactly the same behaviour - works fine for me on Windows 10 with Electron 3.0.10, but couldn't get it work on 4.0.X.

Any ideas?

jrd-rocks:
In my own project i have the same problem. Electron 4.1.4 gives the error above. Downgrading to electron 3.1.8 and it works. Definitely something weird going on.

jrd-rocks:
[...]

PR #404, opened 2019-04-06
Use delayed loading for windows

jrd-rocks:
Electron on windows needs to use delayed loading. For node.js delayed loading works as well (it's the default when building with node.gyp)

For full details see https://github.com/jrd-rocks/neon/blob/electron_delay_hook/README%20ELECTRON%204.md #389 

~~The neon command that creates a default project still needs fixing to add a .cargo/config file with the linker command.~~ Never mind that, i updated the cli script as well (a breeze after figuring out the linker delay load black magic)

### DIFF
--- a/crates/neon-runtime/src/win_delay_load_hook.cc
+++ b/crates/neon-runtime/src/win_delay_load_hook.cc
@@ -12,7 +12,6 @@
 
 #include <delayimp.h>
 #include <string.h>
-#include <stdexcept>
 
 extern "C" FARPROC WINAPI load_exe_hook(unsigned int, DelayLoadInfo*);
 
@@ -21,7 +20,6 @@ FARPROC WINAPI load_exe_hook(unsigned int event, DelayLoadInfo* info) {
   if (event != dliNotePreLoadLibrary)
     return NULL;
 
- printf ( "loading hook\n" );
   if (_stricmp(info->szDll, "iojs.exe") != 0 &&
       _stricmp(info->szDll, "node.exe") != 0)
     return NULL;
--- a/crates/neon-runtime/src/win_delay_load_hook.cc
+++ b/crates/neon-runtime/src/win_delay_load_hook.cc
@@ -13,9 +13,7 @@
 #include <delayimp.h>
 #include <string.h>
 
-extern "C" FARPROC WINAPI __load_exe_hook(unsigned int, DelayLoadInfo*);
-
-FARPROC WINAPI __load_exe_hook(unsigned int event, DelayLoadInfo* info) {
+FARPROC WINAPI load_exe_hook(unsigned int event, DelayLoadInfo* info) {
   HMODULE m;
   if (event != dliNotePreLoadLibrary)
     return NULL;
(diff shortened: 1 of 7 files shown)


amilajack:
I think it might be better to add `README ELECTRON 4.md` to a wiki

jrd-rocks:
> I think it might be better to add `README ELECTRON 4.md` to a wiki

I think it should be part of the documentation on the website, but couldn't find where to change that.

amilajack:
Oh yea that's even better. I'm the maintainer of the docs so I'll PR there.

crates/neon-runtime/src/win_delay_load_hook.cc
```diff
@@ -0,0 +1,31 @@
...
+ * This allows compiled addons to work when node.exe or iojs.exe is renamed.
+ */
+
+ 
+#include <windows.h>
+
+#include <delayimp.h>
+#include <string.h>
+
+extern "C" FARPROC WINAPI __load_exe_hook(unsigned int, DelayLoadInfo*);
```
bnoordhuis: You don't need to export the symbol, only `__pfnDliNotifyHook2` needs to be exported. Just this should suffice:
```c
static FARPROC WINAPI __load_exe_hook(unsigned int event, DelayLoadInfo* info) {
  // ...
}

decltype(__pfnDliNotifyHook2) __pfnDliNotifyHook2 = __load_exe_hook;
```
By the way, I think you could turn this into a .rs file. You'd probably have to provide definitions of `dliNotePreLoadLibrary` and `DelayLoadInfo` though, so it might be not be shorter.

jrd-rocks: Much better, thanks!

I have no idea how to write those definitions they're pretty complicated to me. And as neon-runtime already includes c code anyways, i figured a little more shouldn't be a problem. Though of course pure rust is nicer.

Optionally removing it is as simple as removing `.cargo/config` (though the hook then still gets needlessly compiled in neon-runtime) as long as it's documented it's there i think it shouldn't be an issue.

itserg2008:
Hello @dherman , do you have any plans to review and merge it to the master branch in the nearest time?

kjvalencik:
@jrd-rocks Thanks for this contribution! Are you still interested in working on this? If so, I have some notes on simplifications. Thanks, again!

joshuef:
I was trying to get this going, just to be clear: I'll need to use the `windows-electron` fork as a dep in our neon app, but also build the app on windows using the forked CLI too, right? Should anything else be needed to test this?

I'll be happy to try and dive into neon to try and get the needed changes in assuming i can get them going (so far still seeing `module did not self load` :| )

kjvalencik:
@joshuef Yes, you would also need to use the `electron-build-env` wrapper. I would not expect this on the fork since it is working locally and in CI.

kjvalencik:
@joshuef Are you on Slack? I'm not a windows developer and it would be helpful if you are able to test my work. Thanks!

PR #411, opened 2019-05-15
Support CARGO_BUILD_TARGET triple env

stoically:
Thanks for your work on neon!

I'm currently trying to get [nodejs-mobile](https://code.janeasystems.com/nodejs-mobile) (specifically `nodejs-mobile-react-native` for Android) to work with neon native modules. In order to have a succeeding neon build, the cli needs to respect the [`CARGO_BUILD_TARGET` env](https://doc.rust-lang.org/nightly/cargo/reference/config.html#environment-variables) which could be set by the `nodejs-mobile-react-native` gradle script (issue with some more details is over here https://github.com/janeasystems/nodejs-mobile/issues/193).

### DIFF
--- a/cli/src/target.ts
+++ b/cli/src/target.ts
@@ -48,6 +48,10 @@ export default class Target {
       this.triple = '';
     }
 
+    if (process.env.CARGO_BUILD_TARGET) {
+      this.triple = process.env.CARGO_BUILD_TARGET;
+    }
+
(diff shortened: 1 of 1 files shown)


spruce:
I support that change. I just figured out a very similar solution. In my opinion this should allow neon to do cross compilation.

spruce:
@dherman what is needed to get this merged?

PR #416, opened 2019-07-01
rust 2018

uvd:
rust 2018 and run Cargo fmt


kjvalencik:
@uvd I appreciate the changes! Unfortunately, the code base has diverged pretty far since this PR. I would still like to make these changes, but I would like to defer until #440 lands. Thanks!

PR #410, opened 2019-05-09
Fixes for Node 12.

kjvalencik:
Resolves #409. There are still warnings that should be resolved. Especially, `MakeCallback`, which will break domain support.

PR #420, opened 2019-08-19
Run appveyor against node@12

amilajack:
Complements https://github.com/neon-bindings/neon/pull/410

- Drop node@6 on CI
- Restrict node engines to node@8

Also node@6 builds seem to be failing. Should we drop support for it or fix it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,17 +12,12 @@ matrix:
       rust: beta
     - env: TRAVIS_NODE_VERSION="8"
       rust: nightly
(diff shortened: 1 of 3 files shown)


kjvalencik:
Node 6 is no longer maintained. I think we should drop support for it.

amilajack:
Makes sense. I'll remove it from travis and appveyor and also update the engines in the `package.json`

amilajack:
@kjvalencik Looks good to me!

kjvalencik:
Thanks!

PR #427, opened 2019-08-24
Satisfy cargo validation for neon-runtime

dherman:
This PR fixes the bug where neon-runtime fails cargo's validation for `cargo publish` (or `cargo package`). The `build.rs` was previously building intermediate results in the `neon-runtime` crate root, instead of inside `OUT_DIR`. This PR instead copies the native sources into `$OUT_DIR/native` and builds there.


dherman:
Added a test to ensure package validation doesn't regress.

PR #431, opened 2019-09-02
Disable module registration when building tests. Fixes #425.

### LINKED ISSUE #425: Fail when running tests: undefined reference to `v8::<>
Neon version: 0.3.1
Rust version: 1.37.0
Platform: Linux

Running `cargo build` and `cargo build --release` works fine. However, when I try to run tests, I encounter a lot of errors like:

```
neon.cc:(.text+0x98): undefined reference to `v8::Isolate::Enter()'                                                                                                                                                          
          neon.cc:(.text+0xa7): undefined reference to `v8::HandleScope::HandleScope(v8::Isolate*)'                                                                                                                                    
          neon.cc:(.text+0xc3): undefined reference to `v8::HandleScope::CreateHandle(v8::internal::Isolate*, v8::internal::Object*)'
```

This continues for many more lines. Any idea how to mitigate this?

kjvalencik:


src/lib.rs
```diff
@@ -94,6 +96,8 @@ macro_rules! register_module {
                 // Suppress the default Rust panic hook, which prints diagnostics to stderr.
                 ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
 
+                // During tests, node is not available. Skip module registration.
+                #[cfg(not(test))]
```
kjvalencik: Is it worth adding a crate feature to allow forcing this even when building tests? I can't think of a use case.

dherman: I'd say wait until we know it's an issue.

PR #433, opened 2019-09-05
Squash mem::uninitialized() warnings. Fixes #424.

### LINKED ISSUE #424: Update usage of deprecated `mem::uninitialized()`

kjvalencik:
* Moved to safer `mem::MaybeUninit` for buffers
* Switched to more appropriate `ptr::null_mut()` for `class::get_name()`
* Limited size of unsafe blocks
* Return sizes directly instead of output parameters

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ neon-build = { version = "=0.3.1", path = "crates/neon-build" }
 
 [dev-dependencies]
 rustc_version = "0.2"
-lazy_static = "0.2.8"
+lazy_static = "1.4.0"
 
 [dependencies]
 cslice = "0.2"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ neon-build = { version = "=0.3.1", path = "crates/neon-build" }
 
 [dev-dependencies]
 rustc_version = "0.2"
-lazy_static = "0.2.8"
+lazy_static = "1.4.0"
```
kjvalencik: Snuck this unrelated change in to squash the only remaining warning.

src/lib.rs
```diff
@@ -384,6 +384,11 @@ mod tests {
         log("package_test");
 
         let test_package = project_root().join("crates").join("neon-runtime");
-        run("cargo package", &test_package);
+
+        if std::env::var("CI") == Ok("true".to_string()) {
```
kjvalencik: It was a little annoying having an unrelated test fail during development. Disabling the uncommitted changes check outside of CI seemed like a reasonable compromise.

PR #439, opened 2019-09-12
CLI dev flags

dherman:
Enables two new flags for `neon new`, both of which should become useful for development and experimental work:

- `neon new --neon=<version-or-path>`: specify a non-default version of Neon or path to a local Neon project directory on the filesystem
- `neon new --features=<neon-experimental-features-list>`: specify a space-separated list of experimental Neon features to enable

These should be useful for contributing or testing when we start developing N-API support, so that we can put the experimental support behind a feature flag until it stabilizes.

### DIFF
--- a/cli/src/ops/neon_new.ts
+++ b/cli/src/ops/neon_new.ts
@@ -1,4 +1,4 @@
-import { readFile, writeFile, mkdirs } from '../async/fs';
+import { readFile, writeFile, mkdirs, stat } from '../async/fs';
 import { prompt } from 'inquirer';
 import gitconfig from '../async/git-config';
 import * as path from 'path';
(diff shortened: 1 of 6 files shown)


cli/src/ops/neon_new.ts
```diff
@@ -117,13 +141,39 @@ export default async function wizard(pwd: string, name: string) {
...
+    simple: boolean,
+    paths?: { neon: string, "neon-build": string },
+    version?: string,
+    features?: Array<string>
+  } = { simple: neonVersion.type !== 'path' && !features };
+
+  if (neonVersion.type === 'path') {
+    libs.paths = {
+      neon: path.resolve(neonVersion.value),
+      "neon-build": path.resolve(neonVersion.value, "crates", "neon-build")
```
kjvalencik: What do you think about using [`path.relative`](https://nodejs.org/api/path.html#path_path_relative_from_to) here? `path.resolve` will produce an absolute path which isn't very friendly to switching across environments (e.g., docker).

cli/templates/package.json.hbs
```diff
@@ -16,7 +16,7 @@
   "license": "{{project.license}}",
 {{/if}}
   "dependencies": {
```
kjvalencik: Not part of this change, but shouldn't this be in `devDependencies`? I think most neon projects would be top level applications and not libraries.

cli/src/ops/neon_new.ts
```diff
@@ -117,13 +141,39 @@ export default async function wizard(pwd: string, name: string) {
...
+  }
+
+  if (features) {
+    libs.features = features.split(/\s+/);
+  }
+
+  let cli = neonVersion.type === 'version'
+    ? "^" + neonVersion.value
+    : neonVersion.type === 'path'
+    ? path.resolve(neonVersion.value, "cli")
```
kjvalencik: Should this be prefixed with `file:`?

cli/src/ops/neon_new.ts
dherman: How about we go with: it'll generate a relative path if you pass a relative path and an absolute path if you pass an absolute path.

cli/templates/package.json.hbs
dherman: As long as `neon build` is needed for running the build, it has to be in `"dependencies"`. If we can get rid of `neon build` (which would be awesome!) then we can move it to `"devDependencies"`.

PR #440, opened 2019-09-12
Initial skeleton for N-API support.

dherman:
First step towards N-API support. This PR is exploring an implementation strategy with scaffolding to build N-API support behind a feature flag (`"napi-runtime"`), allowing development to happen on master as we continue maintaining the existing codebase.

What I've implemented so far is almost nothing: it's barely enough to support creating an empty module that basically does nothing when it's loaded. But from here, we can split up the work into incremental, separate chunks.


crates/neon-runtime/native-napi/src/neon.cc
```diff
@@ -0,0 +1,2 @@
+#include <node_api.h>
```
kjvalencik: I don't think I understand why C/C++ code is still required with N-API. Since it's ABI stable, can `node-gyp` be completely removed and it be linked exclusively by cargo?

Qard: If you want to duplicate the entirety of N-API in Rust, I guess. Seems like a bunch of wasted effort when you can just link to the already existing C headers. 🤷‍♂

dherman: @kjvalencik I think we can come close, and even if we can't drop node-gyp, it's a good goal to have as little C glue code as possible and write as much of the runtime in Rust as possible.

The reason I'm not sure we can eliminate it is the way to initialize a module requires a C macro ([`NAPI_MODULE_INIT`](https://nodejs.org/api/n-api.html#n_api_module_registration)) that seems to expand into non-ABI-stable details. It's worth reaching out on the N-API repo to see if they might consider creating a stable way to do this from non-C languages, but so far I can't figure out a way around it otherwise.

@Qard KJ doesn't mean mean replacing the implementation, just porting the headers to Rust as `extern` declarations. The [bindgen](https://rust-lang.github.io/rust-bindgen/) tool can facilitate the porting of the headers to Rust so it doesn't have to be entirely manual.

NickNaso: Hi everyone, 
I'm not an expert about module registration and context awareness, I reported this discussion so we can discuss in the next N-API meeting that will happen on Monday 16th September. I post the link zoom https://zoom.us/j/363665824 and because I don't know your time I post the link to our calendar https://calendar.google.com/calendar/embed?src=nodejs.org_nr77ama8p7d7f9ajrpnu506c98@group.calendar.google.com&pli=1
Hope to see you soon.

dherman: Thanks @NickNaso!

I may not be able to join that call this week but I'll see. If someone happen to have the opportunity to join, I'd love if we could ask two things:

1. Are the `napi_module` type and the `neon_register_module` function ABI stable? They aren't documented (the only documented way to register a module is with C macros) but if they're stable, @kjv has already done a proof of concept that we can use them from Rust without C glue.
2. If they are guaranteed to be stable, could they be added to the docs?

NickNaso: Hi everyone,
today in the N-API meeting we discussed about your questions. You can see the video registration here: https://www.youtube.com/watch?v=rEHyMIwP020  - From minute 15:29 to minute 34:05. 
See: https://github.com/nodejs/abi-stable-node/issues/346#issuecomment-531880219
I hope that this is helpful to you and let us know if you need some other things.

kjvalencik: @geovie That one appears to be unmaintained. `nodejs-sys` appears more active--I used it in my PoC.

dherman:
@kjvalencik I think this is ready for review now! If it would help I can leave in-place GitHub comments explaining what's going on in the PR.

crates/neon-runtime/native-napi/src/neon.cc
dherman: @NickNaso I finally had a chance to watch that video. Thank you again, so much, for bringing this up in the meeting! The N-API team were so helpful. It's really nice that there's such a clean solution now: we can just define the symbol `napi_register_module_v1` and this will automatically get registered (even in a context-aware way for every worker!). No more constructors needed, no more explicit registration needed. So nice!

dherman:
BTW my next step after we finish and merge this PR is to write a [quest issue](https://medium.com/@trek/source-quest-ff7d227d8fed) providing background, a task list, and documentation on how to contribute to the N-API port. My goal is to make the work as parallelizeable and as friendly to potential new contributors as possible.

src/lib.rs
```diff
@@ -24,89 +22,138 @@ pub mod prelude;
...
+        ///     Ok(())
+        /// });
+        /// ```
+        #[macro_export]
+        macro_rules! register_module {
+            ($module:pat, $init:block) => {
+                register_module!(env!("CARGO_PKG_NAME"), $module, $init);
+            };
+
+            ($name:expr, $module:pat, $init:block) => {
```
kjvalencik: I don't expect `$name` to be used. Is it being kept to minimize breaking changes?

dherman: We don't want to expose the N-API data structures to the programmer. The lack of hygiene is unfortunate but I wouldn't want to pollute the UI with implementation details.

Re `$name`, I think I copied that from your PoC 😛 We could separate that out into a doc-hidden helper macro.

dherman: Oh ha, `$name` is unused now! Cleanup coming…

Issue #432, opened 2019-09-04
Update macros so that they can easily be used with `use`


kjvalencik:
Some of the macros call other internal only macros. This can make importing them somewhat unintuitive and verbose.

```rust
use neon::{class_definition, declare_types, impl_managed, register_module};
```

This likely needs two changes:
* Add `declare_types` and `register_module` to `neon::prelude`
* Make `class_definition` and `impl_managed` local inner macros

https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros
https://github.com/rust-lang/rust/pull/51496

PR #441, opened 2019-09-13
Add macros to neon::prelude

theJian:
#432 

I'm not sure where to add a test for this change so I update test/dynamic. Not sure if I test it in the right place.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,7 +111,7 @@ macro_rules! register_module {
 }
 
 #[doc(hidden)]
-#[macro_export]
+#[macro_export(local_inner_macros)]
 macro_rules! class_definition {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -292,6 +292,14 @@ macro_rules! declare_types {
     { } => { };
 }
 
+#[doc(hidden)]
+#[macro_export]
+macro_rules! neon_stringify {
```
kjvalencik: Was this necessary? It doesn't seem like it should use `stringify!` since `use std` is implied.

theJian: class_definition use `stringify!`. According to https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros we need add a local macro that wraps `stringify!` since `stringify!` isn't in the neon crate. Without this code that uses class_definition won't compile.

theJian: `test/dynamic/native/src/js/classes.rs` uses declare_types! to declare classes. It has the error message "could not find 'stringify' in '$crate' when I compile it without wrapping `stringify!` in a local macro.

Issue #297, opened 2018-02-07
support inside a workspace


jgrund:
It would be great if neon bindings would work inside a workspace.

Currently it appears you can build, but since the target is at the top level of the workspace, neon can't locate the generated `.so` file.

dherman:
@jgrund Curious, do you mean yarn workspaces or cargo workspaces? I'm guessing yarn?

jgrund:
Cargo workspaces. Here is an example repo where I'm using neon in a workspace: https://github.com/intel-hpdd/rust-libzfs/

dherman:
This is very much top of mind! There's the beginnings of a discussion of using workspaces in the [cargo wrapper RFC](https://github.com/neon-bindings/rfcs/pull/4). I don't think that RFC is 100% nailed yet, but would love your thoughts about it if you have a chance to read it. I will also run through it again and make sure examples like yours are doable.

dherman:
Labeling as `needs RFC` because it should be addressed by the cargo wrapper RFC once it's in a shape we all like (i.e., it doesn't need a separate RFC from that).

retep998:
Just chiming in to say that I would really absolutely love being able to use use neon inside cargo workspaces.

kjvalencik:
@retep998 My use case for workspaces is pretty simple. I wanted to be able to build from the root as well as use RLS in VS Code (which requires the root of the project be a Rust project). If your use case is similar, there is a workaround. Create a `.cargo/config` file that sets the target to `native/target`.

```sh
mkdir -p .cargo
cat >> .cargo/config << EOF
[build]

target-dir = "native/target"
EOF
```

dbousamra:
@kjvalencik This sort of works. It can now find the .so file.

But I receive: `Error: libstd-e39317eb74365d3c.so: cannot open shared object file: No such file or directory`

When i try to require the .node file

amilajack:
Should this be closed now that https://github.com/neon-bindings/neon/pull/442 is resolved?

kjvalencik:
@amilajack I think it should remain open because it still does not support multiple neon modules in a workspace.

jhg:
@kjvalencik limits in support is only about multiple neon modules in a workspace? if the workspace has a neon module, a lib and a cli is it supported currently?

kjvalencik:
@jhg. Yes, that should work perfectly.

kjvalencik:
[...]

PR #442, opened 2019-09-13
Support workspaces.

kjvalencik:
Implements https://github.com/neon-bindings/neon/issues/297

### DIFF
--- a/cli/src/project.ts
+++ b/cli/src/project.ts
@@ -3,22 +3,39 @@ import Crate from './crate';
 import Target from './target';
 import BuildSettings from './build-settings';
 import log from './log';
-import { spawn } from './async/child_process';
+import { execFile, spawn } from './async/child_process';
 import * as rust from './rust';
 
 export type ProjectOptions = {
-  crate?: string
+  crate: string
+  targetDirectory: string
 };
 
 /** A Neon project and its directory tree. */
 export default class Project {
   readonly root: string;
+  readonly targetDirectory: string;
   readonly crate: Crate;
 
-  constructor(root: string, options: ProjectOptions = {}) {
-    let { crate = 'native' } = options;
+  private constructor(root: string, options: ProjectOptions) {
+    let { crate, targetDirectory } = options;
     this.root = root;
+    this.targetDirectory = targetDirectory;
     this.crate = new Crate(this, { subdirectory: crate });
+  }
+
+  static async create(root: string, options: Partial<ProjectOptions> = {}): Promise<Project> {
+    let { crate = 'native' } = options;
+    const { stdout } = await execFile("cargo", ["metadata", "--format-version=1"], {
+      cwd: path.join(root, crate)
+    });
+    const targetDirectory: string = JSON.parse(stdout).target_directory;
+
+    return new Project(root, {
+      targetDirectory,
+      crate,
+      ...options
+    });
   }
 
   async build(toolchain: rust.Toolchain | null,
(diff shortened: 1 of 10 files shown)


cli/src/project.ts
```diff
@@ -3,24 +3,41 @@ import Crate from './crate';
...
+  private constructor(root: string, options: ProjectOptions) {
+    let { crate, targetDirectory } = options;
     this.root = root;
+    this.targetDirectory = targetDirectory!;
     this.crate = new Crate(this, { subdirectory: crate });
   }
 
+  static async create(root: string, options: Partial<ProjectOptions> = {}): Promise<Project> {
+    let { crate = 'native' } = options;
+    const { stdout } = await execFile("cargo", ["metadata", "--format-version=1"], {
```
kjvalencik: This is only necessary in `build` and would be fewer changes if it was moved there, but it felt wrong not to have `targetDirectory` on the `Project`.

dherman: Seems 👍  to me.

crates/neon-runtime/build.rs
```diff
@@ -26,6 +26,11 @@ fn main() {
 
     // 4. Copy native build artifacts
     copy_build_artifacts(&native_to, &out_dir);
+
+    // Rebuild if any of these environment variables change
```
dherman: Can you help me understand why this was needed for this PR?

dherman:
Actually one more big picture question: what exactly do we mean by a workspace here? Like, `<root>/native` can contain a workspace with multiple crates? Or do we mean somehow nesting a whole Neon project (including the JS) inside an outer cargo workspace? Or does this support both?

Also, it would be good to think about whether we want `neon new` to support any other project structure than the current one. Resolving that question doesn't block this PR, but worth discussing.

kjvalencik:
@dherman Technically, this means supporting projects where the compiled binary isn't in `native/target`. Practically, that means when it's part of a workspace. That could mean any of the following:

* Part of a workspace with multiple neon modules
* Part of a workspace with multiple crates, one being a neon module
* A top level `Cargo.toml` that makes `native/` the only member of the workspace (fixes VS Code and allows easier `cargo test`)

I think it would be nice for the default `neon new` to drop a `Cargo.toml` with `native` in the workspace, but I think that can be a different PR.

cli/src/project.ts
```diff
@@ -3,24 +3,41 @@ import Crate from './crate';
...
   readonly root: string;
+  readonly targetDirectory: string;
   readonly crate: Crate;
 
-  constructor(root: string, options: ProjectOptions = {}) {
-    let { crate = 'native' } = options;
+  private constructor(root: string, options: ProjectOptions) {
+    let { crate, targetDirectory } = options;
     this.root = root;
+    this.targetDirectory = targetDirectory!;
```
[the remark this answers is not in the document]
kjvalencik: It's not; it's from an older version of the code. I need to remove it.

crates/neon-runtime/build.rs
kjvalencik: It's an existing bug that this PR made apparent. These environment variables impact the compatibility of `libneon`. For example, with these changes `test/dynamic` compiles `neon-runtime` and then `test/electron` tries to use it and fails.

dherman:
I think changing the default layout should have an RFC to explain and motivate the design change. I definitely _love_ the idea of having `cargo` work in the project root. I drafted [one RFC that proposed changing the directory layout](https://github.com/neon-bindings/rfcs/pull/4) although I think some of the particular ideas there missed the mark. I wonder how far cargo and/or Neon would need to evolve before we could let people use vanilla cargo without needing a wrapper CLI tool.

PR #443, opened 2019-09-18
Rework N-API runtime to not need node-gyp

kjvalencik:


src/lib.rs
```diff
@@ -31,15 +31,44 @@ use cfg_if::cfg_if;
 
 cfg_if! {
     if #[cfg(feature = "n-api")] {
+        #[doc(hidden)]
+        pub use neon_runtime::sys as sys;
 
         #[macro_export]
         macro_rules! register_module {
```
kjvalencik: Usage:
```rust
register_module!(|_env, exports| {
    println!("Hello, World!");

    exports
});
```

src/lib.rs
```diff
@@ -31,15 +31,44 @@ use cfg_if::cfg_if;
...
+            ($name:expr, |$env:ident, $exports:ident| $body:block) => {
+                extern "C" fn __neon_init_module(
+                    $env: neon::sys::napi_env,
+                    $exports: neon::sys::napi_value,
+                ) -> neon::sys::napi_value $body
+
+                #[cfg_attr(target_os = "linux", link_section = ".ctors")]
+                #[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
+                #[cfg_attr(target_os = "windows", link_section = ".CRT$XCU")]
+                pub static __register_neon_module: extern "C" fn() = {
```
kjvalencik: This is required to be unique, but I _think_ because of Rust name mangling that's guaranteed and we don't actually need to give it a unique name.

If we really need a unique name, we may want to take a look at the` paste` crate.

crates/neon-runtime/src/napi/mod.rs
```diff
@@ -1,3 +1,5 @@
+use std::ptr;
```
kjvalencik: Need to delete these. Leftover from when module initialization was here.

src/lib.rs
```diff
@@ -31,15 +31,44 @@ use cfg_if::cfg_if;
 
 cfg_if! {
     if #[cfg(feature = "n-api")] {
+        #[doc(hidden)]
+        pub use neon_runtime::sys as sys;
```
kjvalencik: This will go away once we aren't leaking `sys` types into `register_module`.

PR #449, opened 2019-10-02
Basic N-API tests

dherman:
Adds N-API tests to the test suite. Currently just has one "hello world" test to show modules can be loaded, since that's all we've implemented so far.


dherman:
Oh whoops! This fails CI for Node 8, which is expected. I'll add logic to disable the N-API tests for Node < 10.

PR #454, opened 2019-10-20
Fix ERR_CHILD_PROCESS_STDIO_MAXBUFFER error

splix:
After upgrading to 0.3.2 I started to get following error:
```
$ neon build --release
neon ERR! stdout maxBuffer length exceeded

RangeError [ERR_CHILD_PROCESS_STDIO_MAXBUFFER]: stdout maxBuffer length exceeded
    at Socket.onChildStdout (child_process.js:348:14)
    at Socket.emit (events.js:189:13)
    at addChunk (_stream_readable.js:284:12)
    at readableAddChunk (_stream_readable.js:261:13)
    at Socket.Readable.push (_stream_readable.js:220:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:94:17)
```

I use NodeJS 10 LTS which has only 200kb, and it seems to be not enough. NodeJS 12 has 1Mb by default, and it works fine on another machine.

PR increases the buffer to 1Mb and fixes error for my case

### DIFF
--- a/cli/src/project.ts
+++ b/cli/src/project.ts
@@ -27,7 +27,8 @@ export default class Project {
   static async create(root: string, options: Partial<ProjectOptions> = {}): Promise<Project> {
     let { crate = 'native' } = options;
     const { stdout } = await execFile("cargo", ["metadata", "--format-version=1"], {
-      cwd: path.join(root, crate)
+      cwd: path.join(root, crate),
+      maxBuffer: 1024 * 1024
     });
     const targetDirectory: string = JSON.parse(stdout).target_directory;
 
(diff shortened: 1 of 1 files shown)


cli/src/project.ts
```diff
@@ -27,7 +27,8 @@ export default class Project {
   static async create(root: string, options: Partial<ProjectOptions> = {}): Promise<Project> {
     let { crate = 'native' } = options;
     const { stdout } = await execFile("cargo", ["metadata", "--format-version=1"], {
```
kjvalencik: Does adding `--no-deps` without increasing the buffer size resolve the issue?

The dependency isn't necessary, only the target path is used.

splix: Yes, just checked, it also helps

splix: I can modify PR to use --no-deps instead

kjvalencik: Please do. It should be faster and make the larger buffer unnecessary.

splix: done

PR #458, opened 2019-10-25
N-API Initial Patterns

kjvalencik:


crates/neon-runtime/src/napi/object.rs
```diff
@@ -1,18 +1,55 @@
...
+
+    *out = true;
+
+    if napi::napi_create_string_utf8(
+        env,
+        key as *const i8,
+        len as usize,
+        key_val.as_mut_ptr(),
+    ) != napi::napi_status::napi_ok {
+        *out = false;
```
kjvalencik: I didn't quite understand the failures in the `nan` implementation that might cause `out` and the return value to be different. They are always the same in the `n-api` implementation and that may or may not be logically the same.

crates/neon-runtime/src/napi/string.rs
```diff
@@ -1,7 +1,47 @@
...
+    len.assume_init() as isize
+}
+
+pub unsafe extern "C" fn data(env: Env, out: *mut u8, len: isize, value: Local) -> isize {
+    let mut read = MaybeUninit::uninit();
+    let status = napi::napi_get_value_string_utf8(
+        env,
+        value,
+        out as *mut i8,
+        len as usize,
```
kjvalencik: There's a ton of casting going on in these string methods. Should we eliminate the casting in the `runtime` methods and branch further in the higher level `neon` methods?

My intuition is that we should keep them the same and flag them for follow-up since `runtime` is an internal crate.

crates/neon-sys/src/lib.rs
```diff
@@ -18,8 +18,14 @@ pub struct Local {
 /// the function is bound to and a mechanism to return a value to the caller.
 pub type FunctionCallbackInfo = c_void;
 
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct Isolate__ {
```
kjvalencik: This is the pattern that `bindgen` uses to give raw pointers specific times instead of `c_void`. It works really well for letting the borrow checker find the mismatches between `node_env` and `Isolate`.

src/context/internal.rs
```diff
@@ -10,17 +10,30 @@ use object::class::ClassMap;
 use result::NeonResult;
 use super::ModuleContext;
 
+#[cfg(feature = "legacy-runtime")]
 #[repr(C)]
 #[derive(Clone, Copy)]
-pub struct Isolate(*mut raw::Isolate);
+pub struct Env(raw::Isolate);
```
kjvalencik: Most of the `neon` code only uses the higher level `Env`, but `runtime` methods accept the result of `to_raw` which may either be an `Isolate` or `node_env`. This pattern lets us trust the type checker when implementing `runtime` with `napi`.

src/context/internal.rs
```diff
@@ -38,15 +51,21 @@ impl Isolate {
         unsafe { std::mem::transmute(ptr) }
     }
 
-    pub(crate) fn current() -> Isolate {
+    #[cfg(feature = "napi-runtime")]
+    pub(crate) fn current() -> Env {
+        panic!("Context::current() will not implemented with n-api")
```
kjvalencik: This method is impossible with `N-API`--there isn't an equivalent to `Isolate->GetCurrent()`. It will be the source of most breaking changes.

src/lib.rs
```diff
@@ -47,16 +47,24 @@ macro_rules! register_module {
...
-            _env: $crate::macro_internal::runtime::nodejs_sys::napi_env,
-            exports: $crate::macro_internal::runtime::nodejs_sys::napi_value
+            env: $crate::macro_internal::runtime::nodejs_sys::napi_env,
+            m: $crate::macro_internal::runtime::nodejs_sys::napi_value
         ) -> $crate::macro_internal::runtime::nodejs_sys::napi_value
         {
             // Suppress the default Rust panic hook, which prints diagnostics to stderr.
-            ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
+            // FIXME: Need this for debugging unimplemented panics
+            // ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
```
kjvalencik: I wasn't seeing the backtraces, even with `RUST_BACKTRACE=1` when the panic hook was enabled. This made finding the method that was `unimplemented!()` difficult.

Do you have a suggestion for this other than commenting it out?

src/object/mod.rs
```diff
@@ -2,70 +2,191 @@
...
 pub use self::class::{Class, ClassDescriptor};
+pub use self::internal::*;
 
-/// A property key in a JavaScript object.
-pub trait PropertyKey {
-    unsafe fn get_from(self, out: &mut raw::Local, obj: raw::Local) -> bool;
-    unsafe fn set_from(self, out: &mut bool, obj: raw::Local, val: raw::Local) -> bool;
-}
+#[cfg(feature = "legacy-runtime")]
+mod internal {
```
kjvalencik: I'm not super keen on this, but it seemed slightly better than adding an attribute to _everything_. Possible alternatives:

* `cfg-if`. This reduces some boilerplate and allows shared `use` lines, but also adds an extra level of indentation.
* Move the implementation to separate files. `internal_legacy.rs` and `internal_napi.rs`.

The name `internal` is inconsistent with other uses of `internal` which imply it's private. Do you have other name suggestions?

src/object/mod.rs
```diff
@@ -2,70 +2,191 @@
...
+    use neon_runtime::raw;
+    use handle::Handle;
+    use types::{Value, JsValue, JsArray, build};
+    use types::utf8::Utf8;
+    use context::Context;
+    use result::{NeonResult, JsResult, Throw};
+
+    /// A property key in a JavaScript object.
+    pub trait PropertyKey {
+        unsafe fn get_from<'c, C: Context<'c>>(
```
kjvalencik: The breaking change is these now require context. I'm not currently collecting a list of breaking changes anywhere. Do you think we should track this in a mardown doc (maybe in the repo) so that we don't need to pour over code to find them all when `n-api` is released?

src/types/mod.rs
```diff
@@ -508,10 +533,11 @@ impl<CL: Object> JsFunction<CL> {
               AS: IntoIterator<Item=Handle<'b, A>>
     {
         let mut args = args.into_iter().collect::<Vec<_>>();
-        let (isolate, argc, argv) = unsafe { prepare_call(cx, &mut args) }?;
+        let (argc, argv) = unsafe { prepare_call(cx, &mut args) }?;
+        let env = cx.env().to_raw();
         build(|out| {
             unsafe {
-                neon_runtime::fun::construct(out, isolate, self.to_raw(), argc, argv)
+                neon_runtime::fun::construct(out, env, self.to_raw(), argc, argv)
```
kjvalencik: FYI, most of the `isolate` -> `env` changes were mechanically done with find/replace to purge the term `Isolate` from `neon`.

test/napi/native/src/lib.rs
```diff
@@ -1,3 +1,10 @@
 use neon::prelude::*;
 
-register_module!(_, { });
+register_module!(mut cx, {
+    let s = cx.string("Hello, World!");
+
+    println!("{}", s.value(&mut cx));
```
kjvalencik: #Fix: This can be tested a little better if I then create a new `JsString` from the value and assert that they are equal in the mocha test.

src/lib.rs
```diff
@@ -47,16 +47,24 @@ macro_rules! register_module {
     ($module:pat, $init:block) => {
```
kjvalencik: While we are making breaking changes, I suggest changing this to use a more idiomatic lambda syntax pattern. I.e.,

```rust
register_module!(|cx| {
    // ...
})
```

src/lib.rs
```diff
@@ -47,16 +47,24 @@ macro_rules! register_module {
...
-            ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
+            // FIXME: Need this for debugging unimplemented panics
+            // ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
 
-            $init
+            fn __init_neon_module($module: $crate::context::ModuleContext) -> $crate::result::NeonResult<()> $init
+
+            $crate::macro_internal::initialize_module(
+                // FIXME: Can we do without these?
+                std::mem::transmute(env),
```
kjvalencik: I think this `transmute` is correct. It closely follows the pattern used [here](https://github.com/neon-bindings/neon/blob/838a3ab9302e4efc84065e640383be12bc81794e/src/object/class/mod.rs#L196-L198).

src/lib.rs
```diff
@@ -47,16 +47,24 @@ macro_rules! register_module {
...
+                std::mem::transmute(m),
```
kjvalencik: I'm not as confident in this one, but I believe that this is simply an explicit version of the implicit type conversion happening in this definition:

```rust
                    register_func: Option<extern "C" fn(
                        $crate::handle::Handle<$crate::types::JsObject>, *mut u8, *mut u8)>,
```

src/lib.rs
dherman: If we want it to be a dynamic decision we can create a Neon-specific environment variable (maybe `NEON_VERBOSE` or `NEON_DEV` or something) for debugging.

Or maybe what we should actually do is gather the backtrace metadata programmatically and when we convert the panic to a JS exception, serialize the metadata as JS values and attach it to the JS error object. But we could do this in a separate PR.

The reason I disabled the hook was that since we convert Rust panics to JS exceptions, we don’t want to pollute end users’ experiences with a bunch of low level Rust spew. For example, a CLI tool that happens to have a deeply nested dependency on a Neon module shouldn’t spit out Rust metadata when some internal failure occurs.

src/lib.rs
dherman: It seems fine to make it public and doc-hidden.

src/lib.rs
kjvalencik: What do you think about reusing the `RUST_BACKTRACE` environment variable instead of introducing a new one?

src/types/mod.rs
```diff
@@ -257,12 +257,23 @@ impl ValueInternal for JsString {
...
+    pub fn size<'a, C: Context<'a>>(self, cx: &mut C) -> isize {
+        let env = cx.env().to_raw();
+
+        unsafe {
+            neon_runtime::string::utf8_len(env, self.to_raw())
+        }
+    }
+
+    #[cfg(feature = "legacy-runtime")]
     pub fn value(self) -> String {
```
goto-bus-stop: would it make sense to also add a dummy Context argument here? then the tests and stuff could be run with either feature. Or is the idea to fully maintain back compat in the legacy runtime?

src/object/mod.rs
goto-bus-stop: separate files would make sense I think, given that almost the entire module is wrapped in feature gates. if keeping them in one file `mod implementation` would work a bit better than `mod internal`, imho

src/types/mod.rs
kjvalencik: The idea is to not make any breaking changes where we aren't adding features. We would like to keep all of the N-API work on the master branch, but don't want to introduce upgrade churn for users that aren't taking advantage.

Once N-API is complete, I would like to revisit some of these, but I think it makes sense for now to add technical debt to keep legacy APIs consistent.

src/lib.rs
```diff
@@ -51,9 +51,20 @@ macro_rules! register_module {
             m: $crate::macro_internal::runtime::nodejs_sys::napi_value
         ) -> $crate::macro_internal::runtime::nodejs_sys::napi_value
         {
+            // Uses the same environment variables and patterns as `std::backtrace`
+            // https://github.com/rust-lang/rust/blob/03a50ae9b87021d4a166c70d2c932f1cb0aa8f28/src/libstd/backtrace.rs#L165-L184
+            let backtrace_enabled = match ::std::env::var("RUST_LIB_BACKTRACE") {
```
dherman: I think where we left this was: let's do either a `NEON_xxx` environment variable, or a feature flag. I don't think I have a strong opinion which one, though. Probably a feature flag is the more conservative option, just a little less convenient.

src/context/internal.rs
kjvalencik: Yes, but, there are a bunch of places that it needs to be refactored out of (`src/context/mod.rs`) before it can be removed. I think it's better to leave that as a future PR.

crates/neon-runtime/src/napi/string.rs
kjvalencik: https://github.com/neon-bindings/neon/issues/472

crates/neon-runtime/src/napi/object.rs
kjvalencik: Since this is a trait and we don't have implementations across all methods, I'm flagging this for follow-up.

https://github.com/neon-bindings/neon/issues/472

PR #482, opened 2020-01-24
Singleton primitives

goto-bus-stop:
Implements `cx.null()`, `cx.undefined()`, and `cx.boolean(b: bool)` for the n-api runtime.

n-api requires the `napi_env` value to be passed in to the `napi_get_*` functions. This patch passes the `raw::Env` to the neon-runtime functions like `neon_runtime::primitive::null`. For the legacy runtime, I've added some dummy `Isolate*` parameters in neon-sys, so we don't automatically need feature flags everywhere `neon_runtime` is used. I assume this should be fine, since `neon_runtime` is not public API(?).

A `This::as_this` trait method implementation was creating a `JsUndefined` value, so that trait was updated to receive the `Env` value as well in the n-api runtime.

(I initially tried to avoid all the feature flag switching it requires with a second trait `ThisCompat` that would have an `impl <T: ThisCompat> This for T {}` to unify the API, but that blanket implementation was not possible without exporting `ThisCompat`, so I decided against it. There were not _that_ many places that `This::as_this` was implemented/used anyway :smile: )

### DIFF
--- a/crates/neon-runtime/src/napi/primitive.rs
+++ b/crates/neon-runtime/src/napi/primitive.rs
@@ -17,7 +17,13 @@ pub unsafe extern "C" fn boolean(out: &mut Local, env: Env, b: bool) {
     napi::napi_get_boolean(env, b, out as *mut Local);
 }
 
-pub unsafe extern "C" fn boolean_value(_p: Local) -> bool { unimplemented!() }
+/// Get the boolean value out of a `Local` object. If the `Local` object does not contain a
+/// boolean, this function panics.
+pub unsafe extern "C" fn boolean_value(env: Env, p: Local) -> bool {
+    let mut value = false;
+    assert_eq!(napi::napi_get_value_bool(env, p, &mut value as *mut bool), napi::napi_status::napi_ok);
+    value
+}
 
 // DEPRECATE(0.2)
 pub unsafe extern "C" fn integer(_out: &mut Local, _isolate: Env, _x: i32) { unimplemented!() }
--- a/crates/neon-runtime/src/napi/primitive.rs
+++ b/crates/neon-runtime/src/napi/primitive.rs
@@ -1,8 +1,16 @@
 use raw::{Local, Env};
 
-pub unsafe extern "C" fn undefined(_out: &mut Local) { unimplemented!() }
+use nodejs_sys as napi;
 
-pub unsafe extern "C" fn null(_out: &mut Local) { unimplemented!() }
+/// Mutates the `out` argument provided to refer to the global `undefined` object.
+pub unsafe extern "C" fn undefined(out: &mut Local, env: Env) {
+    napi::napi_get_undefined(env, out as *mut Local);
+}
+
+/// Mutates the `out` argument provided to refer to the global `null` object.
+pub unsafe extern "C" fn null(out: &mut Local, env: Env) {
+    napi::napi_get_null(env, out as *mut Local);
+}
 
 pub unsafe extern "C" fn boolean(_out: &mut Local, _b: bool) { unimplemented!() }
 
--- a/crates/neon-runtime/src/napi/primitive.rs
+++ b/crates/neon-runtime/src/napi/primitive.rs
@@ -12,7 +12,10 @@ pub unsafe extern "C" fn null(out: &mut Local, env: Env) {
     napi::napi_get_null(env, out as *mut Local);
 }
 
-pub unsafe extern "C" fn boolean(_out: &mut Local, _b: bool) { unimplemented!() }
+/// Mutates the `out` argument provided to refer to one of the global `true` or `false` objects.
+pub unsafe extern "C" fn boolean(out: &mut Local, env: Env, b: bool) {
+    napi::napi_get_boolean(env, b, out as *mut Local);
+}
 
 pub unsafe extern "C" fn boolean_value(_p: Local) -> bool { unimplemented!() }
 
--- a/crates/neon-sys/native/src/neon.cc
+++ b/crates/neon-sys/native/src/neon.cc
@@ -68,7 +68,7 @@ extern "C" void Neon_Primitive_Null(v8::Local<v8::Primitive> *out, v8::Isolate *
   *out = Nan::Null();
 }
 
-extern "C" void Neon_Primitive_Boolean(v8::Local<v8::Boolean> *out, bool b) {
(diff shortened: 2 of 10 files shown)


crates/neon-runtime/src/napi/primitive.rs
```diff
@@ -1,10 +1,21 @@
...
 
-pub unsafe extern "C" fn undefined(_out: &mut Local) { unimplemented!() }
-
-pub unsafe extern "C" fn null(_out: &mut Local) { unimplemented!() }
-
-pub unsafe extern "C" fn boolean(_out: &mut Local, _b: bool) { unimplemented!() }
+use nodejs_sys as napi;
+
+/// Mutates the `out` argument provided to refer to the global `undefined` object.
+pub unsafe extern "C" fn undefined(out: &mut Local, env: Env) {
```
kjvalencik: What are your thoughts on making `env` the first argument? I noticed that I was inconsistent in my previous PR.

I think where possible, we should prefer to match the `napi` methods and have it first. What do you think?

crates/neon-sys/native/src/neon.cc
```diff
@@ -60,15 +60,15 @@ extern "C" void *Neon_Object_GetIsolate(v8::Local<v8::Object> obj) {
   return obj->GetIsolate();
 }
 
-extern "C" void Neon_Primitive_Undefined(v8::Local<v8::Primitive> *out) {
+extern "C" void Neon_Primitive_Undefined(v8::Local<v8::Primitive> *out, v8::Isolate *isolate) {
```
kjvalencik: :+1: This is great, thanks. We're moving forward treating `neon-sys` as an internal API, therefore, this is the preferred approach to limit branching elsewhere.

src/context/mod.rs
```diff
@@ -476,9 +490,16 @@ impl<'a, T: This> CallContext<'a, T> {
...
 
     /// Produces a handle to the `this`-binding.
+    #[cfg(feature = "legacy-runtime")]
     pub fn this(&mut self) -> Handle<'a, T> {
         Handle::new_internal(T::as_this(self.info.this(self)))
     }
+
+    /// Produces a handle to the `this`-binding.
+    #[cfg(feature = "napi-runtime")]
+    pub fn this(&mut self, env: Env) -> Handle<'a, T> {
```
kjvalencik: Does this need an `env` parameter or can it be obtained with `self.env()`?

src/object/class/mod.rs
```diff
@@ -119,9 +119,15 @@ pub trait Class: Managed + Any {
 }
 
 unsafe impl<T: Class> This for T {
+    #[cfg(feature = "legacy-runtime")]
     fn as_this(h: raw::Local) -> Self {
         Self::from_raw(h)
     }
+
+    #[cfg(feature = "napi-runtime")]
+    fn as_this(_env: Env, h: raw::Local) -> Self {
```
kjvalencik: Why does this take an `env`?

test/napi/lib/hello.js
```diff
@@ -6,4 +6,11 @@ describe('hello', function() {
     assert.equal(addon.greeting, "Hello, World!");
     assert.equal(addon.greeting, addon.greetingCopy);
   });
+
+  it('should export global singletons for JS primitives', function () {
+    assert.equal(addon.undefined, undefined);
```
kjvalencik: Can you add an additional assert that `addon.hasOwnProperty('undefined')`? Otherwise, this would still pass with code removed from `lib.rs`.

src/types/mod.rs
```diff
@@ -503,8 +539,8 @@ impl JsFunction {
               U: Value
     {
         build(|out| {
+            let env = cx.env().to_raw();
```
kjvalencik: :tada:

src/context/mod.rs
```diff
@@ -251,15 +251,29 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     /// Convenience method for creating a `JsNull` value.
+    #[cfg(feature = "legacy-runtime")]
```
kjvalencik: Since these methods have identical function signatures, we also have the option of putting the feature flag inside the method. I'm undecided on which is cleaner. Thoughts?

_Note: This comment is applicable to multiple lines._

goto-bus-stop: Yeah, I was looking at that too, but AFAIK you have to put things in a `{ block }` to put an attribute on them:
```rust
fn null(&mut self) -> Handle<'a, JsNull>
	#[cfg(feature = "legacy-runtime")]
    {
        JsNull::new()
    }
	#[cfg(feature = "napi-runtime")]
    {
        JsNull::new(self)
    }
}
```
and that just looked ugly :upside_down_face: 

Saying that, while it looks ugly, it might be actually be clearer.

crates/neon-runtime/src/napi/primitive.rs
goto-bus-stop: Yes, I definitely prefer taking `env` as the first parameter, but did it like this to match the API of the few legacy runtime functions that took an `Isolate` as the second parameter. A good followup PR would be to swap everything

src/context/mod.rs
goto-bus-stop: Ohh good call. I used the `ctx.env()` method elsewhere but forgot about it here.

src/object/class/mod.rs
goto-bus-stop: The trait method takes an `env` so `impl This for JsUndefined` can use it to call the `neon_runtime::primitive::undefined()` function.

src/context/mod.rs
goto-bus-stop: Ok I think it's better if we don't have to manually check if the signature is identical when working with these methods in the future, so I moved the attributes inside the function. With an explicit `return` we can make it quite nice I think:
```rust
#[cfg(feature = "legacy-runtime")]
return JsNull::new();
#[cfg(feature = "napi-runtime")]
return JsNull::new(self);
```

test/napi/lib/hello.js
```diff
@@ -6,4 +6,12 @@ describe('hello', function() {
...
+    assert.ok(addon.hasOwnProperty('undefined'));
+    assert.equal(addon.null, null);
+    assert.equal(addon.true, true);
+    assert.equal(addon.false, false);
```
kjvalencik: Sorry that I missed this before, but these should likely be `strictEqual`.

goto-bus-stop: ah yes :) i should've looked at the chai docs instead of just copying the lines above!

PR #483, opened 2020-01-27
Workaround for Windows node-gyp bug

dherman:
See https://github.com/nodejs/node-gyp/issues/1933

This downgrades npm to the last version that didn't have the bug for Windows CI, to avoid hitting the bug. Once the above bug is fixed we can remove this workaround.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,7 +431,7 @@ mod tests {
     fn static_test() { static_test_impl() }
 
     #[rustversion::beta]
-    #[not(cfg(feature = "enable-static-tests"))]
+    #[cfg(not(feature = "enable-static-tests"))]
     #[test]
     #[ignore]
     fn static_test() { static_test_impl() }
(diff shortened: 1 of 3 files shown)


dherman:
Ok good, that’s working... now it’s just the compile tests that aren’t. I’ll disable them for now until we have a less fragile way to test compiler error messages.

src/lib.rs
```diff
@@ -426,9 +426,16 @@ mod tests {
     // and any associated usability regressions before a new Rust version is shipped
     // but will have more stable results than Nightly.
     #[rustversion::beta]
+    #[cfg(feature = "enable-static-tests")]
     #[test]
     fn static_test() { static_test_impl() }
```
kjvalencik: If I wanted to run just this test, I would use the following command, right?

```sh
cargo test --features enable-static-tests static-test
```

dherman: Right, exactly.

dherman: Maybe I should add a compile error if someone enables the feature with a Rust toolchain other than beta:

```rust
#[rustversion:not(beta)]
#[cfg(feature = "enable-static-tests")]
compile_error!("The `enable-static-tests` feature can only be enabled with a beta Rust toolchain");
```

dherman: Added that check.

PR #447, opened 2019-09-29
Upgrade cli template to rust 2018

theJian:
Resolve #398

Issue #398, opened 2019-03-23
Update template to rust 2018


theJian:
I'm working on updating neon examples to rust 2018(https://github.com/neon-bindings/examples/issues/23). I can update neon project template to rust 2018 correspondingly. What do you think?

dherman:
I love this idea! Are you thinking we could change the `Cargo.toml.hbs` to include the `"2018"` Edition and the `lib.rs.hbs` to remove this boilerplate?
```rust
#[macro_use]
extern crate neon;
```

theJian:
@dherman That will be nice! We can use `use neon::prelude::*` to import all we need just like `test/dynamic/native/src/lib.rs`

andradei:
Would love to see the PR for this merged :)

PR #484, opened 2020-02-03
remove unnecessary macro from lib.rs template

amilajack:
Creating a new neon project will always result in the following warnings:

```
warning: unused `#[macro_use]` import
 --> src/lib.rs:1:1
  |
1 | #[macro_use]
  | ^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

    Finished dev [unoptimized + debuginfo] target(s) in 21.55s
```

To reproduce, run `neon new my-project` with `neon-cli@0.3.3`

Closes https://github.com/neon-bindings/neon/pull/447
Fixes https://github.com/neon-bindings/neon/issues/398

### DIFF
--- a/cli/templates/lib.rs.hbs
+++ b/cli/templates/lib.rs.hbs
@@ -1,5 +1,3 @@
-extern crate neon;
-
 use neon::prelude::*;
 
 fn hello(mut cx: FunctionContext) -> JsResult<JsString> {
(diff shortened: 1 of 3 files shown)


kjvalencik:
@amilajack This should probably be coupled with adding `edition = "2018"` to `Cargo.toml.hbs` to help with backwards compatibility.

cli/templates/lib.rs.hbs
```diff
@@ -1,4 +1,3 @@
-#[macro_use]
 extern crate neon;
```
kjvalencik: This line can also be removed.

amilajack:
All tests should be passing now

Issue #490, opened 2020-02-18
Architecture read from node-gyp incorrectly


dbkr:
https://github.com/neon-bindings/neon/blob/master/crates/neon-sys/build.rs#L96 reads the architecture from `node-gyp configure`, but the value printed is just the architecture that node is running on, not the target architecture (https://github.com/nodejs/node-gyp/blob/3555ff4f0031f067bda9a49cbb2dbe6c04f4b5c7/bin/node-gyp.js#L45).

This means that building on an architecture other than that of the host machine's `node` binary will fail because the final link will use the node library path for the wrong architecture.

PR #491, opened 2020-02-18
Read architecture from node-gyp correctly

dbkr:
Reads the target architecture from the output of `node-gyp build` which prints the target architecture rather than the 'using' string that `node-gyp` prints (which is the platform it's running on).

This allows neon projects to build on platforms where the host's node binary is of a different architecture to the target architecture (eg. on Windows when building for x86 on an x64 host, even if using native x86 toolchains).

Fixes https://github.com/neon-bindings/neon/issues/490

### DIFF
--- a/crates/neon-sys/build.rs
+++ b/crates/neon-sys/build.rs
@@ -89,12 +89,12 @@ mod build {
         cmd
     }
 
-    // The node-gyp output includes platform information in a string
+    // The node-gyp build output includes platform information in a string
     // that looks like:
     //
-    //     gyp info using node@8.3.0 | win32 | x64
+    //     gyp verb architecture ia32
     fn parse_node_arch(node_gyp_output: &str) -> String {
-        let version_regex = Regex::new(r"node@(?P<version>\d+\.\d+\.\d+)\s+\|\s+(?P<platform>\w+)\s+\|\s(?P<arch>ia32|x64)").unwrap();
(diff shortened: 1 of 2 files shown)


crates/neon-sys/build.rs
```diff
@@ -164,18 +164,19 @@ mod build {
         }
 
         if cfg!(windows) {
-            let node_gyp_output = String::from_utf8_lossy(&output.stderr);
-            println!("cargo:node_arch={}", parse_node_arch(&node_gyp_output));
```
kjvalencik: `node_gyp_output` no longer exists. Does it need to be passed into this method?

kjvalencik:
@dbkr Thanks for contributing a fix! Can you take a look at the build failures? Thanks!

crates/neon-sys/build.rs
dbkr: Oops - that line shouldn't have been deleted - fixed. (For complex dependency reasons I had to backport the fix to 0.3.x to test it and had to manually merge as the indenting level had changed, so must have missed this one).

dbkr:
Hi - is anything further needed from me on this? We'd like to start releasing builds soon so are very keen for #490 to be fixed, otherwise we'll have to find a workaround.

kjvalencik:
@dherman can you take a look at this? I'm not familiar enough with this code to review.

PR #498, opened 2020-03-13
Pass async_context to node::MakeCallback()

### LINKED ISSUE #418: Neon background task calls fail when run from the REPL if the domain module has been required
It appears that the following combination of things:
* the calling of a neon background task
* the `domain` module having been first required
* execution of the whole script from the REPL, or with `node -e`
* Node 10.x (or at least, 10.15.3 and 10.16.0)

causes the callback to fail to be called, and instead, for a weird, inscrutable error to be produced:
```
domain.js:119
    domain.enter();
           ^

TypeError: domain.enter is not a function
    at topLevelDomainCallback (domain.js:119:12)
```

To reproduce, create a new neon project and edit `src/lib.rs` to contain:
```
#[macro_use]
extern crate neon;

use neon::prelude::*;

struct BackgroundTask;

impl Task for BackgroundTask {
    type Output = ();
    type Error = String;
    type JsEvent = JsUndefined;

    fn perform(&self) -> Result<(), String> {
        println!("hello");
        Ok(())
    }

    fn complete(self, mut cx: TaskContext, _result: Result<(), String>) -> JsResult<JsUndefined> {
        Ok(cx.undefined())
    }
}

fn hello(mut cx: FunctionContext) -> JsResult<JsUndefined> {
    let f = cx.argument::<JsFunction>(0)?;
    BackgroundTask.schedule(f);
    Ok(cx.undefined())
}

register_module!(mut cx, {
    cx.export_function("hello", hello)
});
```

and edit `lib/index.js` to contain:
```
require('domain');
var addon = require('../native');

console.log(addon.hello(() => {}));
```

and then run it with `node -e "require('.')"`. Note: `node .` works fine; just the REPL or `node -e` causes the failure mode. Likewise, if you comment out the requiring of `domain`, it works when run either way. Sync functions are not affected, and the body of the async tasks runs; failure seems not to occur until we try to call the JS-side callback function.

Other reports about the `domain.enter is not a function` seem to suggest that it's an issue that comes up when running Node 10 in combination with native addons built with NAN versions older than 2.10.0, but neon currently uses 2.10.0, and #410 (which updates to a newer NAN) seems not to fix the issue.

Zireael-N:
This fixes #418 for me.

Changes are based on https://github.com/nodejs/nan/blob/60e86a9d4c4536fbbd2a76814470edd66b23114e/nan.h#L580-L614

### DIFF
--- a/crates/neon-sys/native/src/neon_task.h
+++ b/crates/neon-sys/native/src/neon_task.h
@@ -2,6 +2,8 @@
 #define NEON_TASK_H_
 
 #include <uv.h>
+#include <nan_new.h>
+#include <node_version.h>
 #include "neon.h"
 #include "v8.h"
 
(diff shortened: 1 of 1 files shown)


kjvalencik:
Thanks for the fix!

Issue #444, opened 2019-09-24
Quest: N-API Support


dherman:
Prepare, once and future Neon contributors, for our noblest quest yet!

![Pippin: Great! Where are we going?](https://media.giphy.com/media/w5x8GjarjZADm/giphy.gif)

We are going to port Neon to Node's new [N-API](https://nodejs.org/api/n-api.html)!

![Pippin: Huh?](https://media.giphy.com/media/mBWi28isKkAG4/giphy.gif)

I'll explain. N-API brings to Neon the promise of a stable, backwards-compatible ABI—binary compatibility across all future versions of Node.

This is a big deal.

Portability across Node versions means **Neon will finally be practical for publishing libraries, not just apps:** a few prebuilt binaries should be sufficient for all downstream customers to use your native library without ever knowing the difference.

![Pippin: Oh, I get it!](https://media0.giphy.com/media/zTzZsbf0XapZS/giphy.gif)

The stuff of legend, no?

# Our Quest

## Step 1. Create the feature flag

- [x] Create a [cargo feature flag](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) to allow us to concurrently maintain the main Neon codebase along with the experimental N-API support in the same master branch. ([Merged!](https://github.com/neon-bindings/neon/pull/440))
- [x] Set up a test suite specifically for the N-API backend so each task can easily include adding tests (#449)

## Step 2. Implement the port

- [x] **Module contexts and initialization:** Implement the `neon::context::ModuleContext` type and pass it to the module initialization function inside the `register_module!` macro defined in `/src/lib.rs`. The context struct will likely need to encapsulate the underlying `napi_env` and `napi_value` as private fields. This can be implemented before we implement functions, with an unimplemented `export_function()` method for now.
- [x] **Functions:** This is probably one of the subtler tasks. See the implementation of `neon::types::JsFunction::new()`. The Rust callback can be stored as the extra `void*` data passed to `napi_create_function`.
- [x] **Function arguments:** Implement `CallContext::len()` and `CallContext::argument()`.
- [x] **Function returns:** Implement function return values.
- [x] **`this`:** Implement `CallContext::this()`.
- [x] **Call kinds:** Implement `CallContext::kind()`.
- [x] **Function exports:** Once we have module contexts and functions implemented, we can implement the `ModuleContext::export_function()` shorthand method.
- [x] **Objects:** See `neon::types::JsObject::new()` and the `neon::object::Object` methods.
[...]

PR #505, opened 2020-03-25
Migrated error type to napi error

anshulrgoyal:
# Description
This pull request try to migrate nan error management to napi implementation using nodejs-sys as backend. The napi provide few methods for error which are documented below:
- `napi_throw` this throws an error.
- `napi_create_error` this create a new error.
- `napi_create_type_error` this create a new type error.
-  `napi_create_range_error` this create a new range error.

# Related Issue
#444


anshulrgoyal:
@dherman There is a problem with handling `panic!(error)` all the napi function require `napi_env` which is not available in the panic_hook.

.gitignore
```diff
@@ -10,3 +10,5 @@ cli/lib
 test/cli/lib
 npm-debug.log
 rls*.log
+.idea
```
kjvalencik: #Fix: Remove this line. I recommend creating a global git ignore.

https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

crates/neon-runtime/src/napi/error.rs
```diff
@@ -1,11 +1,23 @@
...
+use nodejs_sys as napi;
 
-pub unsafe extern "C" fn new_error(_out: &mut Local, _msg: Local) { unimplemented!() }
+pub unsafe extern "C" fn throw(env:Env,error: Local)->bool {
+    let status=napi::napi_throw(env,error);
+    status==napi::napi_status::napi_ok
+}
 
-pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimplemented!() }
+pub unsafe extern "C" fn new_error(out: &mut Local,env:Env,code:Local,msg:Local)->bool {
```
kjvalencik: #Fix: Please follow formatting of other functions.

src/context/mod.rs
```diff
@@ -115,22 +115,22 @@ impl<'a> Lock<'a> {
 }
 
 /// An _execution context_, which provides context-sensitive access to the JavaScript engine. Most operations that interact with the engine require passing a reference to a context.
-/// 
+///
```
kjvalencik: #Fix: Please remove unrelated formatting changes.

src/types/error.rs
```diff
@@ -36,6 +36,7 @@ impl Object for JsError { }
 
 impl JsError {
     /// Creates a direct instance of the [`Error`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Error) class.
+    #[cfg(feature = "legacy-runtime")]
```
kjvalencik: #Answer: Since these functions share a function signature, what do you think of putting the cfg flag on the body instead of on the function to ensure conformance?

src/types/error.rs
```diff
@@ -44,7 +45,23 @@ impl JsError {
         })
     }
 
+    #[cfg(feature = "napi-runtime")]
+    pub fn error<'a, C: Context<'a>, S: AsRef<str>>(cx: &mut C, msg: S) -> NeonResult<Handle<'a, JsError>> {
+        let (ptr, len) = if let Some(small) = Utf8::from(msg.as_ref()).into_small() {
```
kjvalencik: Are we able to re-use the `JsString` code for this?

anshulrgoyal:
@kjvalencik I am still stuck with panic hook problem. So the problem is that all napi function require napi_env as there first argument and since it is not available in panic hook i am not able to implement it properly.

kjvalencik:
Which piece of code are you referring to as the panic hook problem? Do you mean `convert_panics`?

crates/neon-runtime/src/napi/error.rs
```diff
@@ -1,11 +1,23 @@
-use raw::Local;
+use raw::{Env, Local};
 
-pub unsafe extern "C" fn throw(_val: Local) { unimplemented!() }
+use nodejs_sys as napi;
 
-pub unsafe extern "C" fn new_error(_out: &mut Local, _msg: Local) { unimplemented!() }
+pub unsafe extern "C" fn throw(env:Env,error: Local)->bool {
+    let status=napi::napi_throw(env,error);
+    status==napi::napi_status::napi_ok
```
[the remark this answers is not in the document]
anshulrgoyal: fixed

.gitignore
anshulrgoyal: fixed

crates/neon-runtime/src/napi/error.rs
anshulrgoyal: fixed

src/context/mod.rs
anshulrgoyal: fixed

src/types/error.rs
anshulrgoyal: fixed

src/types/error.rs
anshulrgoyal: JsString will create a struct and then i have to use to_raw on it is not ideal according to me

anshulrgoyal:
@kjvalencik would you moving forward with this pr?

anshulrgoyal:
> ~Which piece of code are you referring to as the panic hook problem? Do you mean `convert_panics`?~
> 
> I see the problem `convert_panics` requires an env, but, it was consumed by the call. We could potentially alias the raw pointer to `napi_env` via `to_raw()`, but, I would need to think about the implications and if that's safe.
> 
> I think at the very least it's not any _less_ safe because that's what `Isolate::current()` is implicitly doing.

I donot think aliasing will cause a problem since the `napi_env` is valid till the addon is mounted.

crates/neon-runtime/src/napi/error.rs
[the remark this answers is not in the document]
dherman: Let's still remove the `bool` return type from this function as @kjvalencik suggests. Once you've asserted that `status` is `napi_ok` it's not helpful to test that comparison a second time and then ignore the result anyway.

crates/neon-runtime/src/napi/error.rs
```diff
@@ -1,11 +1,24 @@
-use raw::Local;
+use raw::{Env, Local};
 
-pub unsafe extern "C" fn throw(_val: Local) { unimplemented!() }
+use nodejs_sys as napi;
 
-pub unsafe extern "C" fn new_error(_out: &mut Local, _msg: Local) { unimplemented!() }
+pub unsafe extern "C" fn throw(env:Env, error: Local)->bool {
+    let status=napi::napi_throw(env,error);
```
dherman, line 6: If you could reformat this to match the style of the rest of Neon that would make for better consistency. There's a lot of missing whitespace, for example surrounding punctuation like `=` and `,` and `->` and operators like `==`.

src/types/error.rs
```diff
@@ -82,3 +152,38 @@ pub(crate) fn convert_panics<T, F: UnwindSafe + FnOnce() -> NeonResult<T>>(f: F)
...
+pub(crate) fn convert_panics<
+    'a,
+    T,
+    C: Context<'a> + std::panic::UnwindSafe,
+    F: UnwindSafe + FnOnce(C) -> NeonResult<T>,
+>(
+    cx: C,
+    f: F,
+) -> NeonResult<T> {
+    let env = cx.env().to_raw();
```
dherman, line 166: This looks safe to me. The env is supposed to be valid for the duration of the function body, and the `catch_unwind` callback will only be invoked during that time.

crates/neon-runtime/src/napi/error.rs
```diff
@@ -1,11 +1,24 @@
...
+    assert_eq!(status, napi::napi_status::napi_ok);
+    status==napi::napi_status::napi_ok
+}
 
-pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimplemented!() }
+pub unsafe extern "C" fn new_error(out: &mut Local, env:Env, code:Local, msg:Local)->bool {
+    let status=napi::napi_create_error(env,code,msg,out);
```
dherman, line 12: Like `throw` above, all these functions (`new_error`, `new_type_error`, `new_range_error`) should also be converted to assert and not return `bool`.

src/object/class/internal.rs
```diff
@@ -65,7 +72,12 @@ impl Callback<()> for ConstructorCallCallback {
                 let data = info.data();
                 let kernel: fn(CallContext<JsValue>) -> JsResult<JsValue> =
                     mem::transmute(neon_runtime::class::get_call_kernel(data.to_raw()));
-                if let Ok(value) = convert_panics(|| { kernel(cx) }) {
+                #[cfg(feature = "napi-runtime")]
```
dherman, line 75: We can avoid a lot of code duplication in the next couple chunks by changing the signature of `convert_panics` in `"legacy-runtime"` to match the same signature as `convert_panics` in `"napi-runtime"`. Then we can eliminate a lot of the `#[cfg(...)]` conditional-compilation branches.

src/object/class/internal.rs
```diff
@@ -112,6 +131,16 @@ impl<T: Class> Callback<bool> for ConstructCallback<T> {
                 let data = info.data();
                 let kernel: fn(CallContext<T>) -> NeonResult<Option<Handle<JsObject>>> =
                     mem::transmute(neon_runtime::class::get_construct_kernel(data.to_raw()));
+                #[cfg(feature = "napi-runtime")]
```
dherman, line 134: This is one of the duplications we can eliminate by having a consistent signature for `convert_panics`.

src/types/error.rs
```diff
@@ -1,70 +1,140 @@
...
         let msg = cx.string(msg.as_ref());
+
+        #[cfg(feature = "napi-runtime")]
+        let (ptr, len) = if let Some(small) = Utf8::from(msg.as_ref()).into_small() {
+            small.lower()
+        } else {
+            return Err(Throw);
+        };
         build(|out| unsafe {
+            #[cfg(feature = "napi-runtime")]
```
dherman, line 59: You can collapse the three identical `#[cfg(...)]` branches into one by putting the three statements in a block:

```rust
#[cfg(feature = "napi_runtime")]
{
    let mut local: raw::Local = std::mem::zeroed();
    neon_runtime::string::new(&mut local, cx.env().to_raw(), ptr, len);
    neon_runtime::error::new_error(out, cx.env().to_raw(), std::ptr::null_mut(), local);
}
```

src/types/error.rs
```diff
@@ -1,70 +1,140 @@
...
```
dherman, line 86: Again, collapse the three identical `#[cfg(...)]` branches into one by grouping the three statements into a block.

src/types/error.rs
```diff
@@ -1,70 +1,140 @@
...
```
dherman, line 118: Again, collapse the three identical `#[cfg(...)]` branches into one by grouping the three statements into a block.

anshulrgoyal: done

kjvalencik:
@anshulrgoyal Thanks so much for your contribution! I'm sorry that I let this fall behind. I took the liberty of picking up your PR, resolving conflicts, and closing out the last couple of remaining issues.

Thanks so much! https://github.com/neon-bindings/neon/pull/542

PR #506, opened 2020-03-26
Fix #463 ; Debug build mode fixing in Windows

### LINKED ISSUE #463: No way to build without --release?
Hi guys :wave: Is it me or it's only possible to `neon build --release`?
If I run just `neon build` then I get this message `Neon only build with --release...`. What is the reason of that?

There are some conditional attributes in my code that depend on `debug_assertions`, so those don't work because of this. 

Response would be greatly appreciated, especially if soon.

usagi:
Related: https://github.com/neon-bindings/neon/issues/463#issuecomment-604243158

### DIFF
--- a/crates/neon-build/src/lib.rs
+++ b/crates/neon-build/src/lib.rs
@@ -22,6 +22,11 @@ cfg_if! {
             println!("cargo:rustc-link-search=native={}", &node_lib_path.display());
             println!("cargo:rustc-link-lib={}", &node_lib_file_path.file_stem().unwrap().to_str().unwrap());
 
+            if debug
+            {
+                println!("cargo:rustc-link-lib=msvcrtd");
+            }
+
             // Link `win_delay_load_hook.obj` for windows electron
             let node_runtime_env = "npm_config_runtime";
             println!("cargo:rerun-if-env-changed={}", node_runtime_env);
--- a/crates/neon-build/src/lib.rs
+++ b/crates/neon-build/src/lib.rs
@@ -22,7 +22,7 @@ cfg_if! {
             println!("cargo:rustc-link-search=native={}", &node_lib_path.display());
             println!("cargo:rustc-link-lib={}", &node_lib_file_path.file_stem().unwrap().to_str().unwrap());
 
-            if debug
+            if cfg!(target_env = "msvc") && debug
             {
                 println!("cargo:rustc-link-lib=msvcrtd");
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -352,9 +352,6 @@ macro_rules! neon_stringify {
     }
 }
 
-#[cfg(all(windows, not(neon_profile = "release")))]
-compile_error!("Neon only builds with --release. For tests, try `cargo test --release`.");
-
 #[cfg(test)]
 mod tests {
     extern crate rustversion;


crates/neon-build/src/lib.rs
```diff
@@ -22,6 +22,11 @@ cfg_if! {
             println!("cargo:rustc-link-search=native={}", &node_lib_path.display());
             println!("cargo:rustc-link-lib={}", &node_lib_file_path.file_stem().unwrap().to_str().unwrap());
 
+            if debug
+            {
+                println!("cargo:rustc-link-lib=msvcrtd");
```
kjvalencik: This doesn't appear to be impacting other compilers (tests are passing), but, just to be safe, can you add a `cfg` to target just `msvc`?

kjvalencik:
Were you able to reproduce the integer overflow from the linked issue without this fix?

crates/neon-build/src/lib.rs
usagi: 👍 Yes I agree to your suggestion. I'll try to fixup. Thank you.

usagi:
@kjvalencik Ah... unfortunately(or maybe a good news), I could not reproduce the `Error: internal error in Neon module: attempt to subtract with overflow` or any happenings. And I could not contact the original error reporter then I could not know the repro codes that happened the overflow error.

But, the fix works with [examples](https://github.com/neon-bindings/examples). And it's debuggable with msvc-debugger and lldb. Lines, variables, dynamic links, registers and step execution in a rust codes are looks good to me. 

I think maybe, that overflow error occered from the implementation difference of a memory allocation between msvcrt.lib and msvcrtd.lib .

---

I'll happy if the original reporter @anurbol would be share the repro code.

anurbol:
Hi @usagi, you've done great work, wow. I am sorry, I can neither share the code, because I forgot everything about this problem, and where the code is actually, nor I think it could be useful because the code was pretty huge and I never knew which exact part caused the problem.

d4hines:
Any more thoughts on this? We're eagerly awaiting debug builds on Windows!

dherman:
Thank you for this submission @usagi! I think we should merge it, even if we aren't 100% sure all the issues are gone since this is only for debug builds. If those issues start to surface again we can just try to fix them, but they shouldn't break production systems.

dherman:
@usagi After we merge this PR and publish a new Neon version, do people still need to use any of the manual steps in https://github.com/neon-bindings/neon/issues/463#issuecomment-603664370 or can they just do `neon build --debug`?

usagi:
@dherman Just do `neon build --debug` ( = `neon build`).

crates/neon-build/src/lib.rs
dherman: What is msvcrtd? Is that the debug mode of Microsoft's C runtime library?

dherman:
@usagi In retrospect, it seems worthwhile to add a comment to the line you added that links msvcrtd, since it's otherwise fairly mysterious. Would you be up for a quick followup PR to add the comment?

usagi:
@dherman I will PR in near future about what is the trick if I could find the time.

retep998:
This seems incorrect. Rust always uses the release CRT, so linking the debug CRT would conflict with that. Why was this solution preferred over simply building the C code in release on Windows as I did in [my PR from a year ago that was ignored](https://github.com/neon-bindings/neon/pull/400)?

dherman:
Oh wow, thank you @retep998, and my apologies for overlooking your contribution. That never feels good. We do our best to keep on top of things but we don't always live up to our intentions.

Maybe what we could do is, I can write a backout PR for this and then that should fix the merge conflicts for #400. That way we don't lose the history or attribution.

@retep998 Just to make sure I understand: how would the conflicts manifest -- would it result in link-time or runtime errors? Any idea why @usagi's tests didn't run into any errors? Also, do you happen to have any links to info where I can learn about how and why Rust uses the release CRT in Windows?

As always, I really appreciate you sharing your expertise in Windows and Rust!

retep998:
`libc` as depended on by `std` will always link to the release CRT: https://github.com/rust-lang/libc/blob/b61d8c617a782688f81ccb8fe25919c20dcd374e/src/windows/mod.rs#L238-L242

Depending on the order of inputs to the linker, this may result in all linked rust code using only the debug CRT, only the release CRT, or a mixture of both which could cause link errors or runtime UB. If the user depends on any other crates that also link to C code, then that C code may end up linking against the debug CRT despite having been built against the release CRT which could cause link errors or runtime UB. 

In the future Rust may gain the option to link to the debug CRT: https://github.com/rust-lang/rust/issues/39016

dherman:
Thanks @retep998! You saved us from some likely extremely hard to diagnose bugs down the road. I’ll submit a backout PR and then see if your PR can be merged.

dherman:
Submitted #515 to back this one out. @usagi Thank you for contributing this PR—even though this particular solution didn't end up being the one we'll use, it still spurred the right conversation and will get us to a fix for the issue.

Issue #512, opened 2020-04-27
neon cli: create new project in current dir


anton-iskryzhytskyi:
![image](https://user-images.githubusercontent.com/36485067/80346096-df2ae100-8872-11ea-999d-f635c1875cf8.png)
Hello!
Think it'd be a nice feature to add for neon CLI, to create a project in the current directory.

anton-iskryzhytskyi:
Created a [PR](https://github.com/neon-bindings/neon/pull/513)

PR #513, opened 2020-04-27
Create new project in current dir

anton-iskryzhytskyi:
This PR adds an ability to create new neon project in current dir

[Issue](https://github.com/neon-bindings/neon/issues/512)


dherman:
Thank you for this suggestion, @anton-iskryzhytskyi!

Since `new` implies the creation of a new directory, I don't think it's a good user experience to allow Neon to modify the contents of an existing directory—in particular, imagine how a user would feel if they chose an existing directory by accident and `neon new` polluted it with a bunch of new files, or worse, overwrite existing files. As precedent, `cargo new` doesn't allow using an existing directory.

We could change this to a new `neon init` command, though, which takes as its argument a directory that must already exist (including `.` as a valid option). But we should think through what to do if it comes across any existing files that it would overwrite. For precedent, it looks like `npm init` and `cargo init` silently overwrite files, but in both cases they don't touch as many files as `neon init` does. This probably deserves an RFC to document our thinking.

Any thoughts?

dherman:
Thank you again for this contribution, @anton-iskryzhytskyi! I am going to close for now, because we aren't sure what UI we would want for this use case. If you'd like to reopen with an RFC I'm always happy to discuss further!

PR #533, opened 2020-06-22
(wip) try_catch

kjvalencik:
Implementation of https://github.com/neon-bindings/rfcs/pull/29

### DIFF
--- a/crates/neon-sys/native/src/neon.cc
+++ b/crates/neon-sys/native/src/neon.cc
@@ -541,3 +541,19 @@ extern "C" void Neon_EventHandler_Delete(void * thread_safe_cb) {
     neon::EventHandler *cb = static_cast<neon::EventHandler*>(thread_safe_cb);
     cb->close();
 }
+
+extern "C" void* Neon_TryCatch_New() {
+  return new Nan::TryCatch();
+}
+
+extern "C" bool Neon_TryCatch_HasCaught(Nan::TryCatch *try_catch) {
+  return try_catch->HasCaught();
+}
+
+extern "C" void* Neon_TryCatch_Exception(Nan::TryCatch *try_catch) {
+  return *try_catch->Exception();
+}
+
+extern "C" void Neon_TryCatch_Delete(Nan::TryCatch *try_catch) {
+  try_catch->~TryCatch();
+}
--- a/src/result/mod.rs
+++ b/src/result/mod.rs
@@ -31,3 +31,16 @@ pub type JsResult<'b, T> = NeonResult<Handle<'b, T>>;
 pub trait JsResultExt<'a, V: Value> {
     fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, V>;
 }
+
+impl <'a, V, E> JsResultExt<'a, V> for Result<Handle<'a, V>, Handle<'a, E>>
+where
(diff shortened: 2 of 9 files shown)


crates/neon-sys/native/src/neon.cc
```diff
@@ -541,3 +541,19 @@ extern "C" void Neon_EventHandler_Delete(void * thread_safe_cb) {
     neon::EventHandler *cb = static_cast<neon::EventHandler*>(thread_safe_cb);
     cb->close();
 }
+
+extern "C" void* Neon_TryCatch_New() {
+  return new Nan::TryCatch();
```
dherman, line 546: Unfortunately this isn't going to work. :( a `TryCatch` [has to be allocated on the stack](https://v8docs.nodesource.com/node-8.16/d4/dc6/classv8_1_1_try_catch.html#af28a8ba7e1cb6e66bc2468faa4c86eec). There are a couple ways to do this, both of which are pretty annoying:
- Use a cross-language callback pattern, where you pass the Rust closure as a `void*` to C along with a Rust glue function that takes the Rust closure as an argument and calls in. Then in C you stack-allocate the `TryCatch` and invoke the Rust glue function.
- Allocate a `TryCatch` on the Rust stack by creating a parallel Rust struct type that's big enough to contain the sizeof the C++ `TryCatch`. For this you have to have some way at build time to determine the sizeof a `TryCatch` and use a macro or a const declaration to declare the Rust type. And then the constructor and destructor for that Rust type can call into `neon_runtime` with a pointer to `self` for the C++ code to explicitly call the C++ constructor and destructor on it. This is how `HandleScope` works, which also has to be stack allocated.

src/result/mod.rs
```diff
@@ -31,3 +31,16 @@ pub type JsResult<'b, T> = NeonResult<Handle<'b, T>>;
 pub trait JsResultExt<'a, V: Value> {
     fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, V>;
 }
+
+impl <'a, V, E> JsResultExt<'a, V> for Result<Handle<'a, V>, Handle<'a, E>>
```
dherman, line 35: I don't quite understand the use case for this. If I wrap something in a `try_catch` it's to prevent it from throwing. If I want to do something like map on a result that might have thrown, I already have a `Result` that I can do `map()` on and it'll throw.

Is this for cases where you want to map and do something back in the JS VM? Like:

```rust
cx.try_catch(do_something)
  .and_then(|v| {
      ... cx ... // do stuff in the JS VM
  })
  .or_throw(cx)
```

crates/neon-sys/native/src/neon.cc
kjvalencik: I believe that the reason they should be stack allocated is for nesting to work properly they need to be deallocated in the same order they were created.

RII also enforces this. I think deleting with `Drop` would also accomplish this.

With that said, I think your suggestion of passing the Rust closure to be invoked by C is best.

kjvalencik: Hmmm, panicking across FFI is undefined.

kjvalencik: Oh, this is neat. https://github.com/rust-lang/rust/pull/65646

PR #534, opened 2020-06-22
Fix #530: Cache function arguments in N-API

### LINKED ISSUE #530: Remove wasteful allocation in `neon_runtime::napi::call::get`
As of https://github.com/neon-bindings/neon/pull/507, getting a specific argument `N` in a function call internally creates a temporary Vec that N-API can store the arguments up to and includig `N`. It's done this way because `napi_get_cb_info` doesn't let you get a particular argument, but only a list of the arguments.

The end result is that code like:

```rust
cx.argument::<JsString>(1); // allocates vec![napi_value; 1]
cx.argument::<JsString>(2); // allocates vec![napi_value; 2]
cx.argument::<JsString>(3); // allocates vec![napi_value; 3]
```

There could be several ways to improve this:

- Use a fixed array for low numbers of arguments, like `[napi_value; 10]`, and only use a dynamic `Vec<>` when a higher `N` is requested. This avoids the allocations in the common case which are (probably) the most expensive.
- Get _all_ of the arguments in a `Vec<>` once, and store it in the FunctionCallbackInfo on the Rust side. This might be a bit harder but it would eliminate the waste entirely in the N-API runtime.
- Propose an N-API addition in Node.js core that can retrieve a single particular argument :innocent:
- Maybe some other way?

kjvalencik:


dherman:
Are you sure it's issue #121 that this fixes? From clicking the link, #121 is a PR that has to do with string representations.

dherman:
Looks like it's actually fixing #530?

kjvalencik:
Yes, that's the one. I have no idea where I got 121 from. I must have been looking at a different project.

PR #501, opened 2020-03-21
Fix #406

usagi:
- Fix `neon new` could not be complete in Windows
- Update packages

PR #537, opened 2020-07-03
fix: `neon new` not exiting on windows

### LINKED ISSUE #406: neon new and build failed on Node11.10.1 Win10
Simple Installed Rust 1.34 stable. Node 11.10.1 
Then execute like this
> neon new my_project
Press ^C at any time to quit.
? version 0.1.0
? description
? node entry point lib/index.js
? git repository
? author Mario
? email lijh01@mail.jj.cn
? license MIT
^C终止批处理操作吗(Y/N)? y  // I wait for at least 30 minutes without any response. So I Ctrl C to Quit.

Then I execute like this
> neon build

Then Got Error like this.
   Compiling neon-runtime v0.2.0
error: failed to run custom build command for `neon-runtime v0.2.0`
process didn't exit successfully: `C:\Work\Project\GoKart\CenterServer\Rust\rust_center\native\target\release\build\neon-runtime-b6860bb15546a35d\build-script-build` (exit code: 101)
--- stdout
'Skipping node-gyp installation as part of npm install.'
audited 184 packages in 2.313s
found 1 high severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
cargo:node_arch=x64

--- stderr
thread 'main' panicked at 'Couldn't find node_root_dir in node-gyp output.', src\libcore\option.rs:1038:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

neon ERR! cargo build failed

Error: cargo build failed
    at Target.<anonymous> (C:\Users\lijh01\AppData\Roaming\npm\node_modules\neon-cli\lib\target.js:121:35)
    at step (C:\Users\lijh01\AppData\Roaming\npm\node_modules\neon-cli\lib\target.js:32:23)
    at Object.next (C:\Users\lijh01\AppData\Roaming\npm\node_modules\neon-cli\lib\target.js:13:53)
    at fulfilled (C:\Users\lijh01\AppData\Roaming\npm\node_modules\neon-cli\lib\target.js:4:58)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)

I'm pretty sure that windows-built-tools is installed successfully.

amilajack:
Fixes #406
Closes https://github.com/neon-bindings/neon/pull/501

This bug fixes a bug where running `neon new` on windows will hang without ever exiting despite creating a new repo and all.

## Changes
* Migrates to use native fs module
* Bumps TS to latest version
* Adds `make-promises-safe` to CLI entrypoint to ensure promise do not silently fail
* Build targets ES6
* Removes node 8 from CI, adds node 14
* Removes `ts-unknown` and `es6-polyfill` dependencies
* Change methods to sync

## Testing this PR
Make sure you're on a windows machine to test this PR
```bash
git clone https://github.com/neon-bindings/neon test-repo
cd test-repo
npm i
npm run transpile
node bin/cli.js new foo
```

This fix is based on @usagi 's PR (https://github.com/neon-bindings/neon/pull/501). Thanks!


cli/src/target.ts
```diff
@@ -62,7 +63,10 @@ export default class Target {
 
   async clean() {
     // Remove the directory associated with this target.
-    await remove(path.resolve(this.crate.root, 'target', this.subdirectory));
+    const absolutePathSubdir = path.resolve(this.crate.root, 'target', this.subdirectory);
+    if (existsSync(absolutePathSubdir)) {
+      await unlink(absolutePathSubdir);
```
goto-bus-stop, line 68: `unlink` is for files or symlinks not directories, should we be using rimraf instead?

goto-bus-stop:
Other than the question about `unlink`, this looks good to me! Works as expected on my Win10 machine. Love the addition of `make-promises-safe` too :)

cli/src/target.ts
amilajack: Ah good catch. It seems like node has a [rmdir](https://nodejs.org/api/fs.html#fs_fs_rmdirsync_path_options) API. What are your thoughts on that?

goto-bus-stop: If a user is running `neon clean` there's probably still files in the `target` folder, so we need a recursive rmdir. The `recursive` variant of `fs.rmdir` is only supported in node 12 and up, so I think we should use the rimraf module until we drop node 10

amilajack: Makes sense. I'll update this PR to reflect that

cli/src/crate.ts
```diff
@@ -43,12 +43,16 @@ export default class Crate {
   }
 
   async finish(dylib: string) {
-    await remove(this.addon);
-    await copy(dylib, this.addon);
+    if (existsSync(this.addon)) {
```
dherman, line 46: Being very pedantic, this has a race condition: another process could create the file in between the `existsSync` check and the `unlink`. Another way to do this would be to just call `await unlink(this.addon)` and catch and ignore an `ENOENT` error. It's an unlikely race condition, though, not a big deal.

cli/src/target.ts
```diff
@@ -62,7 +66,10 @@ export default class Target {
 
   async clean() {
     // Remove the directory associated with this target.
-    await remove(path.resolve(this.crate.root, 'target', this.subdirectory));
+    const absolutePathSubdir = path.resolve(this.crate.root, 'target', this.subdirectory);
+    if (existsSync(absolutePathSubdir)) {
```
dherman, line 70: You don't need the `existsSync` check here; I'm pretty sure `rimraf` will silently succeed even if the directory already doesn't exist.

cli/src/crate.ts
dherman: Oh actually now that we're depending on `rimraf` we can just use that here.

PR #542, opened 2020-07-07
Migrated error type to napi error

kjvalencik:


src/types/error.rs
```diff
@@ -77,7 +77,9 @@ pub(crate) fn convert_panics<T, F: UnwindSafe + FnOnce() -> NeonResult<T>>(f: F)
             };
             let (data, len) = Utf8::from(&msg[..]).truncate().lower();
             unsafe {
-                neon_runtime::error::throw_error_from_utf8(data, len);
+                #[cfg(feature = "napi-runtime")]
+                neon_runtime::error::clear_exception(env.to_raw());
```
kjvalencik, line 81: The existing API defines panics as taking priority over exceptions. However, in `N-API` you may only throw an exception when not in a throwing state.

src/types/internal.rs
```diff
@@ -59,16 +59,11 @@ impl<T: Value> Callback<raw::Local> for FunctionCallback<T> {
...
                     value.to_raw()
                 } else {
-                    // What should we return if the function panicked?
-                    //
-                    // `ptr::null_mut()` may work, but we should have a test to verify that, which
-                    // can be created after [#505][0]. For now, let's not guess!
-                    //
-                    // [0]: https://github.com/neon-bindings/neon/pull/505.
-                    unimplemented!("cannot return from function after a panic")
+                    // https://github.com/nodejs/node/blob/16f3b6a01b44dea9a320dd9812cdf21e6b91234a/doc/api/util.md#utiltypesisexternalvalue
```
goto-bus-stop, line 65: good reference! i think this comment does need a little more explanation, because without the context of the old comment, it's not totally clear what readers should look at in the linked doc:
```suggestion
                    // Return null if the function threw, like in this example in the Node.js docs
                    // https://github.com/nodejs/node/blob/16f3b6a01b44dea9a320dd9812cdf21e6b91234a/doc/api/util.md#utiltypesisexternalvalue
```

dherman: I'm still struggling to understand this bit. (Partly this is because the whole `Callback` trait and `FunctionCallback` type are confusing by themselves, which is my own fault 😛). Here are some questions to show you my confusion—maybe they'll help you figure out how to comment this with some clarifications:

- How is `ptr::null_mut()` even type-compatible with `raw::Local`?
- What is the purpose of the `raw::Local` result? It's the return value of the JS function this implements, right?
- What does this have to do with that link? What does an External have to do with the return value of a function?

PR #543, opened 2020-07-07
try_catch

dherman:
This PR implements the `cx.try_catch` API of [RFC 29](https://github.com/neon-bindings/rfcs/pull/29). I had to make a few changes to the signature of `try_catch` to make the types work out; I'll update the RFC to match.

Still left to do:

- [x] hide the public API (and tests) behind a feature flag
- [x] add tests for nested `TryCatch` instances on the stack
- [x] add panic handler and tests to ensure panics get caught by try_catch
- [x] use `MaybeUninit` instead of `mem::zeroed()`
- [x] document the use of `Box` and the "unlikely" branch
- [x] fix the N-API tests
- [x] implement N-API backend
- [x] panic instead of returning `undefined` on unexpected `Err(Throw)`
- [x] fix merge conflicts
- [x] return `T: Value` instead of `JsValue`

### DIFF
--- a/crates/neon-runtime/src/napi/error.rs
+++ b/crates/neon-runtime/src/napi/error.rs
@@ -1,23 +1,11 @@
 use std::mem::MaybeUninit;
-use raw::{Env, Local};
+use std::ptr;
 
 use nodejs_sys as napi;
 
-pub unsafe extern "C" fn throw(env: Env, val: Local) {
-    let status = napi::napi_throw(env, val);
+use raw::{Env, Local};
 
-    assert_eq!(status, napi::napi_status::napi_ok);
-}
-
-pub unsafe extern "C" fn new_error(_out: &mut Local, _msg: Local) { unimplemented!() }
-
-pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimplemented!() }
-
-pub unsafe extern "C" fn new_range_error(_out: &mut Local, _msg: Local) { unimplemented!() }
-
-pub unsafe extern "C" fn throw_error_from_utf8(_msg: *const u8, _len: i32) { unimplemented!() }
-
-pub unsafe extern "C" fn is_throwing(env: Env) -> bool {
+pub unsafe fn is_throwing(env: Env) -> bool {
     let mut b: MaybeUninit<bool> = MaybeUninit::zeroed();
 
     let status = napi::napi_is_exception_pending(env, b.as_mut_ptr());
@@ -27,7 +15,7 @@ pub unsafe extern "C" fn is_throwing(env: Env) -> bool {
     b.assume_init()
 }
 
-pub unsafe extern "C" fn catch_error(env: Env, error: *mut Local) -> bool {
+pub unsafe fn catch_error(env: Env, error: *mut Local) -> bool {
     if !is_throwing(env) {
         return false;
     }
@@ -37,4 +25,92 @@ pub unsafe extern "C" fn catch_error(env: Env, error: *mut Local) -> bool {
     assert_eq!(status, napi::napi_status::napi_ok);
 
     true
+}
+
+pub unsafe fn clear_exception(env: Env) {
+    let mut result = MaybeUninit::uninit();
+    let status = napi::napi_is_exception_pending(env, result.as_mut_ptr());
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    if !result.assume_init() {
+        return;
+    }
+
+    let mut result = MaybeUninit::uninit();
+    let status = napi::napi_get_and_clear_last_exception(env, result.as_mut_ptr());
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+}
+
+pub unsafe fn throw(env: Env, val: Local) {
+    let status = napi::napi_throw(env, val);
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+}
+
+pub unsafe fn new_error(env: Env, out: &mut Local, msg: Local) {
+    let mut result = MaybeUninit::uninit();
+    let status = napi::napi_create_error(
+        env,
+        ptr::null_mut(),
+        msg,
+        result.as_mut_ptr(),
+    );
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    *out = result.assume_init();
+}
+
+pub unsafe fn new_type_error(env: Env, out: &mut Local, msg: Local) {
+    let mut result = MaybeUninit::uninit();
+    let status = napi::napi_create_type_error(
+        env,
+        ptr::null_mut(),
+        msg,
+        result.as_mut_ptr(),
+    );
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    *out = result.assume_init();
+}
+
+pub unsafe fn new_range_error(env: Env, out: &mut Local, msg: Local) {
+    let mut result = MaybeUninit::uninit();
+    let status = napi::napi_create_range_error(
+        env,
+        ptr::null_mut(),
+        msg,
+        result.as_mut_ptr(),
+    );
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    *out = result.assume_init();
+}
+
+pub unsafe fn throw_error_from_utf8(env: Env, msg: *const u8, len: i32) {
+    let mut out = MaybeUninit::uninit();
+    let status = napi::napi_create_string_utf8(
+        env,
+        msg as *const i8,
+        len as usize,
+        out.as_mut_ptr(),
+    );
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    let mut err = MaybeUninit::uninit();
+    let status = napi::napi_create_error(
+        env,
+        ptr::null_mut(),
+        out.assume_init(),
+        err.as_mut_ptr(),
+    );
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    throw(env, err.assume_init());
 }
--- a/crates/neon-runtime/src/napi/error.rs
+++ b/crates/neon-runtime/src/napi/error.rs
@@ -1,6 +1,13 @@
-use raw::Local;
+use std::mem::MaybeUninit;
+use raw::{Env, Local};
 
-pub unsafe extern "C" fn throw(_val: Local) { unimplemented!() }
+use nodejs_sys as napi;
+
+pub unsafe extern "C" fn throw(env: Env, val: Local) {
+    let status = napi::napi_throw(env, val);
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+}
 
 pub unsafe extern "C" fn new_error(_out: &mut Local, _msg: Local) { unimplemented!() }
 
@@ -9,3 +16,25 @@ pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimple
 pub unsafe extern "C" fn new_range_error(_out: &mut Local, _msg: Local) { unimplemented!() }
 
 pub unsafe extern "C" fn throw_error_from_utf8(_msg: *const u8, _len: i32) { unimplemented!() }
+
+pub unsafe extern "C" fn is_throwing(env: Env) -> bool {
+    let mut b: MaybeUninit<bool> = MaybeUninit::zeroed();
+
+    let status = napi::napi_is_exception_pending(env, b.as_mut_ptr());
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    b.assume_init()
+}
+
+pub unsafe extern "C" fn catch_error(env: Env, error: *mut Local) -> bool {
+    if !is_throwing(env) {
+        return false;
+    }
+
+    let status = napi::napi_get_and_clear_last_exception(env, error);
+
+    assert_eq!(status, napi::napi_status::napi_ok);
+
+    true
+}
--- a/crates/neon-sys/native/src/neon.cc
+++ b/crates/neon-sys/native/src/neon.cc
@@ -554,6 +554,9 @@ extern "C" bool Neon_TryCatch_With(Neon_TryCatchGlue glue_fn, void *cx, void *ru
     if (ok) {
       *result = ok_result;
     } else {
+      // It's unlikely but possible that a Neon user might return `Err(Throw)` even
+      // though the VM is not actually in a throwing state. In this case we will
+      // simply produce the JS undefined value.
       *result = Nan::Undefined();
     }
     return false;
--- a/crates/neon-sys/native/src/neon.cc
+++ b/crates/neon-sys/native/src/neon.cc
@@ -541,3 +541,24 @@ extern "C" void Neon_EventHandler_Delete(void * thread_safe_cb) {
     neon::EventHandler *cb = static_cast<neon::EventHandler*>(thread_safe_cb);
     cb->close();
 }
+
+extern "C" bool Neon_TryCatch_With(Neon_TryCatchGlue glue_fn, void *cx, void *rust_thunk, v8::Local<v8::Value> *result) {
+  Nan::TryCatch try_catch;
+
+  bool ok;
+  v8::Local<v8::Value> ok_result;
+
+  glue_fn(rust_thunk, cx, &ok, &ok_result);
+
+  if (!try_catch.HasCaught()) {
+    if (ok) {
+      *result = ok_result;
+    } else {
+      *result = Nan::Undefined();
+    }
+    return false;
+  } else {
+    *result = try_catch.Exception();
+    return true;
+  }
+}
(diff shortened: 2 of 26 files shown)


src/context/internal.rs
```diff
@@ -116,6 +117,37 @@ pub trait ContextInternal<'a>: Sized {
 
     fn activate(&self) { self.scope_metadata().active.set(true); }
     fn deactivate(&self) { self.scope_metadata().active.set(false); }
+
+    fn try_catch_internal<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
+        where F: FnOnce(&mut Self) -> JsResult<'b, JsValue>
+    {
+        let p = Box::into_raw(Box::new(f)) as *mut c_void;
+        let mut local: raw::Local = unsafe { std::mem::zeroed() };
```
kjvalencik, line 125: I recommend `Maybeuninit::zeroed()` instead.

src/context/internal.rs
```diff
@@ -116,6 +117,37 @@ pub trait ContextInternal<'a>: Sized {
...
+    where C: ContextInternal<'a>,
+          F: FnOnce(&mut C) -> JsResult<'b, JsValue>
+{
+    let f: F = *unsafe { Box::from_raw(p as *mut F) };
+
+    // FIXME: convert_panics(f)
+    match f(unsafe { std::mem::transmute(cx) }) {
+        Ok(result) => unsafe {
+            *ok = true;
+            *ok_value = result.to_raw();
```
kjvalencik, line 145: Oh, this is awesome! :heart_eyes: To avoid heap allocating the result of the closure, you allocate it on the stack up front and thread a pointer to the stack location through FFI. This is a nice pattern.

src/context/internal.rs
```diff
@@ -116,6 +117,37 @@ pub trait ContextInternal<'a>: Sized {
...
+    }
+}
+
+extern "C" fn try_catch_glue<'a, 'b: 'a, C: ContextInternal<'a>, F>(p: *mut c_void, cx: *mut c_void, ok: *mut bool, ok_value: *mut raw::Local)
+    where C: ContextInternal<'a>,
+          F: FnOnce(&mut C) -> JsResult<'b, JsValue>
+{
+    let f: F = *unsafe { Box::from_raw(p as *mut F) };
+
+    // FIXME: convert_panics(f)
```
kjvalencik, line 141: This is synchronous code, won't panics be caught by the `convert_panics` in `invoke` from the caller?

In my opinion, I don't think `try_catch` should be used to catch panics--which converting panics here would allow. I think it should only catch exceptions in V8.

tl;dr-- I don't think this is necessary or should be added.

crates/neon-sys/native/src/neon.cc
```diff
@@ -541,3 +541,24 @@ extern "C" void Neon_EventHandler_Delete(void * thread_safe_cb) {
...
+  bool ok;
+  v8::Local<v8::Value> ok_result;
+
+  glue_fn(rust_thunk, cx, &ok, &ok_result);
+
+  if (!try_catch.HasCaught()) {
+    if (ok) {
+      *result = ok_result;
+    } else {
+      *result = Nan::Undefined();
```
kjvalencik, line 557: This should be impossible, right? It would be mean having a `Err(Throw)` without being in a throwing state. If I understand correctly, maybe some equivalent to `unreachable!` instead?

src/context/internal.rs
```diff
@@ -116,6 +117,37 @@ pub trait ContextInternal<'a>: Sized {
 
     fn activate(&self) { self.scope_metadata().active.set(true); }
     fn deactivate(&self) { self.scope_metadata().active.set(false); }
+
+    fn try_catch_internal<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
+        where F: FnOnce(&mut Self) -> JsResult<'b, JsValue>
+    {
+        let p = Box::into_raw(Box::new(f)) as *mut c_void;
```
kjvalencik, line 124: I haven't had a chance to mess around with it, but, I _think_ the closure can be unboxed.

crates/neon-sys/native/src/neon.cc
dherman: I think it's possible because we rely on Neon users cooperatively only using `Err(Throw)` when a throw has actually happened, but AFAIK nothing stops them from lying.

src/context/internal.rs
dherman: It doesn't seem obvious to me what the best behavior is here, but I think I'm fine with letting the panic keep unwinding all the way to the top level of the Neon function.

I think I should use `catch_unwind` and signal back to Rust to continue panicking though, right? I think you were saying that it's undefined behavior for Rust to panic across an FFI boundary?

src/context/internal.rs
dherman: Yeah, I wasn't confident about this. I'll ask around to find out if extern functions can be safely cast to and from `void *`.

src/context/internal.rs
kjvalencik: It was in the Unsafe Rust guide that I was reading. But, the comment might be dated. I've since found this:

https://github.com/rust-lang/rust/pull/65646

crates/neon-sys/native/src/neon.cc
kjvalencik: Ah, a user could construct `Err(Throw)` manually. C++20 has an `[[unlikely]]` attribute. I'm not sure if that can be used here to any effect.

src/context/internal.rs
dherman: I've sometimes heard this pattern referred to as "destination-passing style."

src/context/internal.rs
dherman: This comment in the PR:

> I haven't updated any of the documentation, so officially unwinding through FFI is still UB. However this is a step towards making it well-defined.

suggests to me it's still not there yet. To be safe, I'll just catch any panics inside the inner Rust callback, return the boxed panic value across the FFI boundary, and re-panic from the outer `try_catch`. That way it avoids any undefined behavior but `try_catch` doesn't catch panics, as you suggest. I agree with your point about bad behavior -- people shouldn't be using Rust panics like exceptions and we shouldn't be encouraging it.

crates/neon-sys/native/src/neon.cc
dherman: > C++20 has an `[[unlikely]]` attribute. I'm not sure if that can be used here to any effect.

Interesting point, but I'll leave that for a future optimization if we can come up with benchmarks to show it helps. My guess is it doesn't matter because `try_catch` is already probably pretty slow, and it seems like a new enough feature that it might not be fully portable yet across C++ compilers.

src/context/internal.rs
dherman: OK, found it: [this passage in the unsafe code guidelines](https://rust-lang.github.io/unsafe-code-guidelines/layout/function-pointers.html#representation) indicates that a function can always be safely cast to a pointer:

> The ABI and layout of `(unsafe)? (extern "ABI")? fn(Args...) -> Ret` is exactly that of the corresponding C type -- the lack of a null value does not change this. On common platforms, this means that `*const ()` and `fn(Args...) -> Ret` have the same ABI and layout. This is, in fact, guaranteed by POSIX and Windows.

I double-checked with a Rust core team member, too, so we're good :)

dherman: Oh wait, this is an `FnOnce` not an `fn`. That might be different… checking with Niko…

dherman: OK, yeah, a closure is in fact not guaranteed to have any particular layout. I'll add a comment explaining why the boxing is necessary.

src/context/internal.rs
```diff
@@ -122,6 +122,8 @@ pub trait ContextInternal<'a>: Sized {
     fn try_catch_internal<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
         where F: FnOnce(&mut Self) -> JsResult<'b, JsValue>
     {
+        // A closure does not have a guaranteed layout, so we need to box it in order to pass
+        // a pointer to it across the boundary into C++.
```
kjvalencik, line 126: I don't think I understand this. Isn't the pointer carried through back to Rust where the layout would be known? The static `fn` is generic in the closure.

Pretty old. I was having trouble finding more recent examples.
http://aatch.github.io/blog/2015/01/17/unboxed-closures-and-ffi-callbacks/

As you pointed out elsewhere, try/catch is already slow so this is probably an unnecessary micro optimization. Especially since it's most likely best practice to make `try_catch` closures as small as possible since handling non-`Throw` error types would be awkward.

dherman: I believe the point of unboxed closures is that `Fn`/`FnMut`/`FnOnce` closures are not guaranteed to be boxed already, so they might be an inline struct of arbitrary size. Which implies that you can't assume they're a pointer.

crates/neon-runtime/src/napi/error.rs
```diff
@@ -9,3 +16,25 @@ pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimple
 pub unsafe extern "C" fn new_range_error(_out: &mut Local, _msg: Local) { unimplemented!() }
 
 pub unsafe extern "C" fn throw_error_from_utf8(_msg: *const u8, _len: i32) { unimplemented!() }
+
+pub unsafe extern "C" fn is_throwing(env: Env) -> bool {
```
kjvalencik, line 20: FYI, this doesn't need to be `extern "C"`. We have this in several places. I created an issue to clean it up.

https://github.com/neon-bindings/neon/issues/548

crates/neon-runtime/src/napi/error.rs
```diff
@@ -9,3 +16,25 @@ pub unsafe extern "C" fn new_type_error(_out: &mut Local, _msg: Local) { unimple
...
+    let mut b: MaybeUninit<bool> = MaybeUninit::zeroed();
```
kjvalencik, line 21: FYI, Since `bool` has a trivial uninitialized state, this could also be `let mut b = false;`

src/context/internal.rs
```diff
@@ -116,6 +129,96 @@ pub trait ContextInternal<'a>: Sized {
...
+    #[cfg(feature = "legacy-runtime")]
+    fn try_catch_internal<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
+        where F: UnwindSafe + FnOnce(&mut Self) -> JsResult<'b, JsValue>
+    {
+        // A closure does not have a guaranteed layout, so we need to box it in order to pass
+        // a pointer to it across the boundary into C++.
+        let rust_thunk = Box::into_raw(Box::new(f));
+
+        let mut local: MaybeUninit<raw::Local> = MaybeUninit::zeroed();
+        let mut unwind_value: MaybeUninit<*mut c_void> = MaybeUninit::zeroed();
```
kjvalencik, line 142: I think `MaybeUninit::uninit()` may be preferred.

> Incorrect usage of this function: initializing a struct with zero, where some fields cannot hold 0 as a valid value.

https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#zeroed.v

src/context/internal.rs
```diff
@@ -116,6 +129,96 @@ pub trait ContextInternal<'a>: Sized {
...
+            TryCatchControl::Threw => {
+                let local = unsafe { local.assume_init() };
+                Err(JsValue::new_internal(local))
+            }
+        }
+    }
+
+    #[cfg(feature = "napi-runtime")]
+    fn try_catch_internal<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
+        where F: UnwindSafe + FnOnce(&mut Self) -> JsResult<'b, JsValue>
```
kjvalencik, line 172: I don't think `UnwindSafe` is necessary on this version.

src/context/internal.rs
```diff
@@ -116,6 +129,96 @@ pub trait ContextInternal<'a>: Sized {
...
+    {
+        let result = f(self);
+        let mut local: MaybeUninit<raw::Local> = MaybeUninit::zeroed();
+        unsafe {
+            if neon_runtime::error::catch_error(self.env().to_raw(), local.as_mut_ptr()) {
+                Err(JsValue::new_internal(local.assume_init()))
+            } else if let Ok(result) = result {
+                Ok(result)
+            } else {
+                Ok(JsUndefined::new_internal(self.env()).upcast())
```
kjvalencik, line 182: Shouldn't this be `Err`?

src/context/mod.rs
```diff
@@ -229,6 +229,13 @@ pub trait Context<'a>: ContextInternal<'a> {
         result
     }
 
+    #[cfg(feature = "try-catch-api")]
+    fn try_catch<'b: 'a, F>(&mut self, f: F) -> Result<Handle<'a, JsValue>, Handle<'a, JsValue>>
```
kjvalencik, line 233: The current RFC defines the `Ok` value as a generic `T: Value`. This is feature flagged, so this is pretty minor.

src/context/mod.rs
```diff
@@ -229,6 +229,13 @@ pub trait Context<'a>: ContextInternal<'a> {
...
+        where F: UnwindSafe + FnOnce(&mut Self) -> JsResult<'b, JsValue>
```
kjvalencik, line 234: For the final non-feature flagged version, I think we should have separate legacy and N-API function signatures so that we don't need to unnecessarily require `UnwindSafe` on N-API. Removing the trait bound would be a breaking change.

src/context/internal.rs
dherman: Check my logic: `napi::check_and_clear_last_exception()` only produces an exception if the VM was in the throwing state. So in this branch, we aren't in a throwing state, but the Neon user returned `Err(Throw)`—in other words, this is the case where the Neon user incorrectly used the `NeonResult` protocol.

My reasoning was that `cx.try_catch()` should correctly indicate that nothing was caught even if the Neon user was incorrectly using the protocol.

As I thought about how dirty I felt for silently swallowing an error and replacing it with `undefined`, it occurred to me just now that maybe a better route is to panic.

kjvalencik: In order to support a generic `T: Value` on this, this either needs to be an `Err` or a `panic!`. Panic makes more sense to me since it's not the intended use.

src/context/mod.rs
dherman: Yeah, as I was hacking I had myself convinced that didn't make sense and the RFC was wrong. But now that I think about it, I guess it's just strictly more expressive, since choosing a given `T` gives you an implicit downcast, and if you want the dynamic version you can always choose `JsValue` as your `T`.

src/context/mod.rs
dherman: Good point! It's maybe not even a breaking change, since anything that is `UnwindSafe` will still satisfy the relaxed restriction.

crates/neon-runtime/src/napi/error.rs
dherman: I think it's better to use `MaybeUninit` to document that we are expecting the N-API call to do the initializing.

src/context/internal.rs
dherman: The docs are a little misleading here: it's not _incorrect_ usage, it's just not a valid initialization to use zero bits if that's not a valid representation of `T`. The undefined behavior only kicks in if you screw up the `.assume_init()` assertion. But it's still strictly safer in practice to always use `::zeroed()`, because at least your undefined behavior was based on a zeroed-out memory slot rather than whatever bits happened to be in that slot from before.

TL;DR A good defense-in-depth policy is always to use `MaybeUninit::zeroed()` when using the FFI to initialize a slot, even though it's still your responsibility to ensure `.assume_init()` is being used correctly.

src/context/mod.rs
dherman: Oh! I just figured out why it's not clear how to make this generic in `T: Value`. It means we have to do a downcast, which can throw a `TypeError`. But we're trying to catch exceptions in this API. And we don't even return a `NeonResult`. I think I'm going to leave the API as-is for now, since it's feature-flagged, and we can discuss the finer points of this in the RFC.

dherman: Follow-up to the follow-up… @kjvalencik pointed out on Slack that the way to make it generic is for the type signature to enforce that the callback returns a `T` as well. Once I understood this it was pretty easy to implement!

src/context/mod.rs
```diff
@@ -226,6 +226,22 @@ pub trait Context<'a>: ContextInternal<'a> {
         result
     }
 
+    #[cfg(all(feature = "try-catch-api", feature = "napi-runtime"))]
```
kjvalencik, line 229: Note: For the final documented version, it would be beneficial to include the behavior that this method will `panic` if a user manually constructs an `Err(Throw)`.

dherman: Yeah, when we un-feature-gate it, it'll need API docs. In the meantime I'll make sure that gets added to the RFC.

amilajack: I'll try to add an `Exceptions` section to the guides docs as well

PR #545, opened 2020-07-10
n-api: ArrayBuffers and Buffers

goto-bus-stop:
Implements the basic operations to create, read, and write ArrayBuffers and Node.js Buffers.

To pass the `napi_env` to the N-API functions, this threads the `Env` value through the `Lock` instance. Lock already guarantees that the lifetimes are OK, but this did make me wonder—should we add a lifetime parameter to `Env` too, bounded by the lifetime of `Context`? `Env` is not supposed to be stored, and end users of Neon don't have access to it so they can't do that, but it is still possible for us to make a mistake (like maybe some structure with an Env in it gets passed into an asynchronous task at some point).

### DIFF
--- a/crates/neon-runtime/src/napi/buffer.rs
+++ b/crates/neon-runtime/src/napi/buffer.rs
@@ -1,9 +1,21 @@
 use raw::{Env, Local};
 use std::os::raw::c_void;
+use std::ptr::null_mut;
 
 use nodejs_sys as napi;
 
-pub unsafe extern "C" fn new(_out: &mut Local, _size: u32) -> bool { unimplemented!() }
+pub unsafe extern "C" fn new(env: Env, out: &mut Local, size: u32) -> bool {
+    let status = napi::napi_create_buffer(env, size as usize, null_mut(), out as *mut _);
+
+    if status == napi::napi_status::napi_ok {
+        let mut bytes = null_mut();
+        let size = data(env, &mut bytes, *out);
+        std::ptr::write_bytes(bytes, 0, size);
+        true
+    } else {
+        false
+    }
+}
 
 pub unsafe extern "C" fn uninitialized(_out: &mut Local, _size: u32) -> bool { unimplemented!() }
 
--- a/crates/neon-runtime/src/napi/buffer.rs
+++ b/crates/neon-runtime/src/napi/buffer.rs
@@ -5,12 +5,12 @@ use std::ptr::null_mut;
 use nodejs_sys as napi;
 
 pub unsafe extern "C" fn new(env: Env, out: &mut Local, size: u32) -> bool {
-    let status = napi::napi_create_buffer(env, size as usize, null_mut(), out as *mut _);
-
+    let mut bytes = null_mut();
+    let status = napi::napi_create_buffer(env, size as usize, &mut bytes as *mut _, out as *mut _);
     if status == napi::napi_status::napi_ok {
-        let mut bytes = null_mut();
-        let size = data(env, &mut bytes, *out);
-        std::ptr::write_bytes(bytes as *mut u8, 0, size);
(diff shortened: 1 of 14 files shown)


src/context/mod.rs
```diff
@@ -105,14 +105,16 @@ pub enum CallKind {
 /// Types of JS values that support the `Borrow` and `BorrowMut` traits can be inspected while the engine is locked by passing a reference to a `Lock` to their methods.
 pub struct Lock<'a> {
     pub(crate) ledger: RefCell<Ledger>,
+    pub(crate) env: Env,
```
kjvalencik, line 108: Could this hold a `Context<'a>` instead or do we rely on the `Env` aliasing?

crates/neon-runtime/src/napi/buffer.rs
```diff
@@ -1,8 +1,29 @@
-use raw::Local;
+use raw::{Env, Local};
 use std::os::raw::c_void;
+use std::ptr::null_mut;
 
-pub unsafe extern "C" fn new(_out: &mut Local, _size: u32) -> bool { unimplemented!() }
+use nodejs_sys as napi;
+
+pub unsafe extern "C" fn new(env: Env, out: &mut Local, size: u32) -> bool {
+    let status = napi::napi_create_buffer(env, size as usize, null_mut(), out as *mut _);
```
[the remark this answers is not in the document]
goto-bus-stop: Oh right, I misread the documentation and thought that parameter let you pass _in_ a backing array for the buffer, rather than receive one created by node/V8. will update, thanks!

src/context/mod.rs
goto-bus-stop: I tried that first and it was difficult because Lock, Ref, RefMut, and the methods on Borrow and BorrowMut then all become generic over the Context trait. But, I think it might still be worth exploring, since on second thought this is a situation where I'm using the lifetime-less Env to work around the borrow checker. I actually specifically ran into compile errors where lifetimes of `Context::lock()` and `Context::borrow()` mismatched … Adding a lifetime parameter to Env would likely result in the same kind of problem. I didn't fully understand those errors though. I'll push that on a separate branch in a bit

crates/neon-runtime/src/napi/buffer.rs
```diff
@@ -1,8 +1,29 @@
...
+
+    if status == napi::napi_status::napi_ok {
+        let mut bytes = null_mut();
+        let size = data(env, &mut bytes, *out);
+        std::ptr::write_bytes(bytes as *mut u8, 0, size);
```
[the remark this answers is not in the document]
goto-bus-stop: yep, napi_create_arraybuffer gives you a zero initialized buffer but create_buffer does not

src/context/mod.rs
goto-bus-stop: this is where I got more or less: https://github.com/goto-bus-stop/neon/commit/b1cf3170ccc87f27e20740f632addb2359a888f9

src/context/mod.rs
```diff
@@ -122,9 +124,9 @@ pub trait Context<'a>: ContextInternal<'a> {
     /// Lock the JavaScript engine, returning an RAII guard that keeps the lock active as long as the guard is alive.
     /// 
     /// If this is not the currently active context (for example, if it was used to spawn a scoped context with `execute_scoped` or `compute_scoped`), this method will panic.
-    fn lock(&self) -> Lock {
+    fn lock<'c>(&'c self) -> Lock<'c> {
```
dherman, line 127: Any reason for expanding out the lifetimes here? I prefer using elision when it's available. It makes signatures more lightweight and intuitive.

goto-bus-stop: No, is a leftover from an attempt to pass the context reference to the Lock :)
```suggestion
    fn lock(&self) -> Lock<'_> {
```

Issue #551, opened 2020-07-17
Race in closing EventHandler causes abort


kjvalencik:
```
* thread #1: tid = 0, 0x00007fbd575a6438 libc.so.6`__GI_raise(sig=6) + 56 at raise.c:54, name = 'node', stop reason = signal SIGABRT
  * frame #0: 0x00007fbd575a6438 libc.so.6`__GI_raise(sig=6) + 56 at raise.c:54
    frame #1: 0x00007fbd575a803a libc.so.6`__GI_abort + 362 at abort.c:89
    frame #2: 0x000000000133361a node`uv_async_send + 218 at async.c:198
    frame #3: 0x00000000013335f0 node`uv_async_send(handle=<unavailable>) + 176 at async.c:73
    frame #4: 0x00007fbd54af18c7 index.node`neon::EventHandler::close(this=0x00000000035b1360) + 35 at neon_event.h:41
    frame #5: 0x00007fbd54aef806 index.node`::Neon_EventHandler_Delete(thread_safe_cb=0x00000000035b1360) + 32 at neon.cc:542
    frame #6: 0x00007fbd54ade25e index.node`_$LT$neon..event..EventHandlerInner$u20$as$u20$core..ops..drop..Drop$GT$::drop::h833bfb0e044efc73 + 14 at mod.rs:23
    frame #7: 0x00007fbd54a5b81f index.node`core::ptr::drop_in_place::h742ef0626d66849f + 15 at mod.rs:184
    frame #8: 0x00007fbd54a54702 index.node`alloc::sync::Arc$LT$T$GT$::drop_slow::h43359c7263e64074 + 34 at sync.rs:864
    frame #9: 0x00007fbd54a5befa index.node`_$LT$alloc..sync..Arc$LT$T$GT$$u20$as$u20$core..ops..drop..Drop$GT$::drop::h6ce1a32f714f0b7b + 90 at sync.rs:1386
    frame #10: 0x00007fbd54a5b52e index.node`core::ptr::drop_in_place::h341e1dbab8bd61e5 + 14 at mod.rs:184
    frame #11: 0x00007fbd54a5b9be index.node`core::ptr::drop_in_place::hb5cf8e0e4072296b + 14 at mod.rs:184
    frame #12: 0x00007fbd54a6bca7 index.node`_$LT$neon..context..CallContext$LT$neon..types..JsObject$GT$$u20$as$u20$async_await..Spawn$GT$::sp
```

Aborting here:
[...]

PR #552, opened 2020-07-17
EventHandler memory leak and race condition

kjvalencik:
Partial resolution for https://github.com/neon-bindings/neon/issues/551

### DIFF
--- a/src/event/mod.rs
+++ b/src/event/mod.rs
@@ -55,7 +55,16 @@ impl EventHandler {
     pub fn schedule_with<F>(&self, arg_cb: F)
         where F: FnOnce(&mut EventContext, Handle<JsValue>, Handle<JsFunction>),
               F: Send + 'static {
-        let callback = Box::into_raw(Box::new(arg_cb)) as *mut c_void;
+        // HACK: Work around for race condition in `close`. `EventHandler` cannot be
+        // dropped until all callbacks have executed.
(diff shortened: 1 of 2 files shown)


src/event/mod.rs
```diff
@@ -55,7 +55,16 @@ impl EventHandler {
...
         where F: FnOnce(&mut EventContext, Handle<JsValue>, Handle<JsFunction>),
               F: Send + 'static {
-        let callback = Box::into_raw(Box::new(arg_cb)) as *mut c_void;
+        // HACK: Work around for race condition in `close`. `EventHandler` cannot be
+        // dropped until all callbacks have executed.
+        // NOTE: This will still leak memory if the callback is never called
+        let cloned_cb = self.clone();
+        let cb = move |cx, this, cb| {
+            arg_cb(cx, this, cb);
+            std::mem::drop(cloned_cb);
```
kjvalencik, line 64: @dherman When you are holding onto a variable just to control when it's dropped, is it more idiomatic to manually call `std::mem::drop` or to assign to an un-used variable? `let _cloned_cb = cloned_cb;`

goto-bus-stop: I see `let _ = cloned_cb` used maybe a bit more than `drop(cloned_cb)` but both are common AFAIK

dherman (approved):
It seems OK to merge this even though it's not a complete fix. :shipit:

PR #555, opened 2020-07-20
chore: cli infra updates

amilajack:
- Removes unused deps
- Upgrades deps to latest semver
- Use TS ES interop
- Drops `ts-dict` for `Record`


cli/package.json
```diff
@@ -14,43 +14,32 @@
...
-    "mkdirp": "^0.5.1",
-    "quickly-copy-file": "^1.0.0",
-    "rimraf": "^2.6.1",
-    "rsvp": "^4.6.1",
-    "semver": "^5.1.0",
-    "toml": "^2.3.0",
+    "rimraf": "^3.0.2",
+    "semver": "^7.3.2",
+    "toml": "^3.0.0",
+    "ts-dict": "^0.1.1",
```
amilajack, line 28: `ts-dict` can't be dropped because `ts-typed-json` depends on it

dherman: OK, I published 0.3.0 of `ts-typed-json`, which doesn't use `ts-dict` anymore, it uses `Record` instead. Feel free to try using that instead!

cli/src/artifacts.ts
```diff
@@ -58,15 +62,15 @@ export default class Artifacts {
     if (!JSON.isObject(json)) {
       throw new TypeError("expected object, found " + (json === null ? "null" : typeof json));
     }
-    let active = json.active;
+    let { active } = json as ArtifactsOutput;
     if (typeof active !== 'string' && active !== null) {
       throw new TypeError("json.active is not a string or null");
     }
-    let jsonTargets = json.targets;
+    let { targets: jsonTargets } = json as any;
```
dherman, line 69: Again, would `unknown` work here?

I'm a big fan of destructuring generally, but when we have to add the extra cast, I'm not sure how much this refactor to use destructuring is really buying us. The cast feels kind of noisy and irrelevant to the reader. Sam with the `let { active } = ...` line above.

cli/src/build-settings.ts
```diff
@@ -66,11 +65,11 @@ export default class BuildSettings {
     if (!JSON.isObject(value)) {
       throw new TypeError("value must be an object");
     }
-    let { rustc, env, nodeVersion } = value;
+    let { rustc, env, nodeVersion } = value as any;
```
dherman, line 68: Again should we use `unknown`?

cli/src/artifacts.ts
```diff
@@ -58,15 +62,15 @@ export default class Artifacts {
     if (!JSON.isObject(json)) {
       throw new TypeError("expected object, found " + (json === null ? "null" : typeof json));
```
dherman, line 63: I've published a 0.3.1 update of `ts-typed-json` with customizable type error messages for type casts, so in places like this we could do:

```ts
json = JSON.asObject(json, 'json');
```

amilajack:
Failing because `ts-typed-json` uses `fs/promises`, which is not supported in node 10

PR #559, opened 2020-07-24
napi tidbits: finish `neon_runtime::napi::array`

goto-bus-stop:
This PR:
- ports the array tests from NAN
- implements `neon_runtime::napi::array::new`, so Rust can create arrays
- implements `neon_runtime::napi::tag::is_array`, so Rust can downcast to arrays, eg. to receive arrays as parameters
- adds a test for the downcasting-to-array case which was not tested by `test/dynamic` yet

### DIFF
--- a/crates/neon-runtime/src/napi/tag.rs
+++ b/crates/neon-runtime/src/napi/tag.rs
@@ -5,11 +5,8 @@ use nodejs_sys as napi;
 /// Return true if an `napi_value` `val` has the expected value type.
 unsafe fn is_type(env: Env, val: Local, expect: napi::napi_valuetype) -> bool {
     let mut actual = napi::napi_valuetype::napi_undefined;
-    if napi::napi_typeof(env, val, &mut actual as *mut _) == napi::napi_status::napi_ok {
-        actual == expect
-    } else {
-        false
-    }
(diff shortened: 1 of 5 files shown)


crates/neon-runtime/src/napi/tag.rs
```diff
@@ -35,7 +35,14 @@ pub unsafe extern "C" fn is_object(env: Env, val: Local) -> bool {
     is_type(env, val, napi::napi_valuetype::napi_object)
 }
 
-pub unsafe extern "C" fn is_array(_env: Env, _val: Local) -> bool { unimplemented!() }
+pub unsafe extern "C" fn is_array(env: Env, val: Local) -> bool {
+    let mut result = false;
+    if napi::napi_is_array(env, val, &mut result as *mut _) == napi::napi_status::napi_ok {
+        result
+    } else {
+        false
```
kjvalencik, line 43: Should this be asserting instead? It seems like it could be confusing if we say an object isn't an array when it is because the method failed for some other reason (e.g., throwing).

goto-bus-stop: think I agree — the other methods in this file do it this way but they should prob also panic.

PR #575, opened 2020-07-31
napi tidbits: implement remaining tag checks

goto-bus-stop:
Implements the remaining `is_*` runtime functions using `napi_typeof`.
`napi_typeof` uses the V8 `->IsObject()` family of functions, just like neon's NAN backend  does.

This does introduce one minor difference: `neon_runtime::nan::tag::is_object()` matches External values, while `neon_runtime::napi::tag::is_object()` does not. I don't think we currently have any ways that External values can be accessed directly, though, so it should not matter. And if and when we do, we should probably have `is_external()` rather than treating them as objects?

### DIFF
--- a/test/napi/lib/types.js
+++ b/test/napi/lib/types.js
@@ -1,0 +1,78 @@
+var addon = require('../native');
+var assert = require('chai').assert;
+
+describe('type checks', function() {
+  it('is_array', function () {
+    assert(addon.is_array([]));
+    assert(addon.is_array(new Array()));
+    assert(!addon.is_array(null));
+    assert(!addon.is_array(1));
+    assert(!addon.is_array({ 0: 'a', 1: 'b', length: 2 }));
+  });
+
+  it('is_array_buffer', function () {
+    assert(addon.is_array_buffer(new ArrayBuffer(0)));
+    assert(!addon.is_array_buffer(new DataView(new ArrayBuffer(0))));
+    assert(!addon.is_array_buffer(new Uint8Array(1024)));
+    assert(!addon.is_array_buffer(Buffer.alloc(64)));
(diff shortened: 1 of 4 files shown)


test/napi/lib/types.js
```diff
@@ -0,0 +1,78 @@
...
+  it('is_boolean', function () {
+    assert(addon.is_boolean(true));
+    assert(addon.is_boolean(false));
+    assert(!addon.is_boolean(new Boolean(true)));
+    assert(!addon.is_boolean(new Boolean(false)));
+  });
+
+  it('is_buffer', function () {
+    assert(addon.is_buffer(Buffer.alloc(64)));
+    assert(addon.is_buffer(new Uint8Array(64)));
```
goto-bus-stop, line 31: I'm not sure about this. `Buffer.isBuffer()` really checks for Buffer instances, but `napi_is_buffer` will treat any ArrayBuffer view as a buffer. It is fine for all the APIs that `JsBuffer` uses though.

Perhaps my understanding was wrong, and `JsBuffer` really maps to Uint8Array rather than to `require('buffer').Buffer`?

goto-bus-stop: There's a possible weirdness with this where `JsBuffer::new()` will always give you a `require('buffer').Buffer`, while downcasting a `JsValue` to `JsBuffer` may also give you a `Uint8Array` if that is what the user passed in. From the Neon side, there is no difference between the two, but for JavaScript users interacting with a Neon module, the difference might be significant.

This same possible-weirdness also exists for the NAN runtime, so it's not a regression.

goto-bus-stop: > This same possible-weirdness also exists for the NAN runtime, so it's not a regression.

Decided to keep it as is for now for this reason.

PR #577, opened 2020-08-02
README badges

dherman:
Replace Travis and Appveyor badges with Actions badges.

**Screenshot:**

image


kjvalencik:
Why is npm showing twice?

PR #578, opened 2020-08-02
Disable duplicate CI jobs for internal PRs

dherman:
Disable the tests for `pull_request` events created by our own internal PRs, since they already get triggered for the push to the branch, resulting in duplicate CI jobs. See:

https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/5

**Before:**

image

**After:**

image

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,9 +10,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    # Disable this on our own internal PRs since it'll be run on the push to the branch.
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-
     steps:
     - uses: actions/checkout@v2
     - name: Use Rust ${{ matrix.rust-toolchain }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,6 +10,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    # Disable this on our own internal PRs since it'll be run on the push to the branch.
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


kjvalencik:
Do they actually run twice or just display twice? I thought they were deduplicated on commit hash.

What do you think if only running CI on PR instead of push? As a maintainer, if we want CI to run on our branch, we can open a draft PR.

kjvalencik:
Looking at the screenshot, they have different times so they must be duplicates.

What do you think of only running on PRs?

kjvalencik:
@dherman What do you think of limiting builds to pushes on master? As maintainers, if we want CI to run on our branch, we can open a draft PR.

dherman:
@kjvalencik Oh, I missed the branch filtering suggestion from that github.community thread. Is that what you mean? Let me try that and see how it works out.

dherman:
@kjvalencik OK, yeah, that's better. Thanks, I'd missed that solution. WDYT?

.github/workflows/docs.yml
```diff
@@ -10,6 +10,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    # Disable this on our own internal PRs since it'll be run on the push to the branch.
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
kjvalencik, line 14: I think this can be safely removed because this workflow only runs on pushes and not pull requests.

.github/workflows/linux.yml
```diff
@@ -1,6 +1,13 @@
 name: Test on Linux
 
-on: [push, pull_request]
+on:
+  push:
+    # Disable this on our own internal PRs.
```
kjvalencik, line 5: This comment doesn't seem correct. This will still run on our own PRs.

```suggestion
    # Prevent duplicate workflows on PRs from upstream branches
```

PR #585, opened 2020-08-08
Implement FunctionContext::kind() for N-API

dherman:
Implements `FunctionContext::kind()` for the N-API port.

A thought occurred to me when working on this: N-API's `napi_get_new_target()` is a better and more general API than `cx.kind()`. I think `cx.kind()` is harmless and we can keep it around forever, but in the future I think we should add a `FunctionContext::new_target()` API for just the N-API backend. The signature would look something like:

```rust
impl<'a, T: This> CallContext<'a, T> {
    pub fn new_target(&mut self) -> Option<Handle<'a, JsValue>>;
}
```

And then `kind()` wouldn't be much different from asking if `new_target()` returned a `Some` or a `None`.

### DIFF
--- a/crates/neon-runtime/src/napi/call.rs
+++ b/crates/neon-runtime/src/napi/call.rs
@@ -1,3 +1,4 @@
+use std::mem::MaybeUninit;
 use std::os::raw::c_void;
 use std::ptr::null_mut;
 use raw::{FunctionCallbackInfo, Env, Local};
@@ -28,7 +29,17 @@ pub unsafe extern "C" fn get_isolate(_info: FunctionCallbackInfo) -> Env { unimp
 // FIXME: Remove. This will never be implemented
 pub unsafe extern "C" fn current_isolate() -> Env { panic!("current_isolate won't be implemented in n-api") }
 
-pub unsafe extern "C" fn is_construct(_info: FunctionCallbackInfo) -> bool { unimplemented!() }
+pub unsafe extern "C" fn is_construct(env: Env, info: FunctionCallbackInfo) -> bool {
+    let mut target: MaybeUninit<napi::napi_value> = MaybeUninit::zeroed();
+    let status = napi::napi_get_new_target(
+        env,
+        info,
+        target.as_mut_ptr()
+    );
+    assert_eq!(status, napi::napi_status::napi_ok);
+    let target: napi::napi_value = target.assume_init();
+    !target.is_null()
+}
 
 pub unsafe extern "C" fn this(env: Env, info: FunctionCallbackInfo, out: &mut Local) {
     let status = napi::napi_get_cb_info(
--- a/crates/neon-runtime/src/napi/call.rs
+++ b/crates/neon-runtime/src/napi/call.rs
@@ -30,14 +30,22 @@ pub unsafe extern "C" fn get_isolate(_info: FunctionCallbackInfo) -> Env { unimp
 pub unsafe extern "C" fn current_isolate() -> Env { panic!("current_isolate won't be implemented in n-api") }
 
 pub unsafe extern "C" fn is_construct(env: Env, info: FunctionCallbackInfo) -> bool {
-    let mut target: MaybeUninit<napi::napi_value> = MaybeUninit::zeroed();
+    let mut target: MaybeUninit<Local> = MaybeUninit::zeroed();
+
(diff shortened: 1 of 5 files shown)


crates/neon-runtime/src/napi/call.rs
```diff
@@ -28,7 +29,19 @@ pub unsafe extern "C" fn get_isolate(_info: FunctionCallbackInfo) -> Env { unimp
 // FIXME: Remove. This will never be implemented
 pub unsafe extern "C" fn current_isolate() -> Env { panic!("current_isolate won't be implemented in n-api") }
 
-pub unsafe extern "C" fn is_construct(_info: FunctionCallbackInfo) -> bool { unimplemented!() }
+pub unsafe extern "C" fn is_construct(env: Env, info: FunctionCallbackInfo) -> bool {
+    let mut target: MaybeUninit<napi::napi_value> = MaybeUninit::zeroed();
+    println!("CRASHING HERE");
+    let status = napi::napi_get_new_target(
```
dherman, line 35: @kjvalencik @goto-bus-stop Can you see anything obvious here I might be doing wrong? For some reason, `cargo test napi -- --nocapture` prints `CRASHING HERE` and then silently dies.

crates/neon-runtime/src/napi/call.rs
```diff
@@ -28,7 +29,17 @@ pub unsafe extern "C" fn get_isolate(_info: FunctionCallbackInfo) -> Env { unimp
...
-pub unsafe extern "C" fn is_construct(_info: FunctionCallbackInfo) -> bool { unimplemented!() }
+pub unsafe extern "C" fn is_construct(env: Env, info: FunctionCallbackInfo) -> bool {
+    let mut target: MaybeUninit<napi::napi_value> = MaybeUninit::zeroed();
+    let status = napi::napi_get_new_target(
+        env,
+        info,
+        target.as_mut_ptr()
+    );
+    assert_eq!(status, napi::napi_status::napi_ok);
+    let target: napi::napi_value = target.assume_init();
```
kjvalencik, line 40: I think this could use a comment. Maybe it's just me, but I had to think a LOT about whether it's valid to call `assume_init` here when the value could be `null`.

It _is_ valid and correct because `T` is a raw pointer (not a ref) and it's valid for a raw pointer to be `null`.

src/context/mod.rs
```diff
@@ -55,6 +56,15 @@ impl CallbackInfo<'_> {
         }
     }
 
+    #[cfg(feature = "napi-runtime")]
+    fn kind<'b, C: Context<'b>>(&self, cx: &C) -> CallKind {
```
kjvalencik, line 60: We should plan some time to discuss when references to `Context` should be mutable and when they shouldn't. I.e., are there places we are relying on the exclusive reference to ensure correctness?

Some n-api, but not all, will mutate the `Env` (e.g. changing state, throwing), but maybe those should be considered interior mutability and we should still consider that `&Env`?

This code relies on it being a shared reference because `info` is borrowing `self` while also passing it to `kind`.

This comment is kind of long and only partially relevant. I'll put more detail notes in an issue.

crates/neon-runtime/src/napi/call.rs
```diff
@@ -28,7 +29,17 @@ pub unsafe extern "C" fn get_isolate(_info: FunctionCallbackInfo) -> Env { unimp
 // FIXME: Remove. This will never be implemented
 pub unsafe extern "C" fn current_isolate() -> Env { panic!("current_isolate won't be implemented in n-api") }
 
-pub unsafe extern "C" fn is_construct(_info: FunctionCallbackInfo) -> bool { unimplemented!() }
+pub unsafe extern "C" fn is_construct(env: Env, info: FunctionCallbackInfo) -> bool {
+    let mut target: MaybeUninit<napi::napi_value> = MaybeUninit::zeroed();
```
goto-bus-stop, line 33: We use the type alias `Local` in most neon-runtime functions instead of `napi_value`. When N-API is the only runtime that doesn't make much sense anymore, but for now perhaps it's best to use `Local` for consistency? 😄

dherman:
@kjvalencik @goto-bus-stop Addressed both of your suggestions from review—feel free to take another look.

kjvalencik:
@dherman Looks great! Thanks for the clarifying comment!

PR #599, opened 2020-09-19
Remove docs gh workflow

amilajack:
Now that docs are hosted on docs.rs, do we need this anymore?


kjvalencik (approved):

kjvalencik:
@amilajack, what do we need to do to remove the docs domain so that they don't get picked up Google? I wouldn't want users to stumble across outdated docs and not know it.

amilajack:
I've updated netlify to not host the `api.neon-bindings.com` subdomain.

PR #600, opened 2020-09-21
fixed incorrect unsafe closure usage in event handler

kjvalencik:


goto-bus-stop (approved):
LGTM based on the clarification in slack :+1:

PR #601, opened 2020-09-22
n-api: use `napi_get_all_property_names`

goto-bus-stop:
This makes the `JsObject::get_own_property_names` actually correct. `napi_get_all_property_names` was added in N-API version 6, so it [raises our support floor to Node.js v10.20.0](https://nodejs.org/api/n-api.html#n_api_n_api_version_matrix).

### DIFF
--- a/crates/neon-runtime/src/napi/object.rs
+++ b/crates/neon-runtime/src/napi/object.rs
@@ -2,9 +2,6 @@ use std::mem::MaybeUninit;
 
 use nodejs_sys as napi;
 
-use array;
-use convert;
-use tag;
 use raw::{Env, Local};
 
 /// Mutates the `out` argument to refer to a `napi_value` containing a newly created JavaScript Object.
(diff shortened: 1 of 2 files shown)


crates/neon-runtime/src/napi/object.rs
```diff
@@ -15,70 +12,20 @@ pub unsafe extern "C" fn new(out: &mut Local, env: Env) {
...
-        }
-
-        let mut dummy = false;
-        // If we can't convert assign to this array, something went wrong.
-        if !set_index(&mut dummy, env, fixed_names, fixed_len, property_name) {
-            return false;
-        }
-        fixed_len += 1;
-    }
+    *out = property_names.assume_init();
```
kjvalencik, line 28: I just noticed this. I'll write a ticket for some clean-up in `neon`. This should probably also be using a `MaybeUninit` on the calling side or be returning an `Option` instead fo a boolean.

kjvalencik (approved):

kjvalencik:
@goto-bus-stop Great clean-up! Much nicer! 🎉

crates/neon-runtime/src/napi/object.rs
goto-bus-stop: Yeah, this is a common pattern throughout neon-runtime. Since these runtime functions don't need to be FFI-safe anymore with N-API, we can safely use things like Options for this stuff, but it probably needs to wait until the NAN backend is removed.

PR #606, opened 2020-09-26
Fix crash in DowncastError Display impl, remove String member

sbillig:
In the Display impl for DowncastError, fmt() calls self.to_string(), which calls fmt(), which calls self.to_string(), which... causing the following to crash:

```rust
    let s = cx.string("hello");
    if let Err(e) = s.downcast::<JsNumber>() {
        eprintln!("downcast error: {}", e);
    }
```

This commit also removes the String member in DowncastError, eliminating the allocation overhead on creation (which is then paid when/if the error is thrown), and removes the Error::description() impl, which has been deprecated since rust 1.42.

### DIFF
--- a/src/handle/mod.rs
+++ b/src/handle/mod.rs
@@ -52,7 +52,6 @@ impl<'a, T: Managed + 'a> Handle<'a, T> {
 pub struct DowncastError<F: Value, T: Value> {
     phantom_from: PhantomData<F>,
     phantom_to: PhantomData<T>,
-    description: String
 }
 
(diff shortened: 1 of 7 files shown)


kjvalencik:
@sbillig This is great! Thanks for fixing this. Would to mind adding a test to `test/dynamic` and `test/napi` to prevent a regression in the future?

sbillig:
@kjvalencik I added a simple test.

kjvalencik (approved):

kjvalencik:
@sbillig Thanks for your contribution!

PR #607, opened 2020-09-27
Reduce JsFunction::call overhead (via smallvec)

sbillig:
This shaves 10-20% off of a JsFunction call, in the (presumably) common case of <= 8 args, and should be no worse for > 8 args. We're talking about tens of nanoseconds so not a huge win, of course, but it's a simple change (and smallvec is a very lightweight dependency), so why not?

before: (each line is 100k function calls, with 4 args)
```
100000 	 total: 51.223332ms 	 each: 512ns
100000 	 total: 49.608249ms 	 each: 496ns
100000 	 total: 47.576854ms 	 each: 475ns
100000 	 total: 46.925382ms 	 each: 469ns
100000 	 total: 48.606469ms 	 each: 486ns
100000 	 total: 47.447813ms 	 each: 474ns
100000 	 total: 47.039698ms 	 each: 470ns
100000 	 total: 48.588494ms 	 each: 485ns
100000 	 total: 47.96396ms 	 each: 479ns
100000 	 total: 48.267152ms 	 each: 482ns
run: 494.14ms
```

after:
```
100000 	 total: 40.613036ms 	 each: 406ns
100000 	 total: 44.00374ms 	 each: 440ns
100000 	 total: 40.510302ms 	 each: 405ns
100000 	 total: 39.676057ms 	 each: 396ns
100000 	 total: 39.9972ms 	 each: 399ns
100000 	 total: 39.505668ms 	 each: 395ns
100000 	 total: 40.986956ms 	 each: 409ns
100000 	 total: 40.879661ms 	 each: 408ns
100000 	 total: 40.424424ms 	 each: 404ns
100000 	 total: 38.291358ms 	 each: 382ns
run: 413.305ms
```

Changing JsFunction::call to take a `&mut [Handle<...>]` instead of an IntoIterator is a _hair_ faster, and makes it easy for the user to avoid allocating a Vec of args to pass into call(), but that changes the API, of course, and I don't know what kind of shenanigans `neon_runtime::fun::call` might get up to with the pointer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ semver = "0.9"
 cslice = "0.2"
 semver = "0.9.0"
 neon-runtime = { version = "=0.4.2", path = "crates/neon-runtime" }
+smallvec = "1.4.2"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3.9", features = ["minwindef", "libloaderapi", "ntdef"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,7 @@
 extern crate neon_runtime;
 extern crate cslice;
 extern crate semver;
+extern crate smallvec;
 
 #[cfg(test)]
 #[macro_use]
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -19,6 +19,7 @@ use object::{Object, This};
 use object::class::Callback;
 use handle::{Handle, Managed};
 use handle::internal::SuperType;
+use smallvec::SmallVec;
 use self::internal::{ValueInternal, FunctionCallback};
 use self::utf8::Utf8;
 
@@ -587,7 +588,7 @@ impl<CL: Object> JsFunction<CL> {
               A: Value + 'b,
               AS: IntoIterator<Item=Handle<'b, A>>
     {
-        let mut args = args.into_iter().collect::<Vec<_>>();
+        let mut args = args.into_iter().collect::<SmallVec::<[_; 8]>>();
         let (argc, argv) = unsafe { prepare_call(cx, &mut args) }?;
         let env = cx.env().to_raw();
         build(|out| {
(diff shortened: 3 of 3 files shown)


kjvalencik:
This seems like a solid improvement. We are currently driving towards completing the N-API. I would be open to a breaking change for the `napi-runtime` feature.

Since this will always need to be collected into a `Vec` (the n-api function requires an array pointer), I think it would be better to take that directly. I expect most users don't send an iterator. Taking a slice (`&mut [T]`) is probably best because a user could use a `Vec` or in most cases an array.

Given the small performance gain, I would prefer to discuss the API further before bringing in a new dependency. @dherman What do you think?

@sbillig One reason to bring in `smallvec` is that a similar improvement could  be applied to `fn argument`.

dherman:
I think we should be open to this idea, and certainly N-API is an opportunity for a breaking change. I'm always in favor of being thoughtful about taking on dependencies but `smallvec` is well-tested and well-maintained, so I'm not too worried about that. But it would also be good to figure out just how disruptive the cost of migration would be, what the API and implementation would look like for the N-API runtime, and what the performance difference is there. This should probably have an RFC so we can understand the API change.

dherman:
But I forgot to say first of all, thank you @sbillig for this idea and contribution! Function calls are so central to Neon that managing their overhead is important.

kjvalencik:
@dherman I created an issue and gave it the `v1.0` label. https://github.com/neon-bindings/neon/issues/608

In my opinion, it would be rare for a user to construct arguments from an iterator. Most uses will have a fixed number of arguments. Switching to `AsRef<[T]>` will allow us to avoid heap allocation in most places without introducing `smallvec` or copying with the cost of requiring users to manually call `.collect()` if they *are* using an iterator.

dherman:
@kjvalencik That's awesome, thanks for writing it up! I misunderstood how similar this would be to the current API; I don't think it needs an RFC. And `AsRef[T]` just seems like a better design than `IntoIterator<Item=T>`, don't you think?

sbillig:
I agree that AsRef<[T]> would make for a better API, and would likely eliminate a vec allocation on the user's side, perhaps doubling the benefit shown here.

Feel free to close this PR if you'd prefer. I'm about to open another one replacing the vec in the napi-runtime CallContext with a smallvec (as suggested by @kjvalencik (that's what you meant, right?)), which similarly improves calls from js to functions defined in rust.

VulgarnyKarlson:
any updates of this?

PR #609, opened 2020-09-28
Clean-up: Lots of lint fixes

kjvalencik:
This fixes most clippy warnings. A couple would be breaking changes to public APIs and are ignored.

There are also lots of breaking changes to `neon-runtime`, but we consider that internal.

The `missing_safety_docs` lint is not addressed because that will take a pretty long time to implement.

```
cargo clippy --no-default-features --features legacy-runtime -- -A clippy::missing_safety_doc
cargo clippy --no-default-features --features napi-experimental -- -A clippy::missing_safety_doc
```

After this lands, I would like to `cargo fmt` the entire codebase.

### DIFF
--- a/crates/neon-runtime/src/nan/scope.rs
+++ b/crates/neon-runtime/src/nan/scope.rs
@@ -3,8 +3,14 @@
 use crate::raw::{HandleScope, EscapableHandleScope, InheritedHandleScope, Isolate};
 
 pub trait Root {
+    /// # Safety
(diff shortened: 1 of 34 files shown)


goto-bus-stop (approved):
looks pretty solid, the `.cast()`s are very nice, `cx.is_empty()` feels a little weird but i guess it's fine!

dherman (approved):
Love it! 🚢

PR #610, opened 2020-09-28
Use smallvec for CallContext arguments (napi)

sbillig:
Each line is 1 million calls to a variadic 'sum' function, from js to rust.
before:
```
argc = 0: 196.982ms
argc = 1: 359.803ms
argc = 4: 480.051ms
argc = 8: 585.154ms
argc = 9: 632.392ms
```
after:
```
argc = 0: 213.023ms
argc = 1: 293.921ms
argc = 4: 396.136ms
argc = 8: 522.838ms
argc = 9: 663.989ms
```

This adds some overhead to the 0 and > 8 arg cases (probably from the bigger CallContext object, and the branch in smallvec to check whether the buffer has spilled to the heap), in favor of a greater reduction in overhead in the presumably more common 1-8 arg cases.

Test code:

```rust
use neon::prelude::*;

fn sum(mut cx: FunctionContext) -> JsResult<JsNumber> {
    let mut s = 0.0;
    for i in 0..cx.len() {
	let x = cx.argument::<JsNumber>(i)?.value(&mut cx);
	s += x;
    }
    Ok(cx.number(s))
}

register_module!(mut cx, {
    cx.export_function("sum", sum)
});
```

```js
var addon = require('../native');

const n = 1000000

let run = args => {
  let msg = "argc = " + args.length;
  console.time(msg)
  for (let i = 0; i < n; i++) {
    addon.sum(...args)
  }
  console.timeEnd(msg)
}

run([])
run([1])
run([1, 2, 3, 4])
run([1, 2, 3, 4, 5, 6, 7, 8])
run([1, 2, 3, 4, 5, 6, 7, 8, 9])
```

Couple questions:
The smallvec dependency isn't currently feature-gated in neon-runtime. Probably the cleanest way to do that is to introduce a napi feature there that enables nodejs-sys and smallvec, so the users of this library don't have to enable both to use napi. Thoughts? I figured I'd check your preference, and whether you'd be inclined to merge this or not before making that change.

Do the neon-runtime functions need to be `extern "C"`? I don't see a reason for it. `fn argv` produces a "not FFI-safe" warning, whether the return value is Vec or SmallVec.


sbillig:
Oh, another question: do y'all have any thoughts about adding these kinds of benchmarks to the repo? Presumably it'd be best to use a nodejs benchmark lib; any preferences or desire to set such a thing up? (My personal appetite for mucking about in js-land is quite limited, tbh, but I'd gladly add some benchmark code if the infrastructure was in place)

kjvalencik:
@sbillig Benchmarks would be great especially as we drive towards v1.0. We would likely only be interested in benchmarking the `napi-runtime` since `legacy-runtime` will be going away.

Ideally the benchmarks would be in Rust or at least executed by Rust since we mostly run `neon` as a Rust project. For example, we have Rust tests that execute `npm test`. I'm not sure how realistic that would be for testing FFI.

kjvalencik:
@sbillig RE: Do `neon_runtime` functions need to be `extern "C"`?

For the most part, "no." Most of the functions currently marked `extern "C"` don't need to be, including the one that would be invalid if it _were_ used in that context. I have a WIP project clean-up PR that includes removing the unnecessary ones.

https://github.com/neon-bindings/neon/pull/609

sbillig:
@kjvalencik This should be ready to merge. (I kinda forgot about this pr :)

Final speed comparison:
```
0.4.2:  1000000 	 total: 826.375519ms 	 each: 826ns
0.5.1:  1000000 	 total: 710.845505ms 	 each: 710ns
this:   1000000 	 total: 639.310304ms 	 each: 639ns
```
so the two smallvec changes add up to a pretty decent speedup. (Note that this pr only applies to the napi runtime, of course)

code:
```rust
use neon::prelude::*;
use std::time::Instant;
use arrayvec::ArrayVec;

const N: u32 = 1_000_000;

fn sum(mut cx: FunctionContext) -> JsResult<JsNumber> {
    let mut s = 0.0;
    for i in 0..cx.len() {
	let x = cx.argument::<JsNumber>(i)?.value(&mut cx);
	s += x;
    }
    Ok(cx.number(s))
}

fn run(mut cx: FunctionContext) -> JsResult<JsString> {
    let fun = JsFunction::new(&mut cx, sum)?;

    let start = Instant::now();
    for _ in 0..N {
	let null = cx.null();
	let args = ArrayVec::from([cx.number(1), cx.number(2), cx.number(3), cx.number(4)]);
	fun.call(&mut cx, null, args)?.downcast_or_throw::<JsNumber, _>(&mut cx)?;
	// let mut args = [cx.number(1), cx.number(2), cx.number(3), cx.number(4)];
	// fun.call_with_slice(&mut cx, null, &mut args)?.downcast_or_throw::<JsNumber, _>(&mut cx)?;
    }
    let duration = start.elapsed();

    Ok(cx.string(format!("{} \t total: {:?} \t each: {:?}", N, duration, duration / N)))
}

register_module!(mut cx, {
    cx.export_function("run", run)
});
```

PR #617, opened 2020-10-15
Support nodejs-mobile on Android and iOS

staltz:
Neon is really great, and I've been intending to use it on mobile, specifically through [Node.js Mobile](https://code.janeasystems.com/nodejs-mobile) (a fork of Node.js with platform-specific tweaks for iOS and Android).

I've been following [this issue at nodejs-mobile](https://github.com/JaneaSystems/nodejs-mobile/issues/193) where people have figured out how to make things work. My PR contains many ideas from that discussion, and is a follow-up to PR #411. 

I have confirmed that these changes allowed me to compile [ssb-neon-keys](https://github.com/staltz/ssb-neon-keys) to `armv7-linux-androideabi` and `aarch64-linux-android` through nodejs-mobile's tooling (which includes a fork of node-gyp: `nodejs-mobile-gyp`), and it correctly runs on Android devices (both armv7 and arm64).

This PR does not address iOS support. I'm working on that, and I think the changes required in Neon will look considerably different from these changes, so it's worth submitting different PRs to make it clear what changes enable what support.

### DIFF
--- a/cli/src/build-settings.ts
+++ b/cli/src/build-settings.ts
@@ -57,7 +57,11 @@ export default class BuildSettings {
       npm_config_disturl:           process.env.npm_config_disturl || null,
       npm_config_runtime:           process.env.npm_config_runtime || null,
       npm_config_build_from_source: process.env.npm_config_build_from_source || null,
-      npm_config_devdir:            process.env.npm_config_devdir || null
+      npm_config_devdir:            process.env.npm_config_devdir || null,
+      npm_config_node_engine:       process.env.npm_config_node_engine || null,
+      npm_config_nodedir:           process.env.npm_config_nodedir || null,
+      npm_config_node_gyp:          process.env.npm_config_node_gyp || null,
+      npm_config_platform:          process.env.npm_config_platform || null
     });
   }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,6 +107,7 @@ macro_rules! register_module {
         #[cfg_attr(target_os = "linux", link_section = ".ctors")]
         #[cfg_attr(target_os = "android", link_section = ".ctors")]
         #[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
+        #[cfg_attr(target_os = "ios", link_section = "__DATA,__mod_init_func")]
         #[cfg_attr(target_os = "windows", link_section = ".CRT$XCU")]
         #[used]
         pub static __LOAD_NEON_MODULE: extern "C" fn() = {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,6 +105,7 @@ macro_rules! register_module {
         // Mark this function as a global constructor (like C++).
         #[allow(improper_ctypes)]
         #[cfg_attr(target_os = "linux", link_section = ".ctors")]
+        #[cfg_attr(target_os = "android", link_section = ".ctors")]
         #[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
         #[cfg_attr(target_os = "windows", link_section = ".CRT$XCU")]
         #[used]
--- a/crates/neon-sys/build.rs
+++ b/crates/neon-sys/build.rs
@@ -94,7 +94,7 @@ mod build {
     //
     //     gyp verb architecture ia32
     fn parse_node_arch(node_gyp_output: &str) -> String {
-        let version_regex = Regex::new(r"gyp verb architecture (?P<arch>ia32|x64)").unwrap();
+        let version_regex = Regex::new(r"gyp verb architecture (?P<arch>ia32|x64|arm|arm64)").unwrap();
         let captures = version_regex.captures(&node_gyp_output).unwrap();
         String::from(&captures["arch"])
     }
--- a/test/cli/src/unit/build-settings.ts
+++ b/test/cli/src/unit/build-settings.ts
@@ -7,6 +7,10 @@ describe('build settings', () => {
     npm_config_arch: null,
     npm_config_target_arch: null,
     npm_config_disturl: null,
+    npm_config_node_engine: null,
+    npm_config_node_gyp: null,
+    npm_config_nodedir: null,
+    npm_config_platform: null,
     npm_config_runtime: null,
     npm_config_build_from_source: null,
     npm_config_devdir: null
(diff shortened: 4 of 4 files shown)


kjvalencik:
@staltz Thanks for this contribution! I'm excited that you were able to get this working. I have a couple of questions.

Could you briefly explain how these environment variables are being used? They look very similar to the ones created by `electron-build-env` which did not require this change. Is there a subtle bug that might be impacting Electron as well?

In the linked issue, I see you are adding some `RUSTC` args. These can be set in `build.rs`, do you think it's valuable putting a feature in neon to discover this and link it?

staltz:
Thanks for taking a look at this already!

> Could you briefly explain how these environment variables are being used?

I'm not the person who discovered the need for those env vars (I mostly just want to condense those comments into code into this PR), but I understand some of them:

- `npm_config_nodedir`: nodejs-mobile builds a fork of node, so we have to override the path where node is typically found, to point to nodejs-mobile
- `npm_config_node_gyp`: we have to override the default `node-gyp` to point to `nodejs-mobile-gyp` instead
- `npm_config_platform`: I *think* nodejs-mobile needs this env var to detect android or ios cross compilation
- `npm_config_node_engine`: not sure why, but I think this is pointing to the specific "version" of node inside nodejs-mobile

cc @jaimecbernardo who authors nodejs-mobile.

>  Is there a subtle bug that might be impacting Electron as well?

I'm not sure if there would appear bugs in Electron. I ran the tests and all of them passed, including one that flashed an Electron window quickly. I personally care about Neon supporting Electron so I find this important too.

> In the linked issue, I see you are adding some `RUSTC` args. These can be set in `build.rs`, do you think it's valuable putting a feature in neon to discover this and link it?

If possible, yes, this would be a good feature, but we would need some way of detecting (in `build.rs`) that nodejs-mobile is being used (maybe through one of these env vars that point to `nodejs-mobile-gyp`?) in order to point `rustc-link-search` to the `nodejs-mobile-react-native` directory. And we'd have to make sure it's `nodejs-mobile-react-native` and not `nodejs-mobile-cordova`, and there's the question of how to handle `nodejs-mobile-cordova` when that is being used in the project. Due to all of these variables, I thought about starting cutting the complexity and just get things working. I agree with doing it properly though.

kjvalencik:
I believe these environment variables are only used for forcing rebuilds when they change. I.e., they aren't directly part of the fix, but a nice benefit for switching between platforms.

While not strictly necessary, I do think they are good to have because I think this is a small bug that would impact any targets that aren't the current `node` (including `electron`, but others as well).

I agree that it's a good idea to hold off on that discovery. This might be a place for another `npm` CLI like `electron-build-env`. `electron-build-env` searches `node_modules` to find the `electron` version and a similar thing could be done for `nodejs-mobile`; it doesn't need to be part of electron.

Thanks for helping me understand!

cli/src/build-settings.ts
```diff
@@ -57,7 +57,11 @@ export default class BuildSettings {
       npm_config_disturl:           process.env.npm_config_disturl || null,
       npm_config_runtime:           process.env.npm_config_runtime || null,
       npm_config_build_from_source: process.env.npm_config_build_from_source || null,
-      npm_config_devdir:            process.env.npm_config_devdir || null
+      npm_config_devdir:            process.env.npm_config_devdir || null,
+      npm_config_node_engine:       process.env.npm_config_node_engine || null,
+      npm_config_nodedir:           process.env.npm_config_nodedir || null,
+      npm_config_node_gyp:          process.env.npm_config_node_gyp || null,
+      npm_config_platform:          process.env.npm_config_platform || null
```
kjvalencik, line 64: Adding these values is going to force rebuilds for users when they upgrade the CLI, but considering we recommend keeping the CLI and `neon` versions in sync, most users would need to rebuild anyway. 👍

src/lib.rs
```diff
@@ -105,6 +105,7 @@ macro_rules! register_module {
         // Mark this function as a global constructor (like C++).
         #[allow(improper_ctypes)]
         #[cfg_attr(target_os = "linux", link_section = ".ctors")]
+        #[cfg_attr(target_os = "android", link_section = ".ctors")]
```
kjvalencik, line 108: @dherman I'm not sure what other platforms this might impact. Would specifying `unix` match both `linux` and `android`? What about changing this to `all(not(windows), not(macos))`?

jaimecbernardo:
Hi, 
> I'm not the person who discovered the need for those env vars (I mostly just want to condense those comments into code into this PR), but I understand some of them:
> 
> * `npm_config_nodedir`: nodejs-mobile builds a fork of node, so we have to override the path where node is typically found, to point to nodejs-mobile
> * `npm_config_node_gyp`: we have to override the default `node-gyp` to point to `nodejs-mobile-gyp` instead
> * `npm_config_platform`: I _think_ nodejs-mobile needs this env var to detect android or ios cross compilation
> * `npm_config_node_engine`: not sure why, but I think this is pointing to the specific "version" of node inside nodejs-mobile

Sounds accurate to me. `npm_config_node_engine` was for using `chakracore` instead of `v8` on earlier versions of `nodejs-mobile` and I don't think it needs be used anymore, since `v8` is now used on both platforms.

There are a couple more here that might be helpful, if I remember correctly:
https://github.com/JaneaSystems/nodejs-mobile-react-native/blob/651425964f77a0520ba7c34c7cd6783b0ac1b6cf/android/build.gradle#L376-L387

`npm_config_arch`: Target CPU architecture.
`npm_config_format`: So that the `make` toolset is used when building on macOS.

src/lib.rs
dherman: TBH I don't really know where `.ctors` is "standardized" or how broad of a convention it is. It's an interesting question but in the interest of making progress, I'd say let's stick with the PR as written since it's the most conservative incremental improvement.

kjvalencik (approved):

src/lib.rs
staltz: I have another PR coming up to add iOS support and the only change in it would be one line of code here, to specify that the iOS case should look just like the macos (darwin) case, I guess because of LLVM.

Would you prefer that I update this PR with that, or should we do a separate PR for iOS?

dherman: I'm fine either way—I'd say do whatever's easiest for you. Thank you for these contributions!

staltz:
I updated this PR to add one commit to support iOS, it's a one-liner commit.

kjvalencik (approved):

PR #618, opened 2020-10-27
Fix CI after changing master branch name to main

kjvalencik:


crates/neon-sys/native/src/neon_class_metadata.h
```diff
@@ -20,7 +20,7 @@
 // Slots 0 and 1 appear to be reserved by Chrome, and slot 3 is reserved by Node.
 // That apparently leaves only slot 2 available for storing Neon metadata.
 //
-// See: https://github.com/nodejs/node/blob/master/src/env.h#L33-L39
+// See: https://github.com/nodejs/node/blob/main/src/env.h#L33-L39
```
goto-bus-stop, line 23: this one shouldnt be changed :D

```suggestion
// See: https://github.com/nodejs/node/blob/master/src/env.h#L33-L39
```

goto-bus-stop (approved):
oh boy, i grepped through everything so many times to make sure, but of course `rg` ignores dotfiles by default… thanks!

crates/neon-sys/native/src/neon_class_metadata.h
kjvalencik: 🤦

dherman (commented):
:shipit:

dherman (commented):
:shipit:

dherman:
oh lol I was wondering why I couldn't review it! it's already merged 😂

PR #620, opened 2020-10-28
Add redirect support for downloading node.lib

adumbidiot:
I added primitive redirect support for downloading node.lib in NAPI since Electron uses it. While its nice not pulling in any external Rust or Js deps in the build script, I think some time should be spent at some point making the download process more robust.

### DIFF
--- a/crates/neon-build/Cargo.toml
+++ b/crates/neon-build/Cargo.toml
@@ -13,3 +13,6 @@ cfg-if = "0.1.9"
 
 [build-dependencies]
 cfg-if = "0.1.9"
+
+[target.'cfg(windows)'.dependencies]
+ureq = { version = "1.3.0", default-features = false, features = ["native-tls"] }
--- a/crates/neon-build/src/lib.rs
+++ b/crates/neon-build/src/lib.rs
@@ -46,22 +46,41 @@ cfg_if! {
             let stdout = String::from_utf8(output.stdout).map_err(|error| {
                 Error::new(ErrorKind::InvalidData, error)
             })?;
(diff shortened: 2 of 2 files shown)


kjvalencik:
Thanks for the contribution and helping us to test electron. Using node is a neat idea, but it might be complicated enough that we should bring in a build dependency. For example, this doesn't handle 301s.

@goto-bus-stop what do think of using `ureq`?

adumbidiot:
Looking at the history, `ureq` used to be used until it got [replaced](https://github.com/neon-bindings/neon/commit/2574b2dbda25e129892ba97e13d8c20edbdf001b#diff-e0250d1b915873dae9cd4b5052903e0a62cb12df1e462c637f8f9354beabaf79).

kjvalencik:
@adumbidiot That commit was part of the same PR that also replaced it. It's never been used in `neon`.

goto-bus-stop:
> I think some time should be spent at some point making the download process more robust.

I agree … initially we got rid of the dependency since I assumed we'd just be getting things from nodejs.org, but since that is not the case, the `ureq` dependency will be worth it. We might be able to do it just by reverting that commit you linked, if that still applies cleanly…

goto-bus-stop:
Reverted https://github.com/neon-bindings/neon/commit/2574b2dbda25e129892ba97e13d8c20edbdf001b on top of this branch, that should do the trick 🙏

kjvalencik (approved):

goto-bus-stop:
Thanks for bringing this up @adumbidiot :)

PR #622, opened 2020-11-05
Threadsafe Handles

kjvalencik:
Implementation of https://github.com/neon-bindings/rfcs/pull/32


crates/neon-runtime/src/napi/bindings/types.rs
```diff
@@ -50,10 +69,20 @@ pub(crate) type Finalize = Option<
     ),
 >;
 
+#[cfg(feature = "napi-4")]
+pub type ThreadsafeFunctionCallJs = Option<
```
dherman, line 73: This name confused me. Maybe `ThreadsafeFunctionJsCallback`?

crates/neon-runtime/src/napi/tsfn.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Error returned when scheduling a threadsafe function with some data
+pub struct CallError<T> {
+    kind: napi::Status,
+    data: T,
+}
+
+impl<T> CallError<T> {
+    /// The specific error that occurred
+    pub fn kind(&self) -> napi::Status {
```
dherman, line 54: Do we want to expose any `napi::` APIs? It might be better to have our own Neon enum for this. (We can change this in followup work, though, since this isn't yet stabilized.)

src/task/event_queue.rs
```diff
@@ -0,0 +1,161 @@
+use neon_runtime::raw::Env;
```
dherman, line 1: Also suggested on the RFC PR: I'd advocate `neon::event` instead of `neon::task` for this.

dherman (approved):
Looks awesome. I made a few suggestions but I'll leave them all to your discretion.

crates/neon-runtime/src/napi/tsfn.rs
kjvalencik: It's not exposed in `neon` only in `neon-runtime`. I agree that we should have a separate enum, but it's not used until we add bounded queues.

crates/neon-runtime/src/napi/bindings/types.rs
kjvalencik: This is the name that N-API uses. I don't think we should change it unless we have a strong reason:
https://nodejs.org/api/n-api.html#n_api_napi_threadsafe_function_call_js

kjvalencik:
All possible combinations of added feature flags were tested to ensure the code compiles without unused warnings:

```sh
cargo check && cargo check --features event-handler-api && cargo check --no-default-features --features napi-1 && cargo check --no-default-features --features napi-4 && cargo check && cargo check --no-default-features --features napi-4,event-queue-api && cargo check --no-default-features --features napi-1,event-queue-api && cargo check --no-default-features --features napi-4,event-handler-api
````

File: src/event/event_handler.rs [hunk 4]
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct EventHandler(Arc<EventHandlerInner>);
+
+impl EventHandler {
+    #[cfg(feature = "napi-1")]
+    pub fn new<'a, C: Context<'a>, T: Value>(cx: &C, this: Handle<T>, callback: Handle<JsFunction>) -> Self {
+        unimplemented!()
```
dherman, line 34: Maybe include a panic message that explains that this experimental API is not supported in the N-API backend, with a suggestion for an alternative.

dherman (approved):
Looks good to me. One suggestion for a clearer failure message if someone tries to use the deprecated EventHandler API with N-API.

demurgos:
Thank you very much for merging this PR!
I tried it out last week and it was great: it allowed me to completely switch to the N-API runtime and simplify my code.

I hope it will be released soon :slightly_smiling_face:

kjvalencik:
@demurgos Yep! It's on `0.7.0` behind the `event-queue-api` feature flag. It requires at least `napi-4`.

```toml
neon = { version = "0.7", default-features = false, features = ["napi-4", "event-queue-api"] }
```

It's behind an additional feature flag because the API won't be considered stable until the [RFC](https://github.com/neon-bindings/rfcs/pull/32) is merged.

Here are the API docs: https://docs.rs/neon/0.7.0-napi.3/neon/event/struct.EventQueue.html

PR #627, opened 2020-11-09
Build Improvements

kjvalencik:
Currently, the cache will invalidate frequently on Windows when switching between `cargo check` and `electron-build-env neon build`. This is due to two issues:

* `neon-cli` *always* sets a `--target` flag on `cargo` under Windows. This is treated as different from no target, even when the default matches the explicit value. This causes `cargo check` and `neon build` not to share a cache on Windows.
* In order to prevent a module build for the user's node from being used for Electron, we set a `rerun-if-env-changed`. This requires users to `cargo check` under `electron-build-env`.

This diff reads a little nicer [ignoring whitespace](https://github.com/neon-bindings/neon/pull/627/files?diff=unified&w=1).

### DIFF
--- a/crates/neon-build/src/lib.rs
+++ b/crates/neon-build/src/lib.rs
@@ -22,14 +22,11 @@ cfg_if! {
             println!("cargo:rustc-link-search=native={}", &node_lib_path.display());
             println!("cargo:rustc-link-lib={}", &node_lib_file_path.file_stem().unwrap().to_str().unwrap());
 
-            // Link `win_delay_load_hook.obj` for windows electron
-            let node_runtime_env = "npm_config_runtime";
-            println!("cargo:rerun-if-env-changed={}", node_runtime_env);
-            if var(node_runtime_env).map(|s| s == "electron") == Ok(true) {
-                println!("cargo:rustc-cdylib-link-arg=win_delay_load_hook.obj");
-                println!("cargo:rustc-cdylib-link-arg=delayimp.lib");
-                println!("cargo:rustc-cdylib-link-arg=/DELAYLOAD:node.exe");
-            }
+            // Link `win_delay_load_hook.obj`
+            // Needed for electron, but okay for other environments
+            println!("cargo:rustc-cdylib-link-arg=win_delay_load_hook.obj");
+            println!("cargo:rustc-cdylib-link-arg=delayimp.lib");
+            println!("cargo:rustc-cdylib-link-arg=/DELAYLOAD:node.exe");
         }
     } else if #[cfg(windows)] {
         // ^ automatically not neon-sys
--- a/cli/src/target.ts
+++ b/cli/src/target.ts
@@ -31,22 +31,27 @@ export type TargetOptions = {
 export default class Target {
   readonly crate: Crate;
   readonly release: boolean;
-  readonly arch: string;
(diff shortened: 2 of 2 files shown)


crates/neon-build/src/lib.rs
```diff
@@ -22,14 +22,11 @@ cfg_if! {
...
-            // Link `win_delay_load_hook.obj` for windows electron
-            let node_runtime_env = "npm_config_runtime";
-            println!("cargo:rerun-if-env-changed={}", node_runtime_env);
-            if var(node_runtime_env).map(|s| s == "electron") == Ok(true) {
-                println!("cargo:rustc-cdylib-link-arg=win_delay_load_hook.obj");
-                println!("cargo:rustc-cdylib-link-arg=delayimp.lib");
-                println!("cargo:rustc-cdylib-link-arg=/DELAYLOAD:node.exe");
-            }
+            // Link `win_delay_load_hook.obj`
+            // Needed for electron, but okay for other environments
```
kjvalencik, line 26: This follows the pattern of `node-gyp` which always links this.

goto-bus-stop:
> In order to prevent a module build for the user's node from being used for Electron, we set a rerun-if-env-changed. This requires users to cargo check under electron-build-env. This should be opt-in.

I'm not sure if I fully grasp what this means. My understanding is that:

Previously, if you did
```
electron-build-env cargo build
cargo check
electron-build-env cargo build
```
it would have to redo work twice.

So with this change, you can do the above without it rebuilding. And if you do `cargo build` followed by `electron-build-env cargo build`, it would _not_ rebuild.

Does that sound right? 🤔

kjvalencik:
@goto-bus-stop Yes, that's correct.

goto-bus-stop:
So if you start with a `cargo build` first, you'll have to manually invalidate the cache when you start using `electron-build-env`?

Not sure if that is a big problem. I think conceptually it makes sense to me that the target runtime would invalidate the build even if it is inconvenient in this case. But I haven't done any Neon+Electron work so I don't have the awful experience with it yet :)

kjvalencik:
@goto-bus-stop As far as I know, the differences between Electron and normal node are all in the linking stage which doesn't get cached. `neon-cli` will force a rebuild with `artifacts.json` if other parts change.

I don't _think_ this will cause an issue for the user, but it's why I originally made it optional. I can revert my change and make it optional again.

I don't feel strongly _which_ way it gets fixed, but it's a very bad experience on Windows. Especially with `rust-analyzer` which defaults to doing a `cargo check` after every change.

goto-bus-stop:
should we just merge this as is? I forgot if we decided on anything in the call.

ab-cgh:
I got stung by this one at work today. Rebuilt an Electron project to test some UI changed and RA also picked up on the changes, of course. The next Electron built itself from scratch. +1 to this being merged.

kjvalencik:
@ab-cgh Unfortunately, I'm not clear on the action to take. This could _cause_ as many issues at it fixes since Rust shares a build cache across check and build. It also could introduce a regression for modules that switch between targets.

I'm not certain the best way to handle it. FWIW, N-API does not have any of these issues.

kjvalencik:
Closing since this only applies to legacy.

Issue #630, opened 2020-11-10
Feature request: loosen try_catch's type to return Result<_, Handle<'a, JsValue>>


jrose-signal:
`Context::try_catch`'s current signature is

```rust
fn try_catch<'b: 'a, T, F>(
    &mut self,
    f: F
) -> Result<Handle<'a, T>, Handle<'a, JsValue>>
where
    T: Value,
    F: UnwindSafe + FnOnce(&mut Self) -> JsResult<'b, T>, 
```

However, nothing in the implementation (as far as I can tell) actually relies on the return value being a JS value. It could just as easily be

```rust
fn try_catch<'b: 'a, T, F>(
    &mut self,
    f: F
) -> Result<T, Handle<'a, JsValue>>
where
    F: UnwindSafe + FnOnce(&mut Self) -> NeonResult<T>, 
```

kjvalencik:
@jrose-signal I like this idea and it would be good for the N-API backend, but unfortunately it would be complicated for the legacy backend.

`v8::TryCatch` need to be allocated on the stack and this forces our Rust to execute from C++. The returned data would need to be heap allocated and `Send`.

Can you share more about your use case? In my opinion, `try_catch` blocks should be fairly tightly scoped. Is there something in Neon's API that returns a `T` but, can also throw that you would like to catch?

Thanks!

jrose-signal:
> `v8::TryCatch` need to be allocated on the stack and this forces Rust to execute from C++. The returned data would need to be heap allocated and `Send`.

I think it just has to be `Sized`: you've already got a MaybeInit here

https://github.com/neon-bindings/neon/blob/1f217f56cb2a1760621e76c1dc317526229de3b3/src/context/internal.rs#L137

that gets initialized from pure Rust code here

https://github.com/neon-bindings/neon/blob/1f217f56cb2a1760621e76c1dc317526229de3b3/src/context/internal.rs#L208

without C++ doing anything more than passing the pointer down the stack. That should be safe to treat as a mutable borrow, yeah?

> Can you share more about your use case? In my opinion, `try_catch` blocks should be fairly tightly scoped. Is there something in Neon's API that returns a `T` but, can also throw that you would like to catch?

[...]

PR #631, opened 2020-11-10
feat(neon): Update `try-catch` to match the RFC; return a `T: Sized` instead of `T: Value`

kjvalencik:
As requested by https://github.com/neon-bindings/neon/issues/630

* Removes unnecessary `UnwindSafe` bound
* Changes `T: Value` to `T: Sized`
* Updates a test to leverage a non-JavaScript `T` (`()`)

### DIFF
--- a/src/context/internal.rs
+++ b/src/context/internal.rs
@@ -127,8 +127,7 @@ pub trait ContextInternal<'a>: Sized {
 
     #[cfg(feature = "legacy-runtime")]
     fn try_catch_internal<'b: 'a, T, F>(&mut self, f: F) -> Result<T, Handle<'a, JsValue>>
-        where T: Sized,
-              F: FnOnce(&mut Self) -> NeonResult<T>,
+        where F: FnOnce(&mut Self) -> NeonResult<T>,
     {
         // A closure does not have a guaranteed layout, so we need to box it in order to pass
         // a pointer to it across the boundary into C++.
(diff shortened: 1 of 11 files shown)


src/context/internal.rs
```diff
@@ -205,7 +204,7 @@ extern "C" fn try_catch_glue<'a, 'b: 'a, C, T, F>(rust_thunk: *mut c_void,
     match catch_unwind(AssertUnwindSafe(|| f(cx))) {
         // No Rust panic, no JS exception.
         Ok(Ok(result)) => unsafe {
-            *returned = result.to_raw();
+            *(returned as *mut T) = result;
```
jrose-signal, line 207: This should be using `ptr::write`, since the existing value is garbage and should not be Dropped.

src/context/internal.rs
```diff
@@ -126,22 +126,24 @@ pub trait ContextInternal<'a>: Sized {
...
+        let mut ok: MaybeUninit<T> = MaybeUninit::zeroed();
+        let mut err: MaybeUninit<raw::Local> = MaybeUninit::zeroed();
         let mut unwind_value: MaybeUninit<*mut c_void> = MaybeUninit::zeroed();
 
         let ctrl = unsafe {
             neon_runtime::try_catch::with(try_catch_glue::<Self, T, F>,
                                           rust_thunk as *mut c_void,
                                           (self as *mut Self) as *mut c_void,
-                                          local.as_mut_ptr(),
+                                          ok.as_mut_ptr() as *mut c_void,
```
jrose-signal, line 145: I think if there were problems around Sized, this cast would fail, since Rust `as` casts only allow casting between pointers of the same size.

jrose-signal: Okay, I found the citation for you: https://doc.rust-lang.org/reference/type-layout.html#pointers-and-references-layout

> Pointers to sized types have the same size and alignment as `usize`.

kjvalencik: Awesome, thanks!

goto-bus-stop (commented):
LGTM overall

dherman (changes requested):
Since this changes behavior we should have a test for the new behavior, i.e. a result that produces a non-`Value` value.

Otherwise looks great!

dherman (approved):
:shipit:

Issue #471, opened 2019-12-15
Problem: Missing --feature conditional building support


calvinlauco:
### Description:

Trying to do a conditional compilation using `neon build --feature=mock`

### Expected behaviour

Build with `mock` feature

### Actual behaviour

```
UNKNOWN_OPTION: Unknown option: --feature
...
```

kjvalencik:
The flag is pluralized. `--features`

kjvalencik:
Wow! This is a pretty big gap since unlike most cargo options, features can't be enabled with environment variables.

PR #633, opened 2020-11-12
feat: Allow additional arguments to be passed to `cargo build`

kjvalencik:
Resolves https://github.com/neon-bindings/neon/issues/471

Instead of adding a special case to `features` and needing to keep up with `cargo`, all arguments after a bare `--` will be passed to cargo.

```sh
neon build -- --features awesome
```

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -29,6 +29,23 @@ function logIf(multiple: boolean, action: string, cwd: string, module: string) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


goto-bus-stop (approved):

dherman (approved):
LGTM! Might be good to follow up with CLI tests for this but if there are bugs I don't expect them to be showstoppers.

PR #636, opened 2020-11-19
feat(neon-macros): Add `neon::main` proc attr macro

kjvalencik:
## Summary

Adds a `#[neon::main]` proc macro.

## tl;dr

Use an attribute to initialize either a `legacy` or `napi` neon module. The `fn` doesn't even need to be `pub`!

```rust
use neon::prelude::*;

#[neon::main]
    Ok(())
}
```

### Migration

```diff
diff --git a/test/dynamic/native/src/lib.rs b/test/dynamic/native/src/lib.rs
index 3f2c937..409ab55 100644
--- a/test/dynamic/native/src/lib.rs
+++ b/test/dynamic/native/src/lib.rs
@@ -20,7 +20,8 @@ use js::classes::*;
 use js::tasks::*;
 use js::eventhandler::*;

-register_module!(mut cx, {
+#[neon::main]
+fn main(mut cx: ModuleContext) -> NeonResult<()> {
     cx.export_function("return_js_string", return_js_string)?;

     cx.export_function("return_js_number", return_js_number)?;
@@ -84,4 +85,4 @@ register_module!(mut cx, {
     cx.export_class::<JsPanickyConstructor>("PanickyConstructor")?;

     Ok(())
-});
+}
```

### `neon-macros`

Procedural macros must be in their own crate. A new crate, `neon-macros`, is created. The crate has a single feature flag, `napi` for toggling between `legacy` and `napi` implementations.

Proc macros must be at the root of a crate. To keep the structure clean, actual implementations are in `napi.rs` and `legacy.rs` modules that are conditionally imported to `lib.rs`. The `lib.rs` has thin glue code for delegating to the correct implementation.

### Re-export

Proc macros are re-exported from `neon` to keep users from needing to specify another crate.

### Feature Flags

Proc macros can have a negative impact on compile time. A new feature flag, `proc-macros` is added to enable the feature. It is default disabled since most legacy users will continue to use the normal macros.

It is planned for proc macros to be the recommended way to use N-API. Therefore, the `napi-runtime` feature flag _automatically_ enables the `proc-macros` feature flag.

## Open Questions

1. ~Should we keep the panic hook? I think it should be removed. In my opinion, it's more harmful than helpful; especially since it is global.~ The hook is a holdover from earlier implementations and is no longer necessary. It will be removed.
1. ~Naming. I like `neon::init` because it initializes and can be called multiple times (once per context). However, `neon::main` is another option.~ Consensus on `neon::main`.
1. ~Should we try to protect against using `#[neon::init]` multiple times? I think we shouldn't because it would be complicated and brittle for limited value.~ This can be a follow-up.

### DIFF
--- a/crates/neon-macros/src/lib.rs
+++ b/crates/neon-macros/src/lib.rs
@@ -1,0 +1,43 @@
+//! Procedural macros supporting [Neon](https://docs.rs/neon/latest/neon/)
+
+#[cfg(feature = "napi")]
+mod napi;
+#[cfg(feature = "napi")]
+use napi as macros;
+
+#[cfg(not(feature = "napi"))]
+mod legacy;
+#[cfg(not(feature = "napi"))]
+use legacy as macros;
+
+// Proc macro definitions must be in the root of the crate
+// Implementations are in the backend dependent module
+
+#[proc_macro_attribute]
+/// Marks a method as the main entrypoint for initialization in a Neon
+/// module. This attribute should only be used _once_ in a module and will
+/// be called each time the module is initialized in a context.
+///
+/// ```no_run
+/// # use neon::prelude::*;
+/// #[neon::main]
+/// fn my_module(mut cx: ModuleContext) -> NeonResult<()> {
+///     let version = cx.string("1.0.0");
+/// 
+///     cx.export_value("version", version)?;
+/// 
+///     Ok(())
+/// }
+/// ```
+///
+/// If multiple functions are marked with `#[neon::main]`, there may be a compile error:
+///
+/// ```sh
+/// error: symbol `napi_register_module_v1` is already defined
+/// ```
+pub fn main(
+    attr: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    macros::main(attr, item)
+}
--- a/crates/neon-macros/src/napi.rs
+++ b/crates/neon-macros/src/napi.rs
@@ -1,0 +1,34 @@
+pub(crate) fn main(
+    _attr: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let input = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let attrs = &input.attrs;
+    let vis = &input.vis;
+    let sig = &input.sig;
+    let block = &input.block;
+    let name = &sig.ident;
+
+    quote::quote!(
+        #(#attrs) *
+        #vis #sig {
+            #[no_mangle]
+            unsafe extern "C" fn napi_register_module_v1(
+                env: ::neon::macro_internal::runtime::nodejs_sys::napi_env,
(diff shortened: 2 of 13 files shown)


crates/neon-macros/src/napi.rs
```diff
@@ -0,0 +1,38 @@
...
+        #(#attrs) *
+        #vis #sig {
+            #[no_mangle]
+            unsafe extern "C" fn napi_register_module_v1(
+                env: ::neon::macro_internal::runtime::nodejs_sys::napi_env,
+                m: ::neon::macro_internal::runtime::nodejs_sys::napi_value,
+            ) -> ::neon::macro_internal::runtime::nodejs_sys::napi_value {
+                // Suppress the default Rust panic hook, which prints diagnostics to stderr.
+                #[cfg(not(feature = "default-panic-hook"))]
+                ::std::panic::set_hook(::std::boxed::Box::new(|_| { }));
```
kjvalencik, line 23: @dherman I would like to discuss this. In my opinion, overriding the default panic hook does more harm than good and I would like to reconsider including it.

crates/neon-macros/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+mod legacy;
+#[cfg(not(feature = "napi"))]
+use legacy as macros;
+
+// Proc macro definitions must be in the root of the crate
+// Implementations are in the backend dependent module
+
+#[proc_macro_attribute]
+/// Marks a method as the primary entrypoint for initialization in a Neon
+/// module. This attribute should only be used _once_ in a module and will
```
kjvalencik, line 18: State can only be shared within a _single_ crate. Do you think the complexity is worth it for a clearer error message when we can only catch multiple uses _within_ a crate and not across crates?

Using this macro multiple times is a compile time error for N-API and a runtime error for legacy.

```
error: symbol `napi_register_module_v1` is already defined
  --> test/napi/native/src/js/types.rs:63:1
   |
63 | #[neon::init]
   | ^^^^^^^^^^^^^
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

https://github.com/rust-lang/rust/issues/44034#issuecomment-371631945

goto-bus-stop: We could require that users actually call the function `init`. You could technically still have multiple `init` functions in different modules, but…

kjvalencik: For comparison, the exact same bug exists with the current macro and I can't think of a single bug report we've ever had because of that issue. With that said, a proc_macro hides a bit more than the big normal macro.

How would we ensure that `neon::init` is used at least once?

kjvalencik:
Summary from call:
* Rename `neon::init` to `neon::main` as a subtle clue that it should be used exactly once
* File a follow-up issue for catching mistakes (using the macro more or less than once)
* Include the error messages when using more or less than once in the documentation to help Google searches

kjvalencik:
Updated with changes from feedback.

goto-bus-stop (approved):
i like it

PR #639, opened 2020-11-21
feat: impl Date API

amilajack:
Implements https://github.com/neon-bindings/rfcs/pull/18

- [x] impl `JsDate::new`
- [x] impl `JsDate::new_lossy`
- [x] impl `JsDate::try_new`
- [x] impl `JsDate::is_date`
- [x] impl `JsDate::is_valid`
- [x] Write tests

### DIFF
--- a/crates/neon-runtime/src/napi/bindings/functions.rs
+++ b/crates/neon-runtime/src/napi/bindings/functions.rs
@@ -19,6 +19,10 @@ mod napi1 {
 
         fn get_value_double(env: Env, value: Value, result: *mut f64) -> Status;
 
+        fn create_date(env: Env, value: f64, result: *mut Value) -> Status;
+
+        fn get_date_value(env: Env, value: Value, result: *mut f64) -> Status;
+
         fn create_array_with_length(env: Env, length: usize, result: *mut Value) -> Status;
 
         fn get_array_length(env: Env, value: Value, result: *mut u32)-> Status;
@@ -53,6 +57,7 @@ mod napi1 {
         fn is_buffer(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_error(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_array(env: Env, value: Value, result: *mut bool) -> Status;
+        fn is_date(env: Env, value: Value, result: *mut bool) -> Status;
 
         fn get_value_string_utf8(
             env: Env,
@@ -252,7 +257,7 @@ mod napi4 {
 }
 
 #[cfg(feature = "napi-6")]
-mod napi6 {  
+mod napi6 {
     use super::super::types::*;
 
     generate!(extern "C" {
@@ -264,7 +269,7 @@ mod napi6 {
             key_conversion: KeyConversion,
             result: *mut Value,
         ) -> Status;
-    });    
+    });
 }
 
 pub(crate) use napi1::*;
--- a/crates/neon-runtime/src/napi/bindings/functions.rs
+++ b/crates/neon-runtime/src/napi/bindings/functions.rs
@@ -19,10 +19,6 @@ mod napi1 {
 
         fn get_value_double(env: Env, value: Value, result: *mut f64) -> Status;
 
-        fn create_date(env: Env, value: f64, result: *mut Value) -> Status;
-
-        fn get_date_value(env: Env, value: Value, result: *mut f64) -> Status;
-
         fn create_array_with_length(env: Env, length: usize, result: *mut Value) -> Status;
 
         fn get_array_length(env: Env, value: Value, result: *mut u32)-> Status;
@@ -57,7 +53,6 @@ mod napi1 {
         fn is_buffer(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_error(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_array(env: Env, value: Value, result: *mut bool) -> Status;
-        fn is_date(env: Env, value: Value, result: *mut bool) -> Status;
 
         fn get_value_string_utf8(
             env: Env,
@@ -256,6 +251,19 @@ mod napi4 {
     });
 }
 
+#[cfg(feature = "napi-5")]
+mod napi5 {
+    use super::super::types::*;
+
+    generate!(extern "C" {
+        fn create_date(env: Env, value: f64, result: *mut Value) -> Status;
+
+        fn get_date_value(env: Env, value: Value, result: *mut f64) -> Status;
+
+        fn is_date(env: Env, value: Value, result: *mut bool) -> Status;
+    });
+}
+
 #[cfg(feature = "napi-6")]
 mod napi6 {
     use super::super::types::*;
@@ -275,6 +283,8 @@ mod napi6 {
 pub(crate) use napi1::*;
 #[cfg(feature = "napi-4")]
 pub(crate) use napi4::*;
+#[cfg(feature = "napi-5")]
+pub(crate) use napi5::*;
 #[cfg(feature = "napi-6")]
 pub(crate) use napi6::*;
 
@@ -307,6 +317,9 @@ pub(crate) unsafe fn load(env: Env) -> Result<(), libloading::Error> {
 
     #[cfg(feature = "napi-4")]
     napi4::load(&host, version, 4)?;
+
+    #[cfg(feature = "napi-5")]
+    napi5::load(&host, version, 5)?;
 
     #[cfg(feature = "napi-6")]
     napi6::load(&host, version, 6)?;
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -1,17 +1,17 @@
 use std::mem::MaybeUninit;
-use nodejs_sys as napi;
-use raw::{Env, Local};
+use crate::napi::bindings as napi;
+use crate::raw::{Env, Local};
 
 pub unsafe fn new_date(env: Env, value: f64) -> Local {
     let mut local = MaybeUninit::zeroed();
-    let status = napi::napi_create_date(env, value, local.as_mut_ptr());
-    assert_eq!(status, napi::napi_status::napi_ok);
+    let status = napi::create_date(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::Status::Ok);
     local.assume_init()
 }
 
 pub unsafe fn value(env: Env, p: Local) -> f64 {
     let mut value = 0.0;
-    let status = napi::napi_get_date_value(env, p, &mut value as *mut _);
-    assert_eq!(status, napi::napi_status::napi_ok);
+    let status = napi::get_date_value(env, p, &mut value as *mut _);
+    assert_eq!(status, napi::Status::Ok);
     return value;
 }
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -2,14 +2,14 @@ use std::mem::MaybeUninit;
 use nodejs_sys as napi;
 use raw::{Env, Local};
 
-pub unsafe extern "C" fn new_date(env: Env, value: f64) -> Local {
+pub unsafe fn new_date(env: Env, value: f64) -> Local {
     let mut local = MaybeUninit::zeroed();
     let status = napi::napi_create_date(env, value, local.as_mut_ptr());
     assert_eq!(status, napi::napi_status::napi_ok);
     local.assume_init()
 }
 
-pub unsafe extern "C" fn value(env: Env, p: Local) -> f64 {
+pub unsafe fn value(env: Env, p: Local) -> f64 {
     let mut value = 0.0;
     let status = napi::napi_get_date_value(env, p, &mut value as *mut _);
     assert_eq!(status, napi::napi_status::napi_ok);
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -1,0 +1,14 @@
+use nodejs_sys as napi;
+use raw::{Env, Local};
+
+pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, num: f64) {
+    let status = napi::napi_create_date(env, num, out);
+    assert_eq!(status, napi::napi_status::napi_ok);
+}
+
+pub unsafe extern "C" fn value(env: Env, p: Local) -> f64 {
+    let mut value = 0.0;
+    let status = napi::napi_get_date_value(env, p, &mut value as *mut f64);
+    assert_eq!(status, napi::napi_status::napi_ok);
+    return value;
+}
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -1,14 +1,17 @@
+use std::mem::MaybeUninit;
 use nodejs_sys as napi;
 use raw::{Env, Local};
 
-pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, value: f64) {
-    let status = napi::napi_create_date(env, value, out);
+pub unsafe extern "C" fn new_date(env: Env, value: f64) -> Local {
+    let mut local = MaybeUninit::zeroed();
+    let status = napi::napi_create_date(env, value, local.as_mut_ptr());
     assert_eq!(status, napi::napi_status::napi_ok);
+    local.assume_init()
 }
 
 pub unsafe extern "C" fn value(env: Env, p: Local) -> f64 {
     let mut value = 0.0;
-    let status = napi::napi_get_date_value(env, p, &mut value as *mut f64);
+    let status = napi::napi_get_date_value(env, p, &mut value as *mut _);
     assert_eq!(status, napi::napi_status::napi_ok);
     return value;
 }
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -1,8 +1,8 @@
 use nodejs_sys as napi;
 use raw::{Env, Local};
 
-pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, num: f64) {
-    let status = napi::napi_create_date(env, num, out);
+pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, value: f64) {
+    let status = napi::napi_create_date(env, value, out);
     assert_eq!(status, napi::napi_status::napi_ok);
 }
 
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -2,6 +2,11 @@ use std::mem::MaybeUninit;
 use crate::napi::bindings as napi;
 use crate::raw::{Env, Local};
 
+/// Create a new date object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
 pub unsafe fn new_date(env: Env, value: f64) -> Local {
     let mut local = MaybeUninit::zeroed();
     let status = napi::create_date(env, value, local.as_mut_ptr());
@@ -9,6 +14,11 @@ pub unsafe fn new_date(env: Env, value: f64) -> Local {
     local.assume_init()
 }
 
+/// Get the value of a date object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
 pub unsafe fn value(env: Env, p: Local) -> f64 {
     let mut value = 0.0;
     let status = napi::get_date_value(env, p, &mut value as *mut _);
--- a/crates/neon-runtime/src/napi/date.rs
+++ b/crates/neon-runtime/src/napi/date.rs
@@ -19,6 +19,7 @@ pub unsafe fn new_date(env: Env, value: f64) -> Local {
 /// # Safety
 ///
 /// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+/// `Local` must be an NAPI value associated with the given `Env`
 pub unsafe fn value(env: Env, p: Local) -> f64 {
     let mut value = 0.0;
     let status = napi::get_date_value(env, p, &mut value as *mut _);
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -18,7 +18,7 @@ use handle::{Managed, Handle};
 use event::EventQueue;
 use types::{JsValue, Value, JsObject, JsArray, JsFunction, JsBoolean, JsNumber, JsString, StringResult, JsNull, JsUndefined};
 #[cfg(feature = "napi-5")]
-use types::JsDate;
+use types::date::{JsDate, DateError};
 #[cfg(feature = "napi-1")]
 use types::boxed::{Finalize, JsBox};
 use types::binary::{JsArrayBuffer, JsBuffer};
@@ -29,6 +29,7 @@ use object::class::Class;
 use result::{NeonResult, JsResult, Throw};
 #[cfg(feature = "napi-1")]
 use smallvec::SmallVec;
+
 use self::internal::{ContextInternal, Scope, ScopeMetadata};
 
 #[repr(C)]
@@ -316,7 +317,7 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(feature = "napi-5")]
-    fn date(&mut self, value: impl Into<f64>) -> Handle<'a, JsDate> {
+    fn date(&mut self, value: impl Into<f64> + PartialOrd) -> Result<Handle<'a, JsDate>, DateError> {
         JsDate::new(self, value)
     }
 
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -316,7 +316,7 @@ pub trait Context<'a>: ContextInternal<'a> {
         JsBuffer::new(self, size)
     }
 
-    /// Create a `JsDate`. This prevents
+    /// Convenience method for creating a `JsDate` value.
     #[cfg(feature = "napi-5")]
     fn date(&mut self, value: impl Into<f64>) -> Result<Handle<'a, JsDate>, DateError> {
         JsDate::new(self, value)
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -16,7 +16,9 @@ use context::internal::Env;
 use handle::{Managed, Handle};
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
 use event::EventQueue;
-use types::{JsValue, Value, JsObject, JsArray, JsFunction, JsBoolean, JsNumber, JsString, StringResult, JsNull, JsUndefined, JsDate};
+use types::{JsValue, Value, JsObject, JsArray, JsFunction, JsBoolean, JsNumber, JsString, StringResult, JsNull, JsUndefined};
+#[cfg(feature = "napi-5")]
+use types::JsDate;
 #[cfg(feature = "napi-1")]
 use types::boxed::{Finalize, JsBox};
 use types::binary::{JsArrayBuffer, JsBuffer};
@@ -313,7 +315,7 @@ pub trait Context<'a>: ContextInternal<'a> {
         JsBuffer::new(self, size)
     }
 
-    #[cfg(feature = "napi-runtime")]
+    #[cfg(feature = "napi-5")]
     fn date(&mut self, value: impl Into<f64>) -> Handle<'a, JsDate> {
         JsDate::new(self, value)
     }
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -16,7 +16,7 @@ use context::internal::Env;
 use handle::{Managed, Handle};
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
 use event::EventQueue;
-use types::{JsValue, Value, JsObject, JsArray, JsFunction, JsBoolean, JsNumber, JsString, StringResult, JsNull, JsUndefined};
+use types::{JsValue, Value, JsObject, JsArray, JsFunction, JsBoolean, JsNumber, JsString, StringResult, JsNull, JsUndefined, JsDate};
 #[cfg(feature = "napi-1")]
 use types::boxed::{Finalize, JsBox};
 use types::binary::{JsArrayBuffer, JsBuffer};
@@ -135,12 +135,12 @@ impl<'a> Lock<'a> {
 }
 
 /// An _execution context_, which provides context-sensitive access to the JavaScript engine. Most operations that interact with the engine require passing a reference to a context.
-/// 
+///
 /// A context has a lifetime `'a`, which ensures the safety of handles managed by the JS garbage collector. All handles created during the lifetime of a context are kept alive for that duration and cannot outlive the context.
 pub trait Context<'a>: ContextInternal<'a> {
 
     /// Lock the JavaScript engine, returning an RAII guard that keeps the lock active as long as the guard is alive.
-    /// 
+    ///
     /// If this is not the currently active context (for example, if it was used to spawn a scoped context with `execute_scoped` or `compute_scoped`), this method will panic.
     fn lock(&self) -> Lock<'_> {
         self.check_active();
@@ -148,9 +148,9 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     /// Convenience method for locking the JavaScript engine and borrowing a single JS value's internals.
-    /// 
+    ///
     /// # Example:
-    /// 
+    ///
     /// ```no_run
     /// # use neon::prelude::*;
     /// # fn my_neon_function(mut cx: FunctionContext) -> JsResult<JsNumber> {
@@ -160,7 +160,7 @@ pub trait Context<'a>: ContextInternal<'a> {
     /// # Ok(n)
     /// # }
     /// ```
-    /// 
+    ///
     /// Note: the borrowed value is required to be a reference to a handle instead of a handle
     /// as a workaround for a [Rust compiler bug](https://github.com/rust-lang/rust/issues/29997).
     /// We may be able to generalize this compatibly in the future when the Rust bug is fixed,
@@ -177,9 +177,9 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     /// Convenience method for locking the JavaScript engine and mutably borrowing a single JS value's internals.
-    /// 
+    ///
     /// # Example:
-    /// 
+    ///
     /// ```no_run
     /// # use neon::prelude::*;
     /// # fn my_neon_function(mut cx: FunctionContext) -> JsResult<JsUndefined> {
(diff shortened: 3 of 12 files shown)


.github/workflows/linux.yml
```diff
@@ -33,8 +33,9 @@ jobs:
       run: sudo apt-get install -y g++-4.8
     - name: Start Xvfb
       run: |
-        Xvfb :99 &
-        disown -ar
-        echo "::set-env name=DISPLAY:::99"
+        sudo apt-get install libwoff1 libopus0 libwebp6 libwebpdemux2 libenchant1c2a libgudev-1.0-0 libsecret-1-0 libhyphen0 libgdk-pixbuf2.0-0
```
amilajack, line 36: Taken from https://github.com/microsoft/playwright/actions/runs/25967532/workflow

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,14 @@
+use nodejs_sys as napi;
+use raw::{Env, Local};
+
+pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, num: f64) {
```
dherman, line 4: I'd probably call the `num` parameter `value`, for consistency with the `value()` method.

src/context/mod.rs
```diff
@@ -318,6 +320,11 @@ pub trait Context<'a>: ContextInternal<'a> {
         JsBuffer::new(self, size)
     }
 
+    #[cfg(feature = "napi-runtime")]
+    fn date<V: Into<f64>>(&mut self, date: V) -> Handle<'a, JsDate> {
```
dherman, line 324: Oh good catch, this is missing from the RFC. I added it.

I'd call the `date` parameter here `value` as well.

src/types/mod.rs
```diff
@@ -424,6 +424,62 @@ impl ValueInternal for JsNumber {
...
+    fn to_raw(self) -> raw::Local { self.0 }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self { JsDate(h) }
+}
+
+impl JsDate {
+    pub const MIN_VALID_VALUE: f64 = -8640000000000000.0;
+    pub const MAX_VALID_VALUE: f64 = 8640000000000000.0;
+
+    #[cfg(feature = "napi-runtime")]
```
dherman, line 444: This feature flag should be around the top level declarations instead of the individual methods. IOW, the `JsDate` type shouldn't even exist unless the `"napi-runtime"` feature is enabled.

src/types/mod.rs
```diff
@@ -424,6 +424,62 @@ impl ValueInternal for JsNumber {
...
+    pub(crate) fn new_internal<'a>(env: Env, value: f64) -> Handle<'a, JsDate> {
+        unsafe {
+            let mut local: raw::Local = std::mem::zeroed();
```
dherman, line 447: @kjvalencik Do we want to use `MaybeUninit` for this kind of pattern instead of `std::mem::zeroed()`? I'm not sure what our current best practice is.

src/types/mod.rs
```diff
@@ -424,6 +424,62 @@ impl ValueInternal for JsNumber {
...
+        let value = self.value(cx);
+        return value <= JsDate::MAX_VALID_VALUE && value >= JsDate::MIN_VALID_VALUE;
+    }
+}
+
+impl ValueInternal for JsDate {
+    fn name() -> String { "object".to_string() }
+
+    fn is_typeof<Other: Value>(env: Env, other: Other) -> bool {
+        unsafe { neon_runtime::tag::is_object(env.to_raw(), other.to_raw()) }
```
dherman, line 477: I think we probably want this to check if the value is a `Date` object, not just any object. We could add a `neon_runtime::tag::is_date` method to the runtime, which uses [napi_is_date](https://nodejs.org/api/n-api.html#n_api_napi_is_date) to do the check.

dherman (changes requested):
Looks good! Just a few minor comments.

.github/workflows/linux.yml
kjvalencik: These deps shouldn't be needed for our tests.

.github/workflows/linux.yml
```diff
@@ -33,8 +33,9 @@ jobs:
...
+        sudo apt-get install xvfb
+        export DISPLAY=:99.0
```
kjvalencik, line 38: This won't have any impact on the test. I think the only fix that was needed was how actions handles environment variables.

.github/workflows/linux.yml
```diff
@@ -33,8 +33,9 @@ jobs:
...
+        Xvfb -ac :99 -screen 0 1280x1024x16 > /dev/null 2>&1 &
     - name: run cargo test
-      run: cargo test --release -- --nocapture
+      run: xvfb-run --auto-servernum cargo test --release -- --nocapture
```
kjvalencik, line 41: Can we remove the entire previous step and use just this? It looks cleaner than forking.

src/types/mod.rs
```diff
@@ -424,6 +424,64 @@ impl ValueInternal for JsNumber {
...
+#[cfg(feature = "napi-runtime")]
+impl Managed for JsDate {
+    fn to_raw(self) -> raw::Local { self.0 }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self { JsDate(h) }
+}
+
+#[cfg(feature = "napi-runtime")]
+impl JsDate {
+    pub const MIN_VALID_VALUE: f64 = -8640000000000000.0;
```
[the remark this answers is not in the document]
kjvalencik: These are great additions! Can we name these `MAX_VALUE` and `MIN_VALUE`?

src/types/mod.rs
```diff
@@ -424,6 +424,64 @@ impl ValueInternal for JsNumber {
...
+    pub fn value<'a, C: Context<'a>>(self, cx: &mut C) -> f64 {
+        let env = cx.env().to_raw();
+        unsafe {
+            neon_runtime::date::value(env, self.to_raw())
+        }
+    }
+
+    pub fn is_valid<'a, C: Context<'a>>(self, cx: &mut C) -> bool {
+        let value = self.value(cx);
+        return value <= JsDate::MAX_VALID_VALUE && value >= JsDate::MIN_VALID_VALUE;
```
kjvalencik, line 469: Does n-api really return the original value here? In Js it returns `NaN`. Is this documented behavior?

.github/workflows/linux.yml
kjvalencik: This was a great find! It turns out, this was the only line needed and we can remove the entire previous step!

https://github.com/neon-bindings/neon/pull/640

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,14 @@
+use nodejs_sys as napi;
+use raw::{Env, Local};
+
+pub unsafe extern "C" fn new_date(env: Env, out: *mut Local, value: f64) {
+    let status = napi::napi_create_date(env, value, out);
+    assert_eq!(status, napi::napi_status::napi_ok);
```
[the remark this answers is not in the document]
amilajack: It doesn't panic. I tested this in the `check_date_is_invalid` test. It still returns a Date object but `is_valid` returns false.

src/types/mod.rs
```diff
@@ -424,6 +424,64 @@ impl ValueInternal for JsNumber {
...
+
+    pub(crate) fn new_internal<'a>(env: Env, value: f64) -> Handle<'a, JsDate> {
+        unsafe {
+            let mut local: raw::Local = std::mem::zeroed();
+            neon_runtime::date::new_date( env.to_raw(), &mut local, value);
+            Handle::new_internal(JsDate(local))
+        }
+    }
+
+    pub fn new<'a, C: Context<'a>, T: Into<f64>>(cx: &mut C, time: T) -> Handle<'a, JsDate> {
```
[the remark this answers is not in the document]
amilajack: I think this is an interesting idea. I'm wondering if its consistent with existing notions of when we return `Result`.

src/types/mod.rs
amilajack: `self.value` returns `std::f64::NAN` if >= 2**53. Will add test case for this.

.github/workflows/linux.yml
```diff
@@ -32,4 +32,4 @@ jobs:
     - name: install g++-4.8
       run: sudo apt-get install -y g++-4.8
     - name: run cargo test
-      run: xvfb-run --auto-servernum cargo test --release -- --nocapture
+      run: cargo test --release -- --nocapture
```
kjvalencik, line 35: Looks like this might have gotten reverted in the merge.

src/types/mod.rs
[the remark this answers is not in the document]
dherman: Un-resolving this one just for discussion -- my only reason for naming them "VALID" in the RFC was that higher values are possible, they just lead to invalid dates. E.g., you can do:

```rust
cx.date(JsDate::MAX_VALID_VALUE + 1)
```

and it's not an error, it just produces a `Date` object that doesn't represent a valid date.

@kjvalencik Does that change your perspective at all, or do you think the "VALID" is still overkill? Whatever we decide, I'll update the RFC PR to match it.

src/types/mod.rs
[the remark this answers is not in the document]
dherman: @kjvalencik Interesting idea. It feels a _little_ unidiomatic to me, or somehow strange to say "I failed to create a `Date` and btw here it is."

Oh interesting -- in the RFC, I had it returning a `JsResult` in case the `napi_status` is not `napi_ok`, but is that impossible? I guess this API can never throw a JS exception?

If that's the case, another option would be two or three different entry points, with a custom `DateErrorKind`:

```rust
pub enum DateErrorKind {
    Overflow,
    Underflow,
}

impl JsDate {
    // checks the range and returns an Error if out of range
    pub fn try_new<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> Result<Handle<'a, JsDate>, DateErrorKind>;
    // checks the range and panics if out of range
    pub fn new<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> Handle<'a, JsDate>;
    // doesn't check the range and produces a Date value no matter what
    pub fn new_unchecked<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> Handle<'a, JsDate>;
}
```

And then for convenience we'd have `cx.date(v)` be equivalent to `JsDate::new(&mut cx, v)` and maybe `cx.try_date(v)` would be equivalent to `JsDate::try_new(&mut cx, v)` (similar to `cx.try_string(...)`). I wouldn't make a convenience method for `new_unchecked` since it's probably a rare case.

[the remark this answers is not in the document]
amilajack: I think the three different kinds of entrypoints might be too much API surface area. I favor going with the 'checks the range and panics if out of range' API and adding additional APIs if users find it insufficient for their use cases.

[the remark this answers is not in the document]
kjvalencik: I agree that three seems like one to many, but a panic on a recoverable error doesn't seem ideal to me. I _really_ like the `try_new` with the `DateError`. One of the reasons I like it is it allows us to extend this function to accept `i64` as well since out of range for the date implies overflow on an `f64`.

src/types/mod.rs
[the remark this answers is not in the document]
kjvalencik: @dherman I couldn't find anything in the specification that indicated what happens with an out of range value. I don't feel strongly on this naming.

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,17 @@
+use std::mem::MaybeUninit;
+use nodejs_sys as napi;
+use raw::{Env, Local};
+
+pub unsafe extern "C" fn new_date(env: Env, value: f64) -> Local {
```
kjvalencik, line 5: Fix please, this method shouldn't be `extern "C"`

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,17 @@
...
+    let mut local = MaybeUninit::zeroed();
+    let status = napi::napi_create_date(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::napi_status::napi_ok);
+    local.assume_init()
+}
+
+pub unsafe extern "C" fn value(env: Env, p: Local) -> f64 {
```
kjvalencik, line 12: Fix please, this method shouldn't be `extern "C"`

src/types/mod.rs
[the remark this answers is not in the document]
amilajack: So is there consensus on `MAX_VALID_VALUE` and `MIN_VALID_VALUE`?

src/types/mod.rs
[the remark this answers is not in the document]
amilajack: +1. Will implement `try_new`

arthrarnld:
Hey peeps. Any news on this one? Having it would be really helpful in our team. Thanks!

kjvalencik:
@arthrarnld It's a work in progress. We're still discussing the best way to handle an out of range time. Let us know if you have any thoughts!

test/napi/src/js/date.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    let time = JsDate::MIN_VALUE - 1.0;
+    let date = cx.date(time).value(&mut cx);
+    assert!(!cx.date(time).is_valid(&mut cx));
+    assert!(cx.date(time).value(&mut cx).is_nan());
+
+    let time = JsDate::MAX_VALUE + 2.0;
+    let date = cx.date(time).value(&mut cx);
+    assert!(!cx.date(time).is_valid(&mut cx));
+    assert!(cx.date(time).value(&mut cx).is_nan());
```
amilajack, line 52: @dherman Added the test cases you mentioned

src/types/mod.rs
```diff
@@ -424,6 +424,76 @@ impl ValueInternal for JsNumber {
...
+    }
+
+    pub fn try_new<'a, C: Context<'a>, V: Into<f64> + std::cmp::PartialOrd>(cx: &mut C, value: V) -> Result<Handle<'a, JsDate>, DateErrorKind> {
+        let env = cx.env().to_raw();
+        let local = unsafe {
+            neon_runtime::date::new_date(env, value.into())
+        };
+        let date = Handle::new_internal(JsDate(local));
+        if date.is_valid(cx) { return Ok(date); }
+        if date.value(cx) > JsDate::MAX_VALUE { Err(DateErrorKind::Overflow) } else { Err(DateErrorKind::Underflow) }
```
kjvalencik, line 469: This check should occur as the first step if this function. We aren't guaranteed that `date.value(cx)` will return a non-NaN value.

Separately, this function will return `Ok` if the user passes `NaN`. That seems correct to me, but I wanted to confirm others agree.

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -19,6 +19,10 @@ mod napi1 {
 
         fn get_value_double(env: Env, value: Value, result: *mut f64) -> Status;
 
+        fn create_date(env: Env, value: f64, result: *mut Value) -> Status;
```
kjvalencik, line 22: This API was added in N-API 5 and not N-API 1. Please add a `napi5` module add add it to loading in `fn load` at the bottom of this file.

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -19,6 +19,10 @@ mod napi1 {
...
+
+        fn get_date_value(env: Env, value: Value, result: *mut f64) -> Status;
```
kjvalencik, line 24: Also, N-API 5.

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -53,6 +57,7 @@ mod napi1 {
         fn is_buffer(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_error(env: Env, value: Value, result: *mut bool) -> Status;
         fn is_array(env: Env, value: Value, result: *mut bool) -> Status;
+        fn is_date(env: Env, value: Value, result: *mut bool) -> Status;
```
kjvalencik, line 60: Version 5.

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,17 @@
+use std::mem::MaybeUninit;
+use crate::napi::bindings as napi;
+use crate::raw::{Env, Local};
+
+pub unsafe fn new_date(env: Env, value: f64) -> Local {
```
kjvalencik, line 5: Please add doc comments with a `# Safety` section.

src/context/mod.rs
```diff
@@ -313,6 +315,11 @@ pub trait Context<'a>: ContextInternal<'a> {
         JsBuffer::new(self, size)
     }
 
+    #[cfg(feature = "napi-5")]
+    fn date(&mut self, value: impl Into<f64>) -> Handle<'a, JsDate> {
```
kjvalencik, line 319: Please change this to the version that returns a `Result`.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
     }
 }
 
+/// A JavaScript Date object.
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+#[cfg(feature = "napi-5")]
+pub struct JsDate(raw::Local);
+
+#[cfg(feature = "napi-5")]
```
kjvalencik, line 435: I think it will be cleaner to crate a `date.rs` file that is conditionally declared instead of adding the attribute in many places.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
+pub enum DateErrorKind {
+    Overflow,
+    Underflow,
+}
+
+#[cfg(feature = "napi-5")]
+impl DateErrorKind {
+    pub fn as_str(&self) -> &'static str {
```
kjvalencik, line 475: This should not be public.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+impl DateError {
+    pub fn kind(&self) -> DateErrorKind {
+        self.1
+    }
+}
+
+#[cfg(feature = "napi-5")]
+impl fmt::Display for DateError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}", self.1.as_str())
```
kjvalencik, line 459: I think this should have a more idiomatic display implementation. For example, `DateError(Overflow)`.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+#[cfg(feature = "napi-5")]
+impl Managed for JsDate {
+    fn to_raw(self) -> raw::Local { self.0 }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self { JsDate(h) }
+}
+
+#[cfg(feature = "napi-5")]
+#[derive(PartialEq, PartialOrd, Clone, Debug)]
+pub struct DateError(f64, DateErrorKind);
```
kjvalencik, line 447: The `f64` appears to be unused. Can it be removed?

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+    pub fn as_str(&self) -> &'static str {
+        match *self {
+            DateErrorKind::Overflow => "date overflow",
+            DateErrorKind::Underflow => "date underflow",
+        }
+    }
+}
+
+#[cfg(feature = "napi-5")]
+pub type DateResult<'a> = Result<Handle<'a, JsDate>, DateError>;
```
kjvalencik, line 484: `Result` aliases are not considered best practice anymore. I think this should be removed, but it should at least not be public.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+
+#[cfg(feature = "napi-5")]
+impl<'a> JsResultExt<'a, JsDate> for DateResult<'a> {
+    /// Creates a `Error` on error
+    fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, JsDate> {
+        match self {
+            Ok(v) => Ok(v),
+            Err(e) => cx.throw_error(&e.to_string())
```
kjvalencik, line 492: I think this would be more clear as a `RangeError` instead of `Error`.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+        }
+    }
+}
+
+#[cfg(feature = "napi-5")]
+impl JsDate {
+    pub const MIN_VALUE: f64 = -8.64e15;
+    pub const MAX_VALUE: f64 = 8.64e15;
+
+    pub fn new<'a, C: Context<'a>, T: Into<f64>>(cx: &mut C, time: T) -> Handle<'a, JsDate> {
```
kjvalencik, line 502: The names on these functions are reversed. `new` should return a `Result` and `new_lossy` should always return the date (which could lose the `f64`).

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+    pub fn new_lossy<'a, C: Context<'a>, V: Into<f64> + std::cmp::PartialOrd>(cx: &mut C, value: V) -> DateResult<'a> {
+        let env = cx.env().to_raw();
+        let time = value.into();
+        let local = unsafe {
+            neon_runtime::date::new_date(env, time.clone())
+        };
+        let date = Handle::new_internal(JsDate(local));
+        if date.is_valid(cx) { return Ok(date); }
+
+        if time > JsDate::MAX_VALUE {
```
kjvalencik, line 519: Can we move this to the beginning of the function right after `let time = value.into();`? It saves the construction of the date object if we're going to throw it away.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+    pub fn value<'a, C: Context<'a>>(self, cx: &mut C) -> f64 {
+        let env = cx.env().to_raw();
+        unsafe {
+            neon_runtime::date::value(env, self.to_raw())
+        }
+    }
+
+    pub fn is_valid<'a, C: Context<'a>>(self, cx: &mut C) -> bool {
+        let value = self.value(cx);
+        return value <= JsDate::MAX_VALUE && value >= JsDate::MIN_VALUE;
```
kjvalencik, line 535: Remove unnecessary `return` from the statement.

src/types/mod.rs
```diff
@@ -424,6 +426,128 @@ impl ValueInternal for JsNumber {
...
+    }
+
+    pub fn new_lossy<'a, C: Context<'a>, V: Into<f64> + std::cmp::PartialOrd>(cx: &mut C, value: V) -> DateResult<'a> {
+        let env = cx.env().to_raw();
+        let time = value.into();
+        let local = unsafe {
+            neon_runtime::date::new_date(env, time.clone())
+        };
+        let date = Handle::new_internal(JsDate(local));
+        if date.is_valid(cx) { return Ok(date); }
```
kjvalencik, line 517: This is redundant with the comparisons against `time` and can be removed.

kjvalencik (changes requested):

src/types/mod.rs
kjvalencik: Changing my mind on this. JavaScript errors tend to me more text focused. For example:

```
> new Array(-1)
Uncaught RangeError: Invalid array length
```

Maybe we can mirror that form?

src/types/mod.rs
amilajack: Good catch! What is the expected behavior when NaN is passed to construct a date?

kjvalencik: In my opinion, the expected behavior is to successfully return a JS `Date(Invalid Date)` since it wouldn't result in data loss (Rust doesn't differentiate different NaN values).

src/types/mod.rs
amilajack: +1, good idea!

crates/neon-runtime/src/napi/date.rs
amilajack: I don't think I understand how this API is unsafe. I see that this is using the `unsafe` keyword but I'm still not really sure what this code is doing.

kjvalencik: `Env` is a raw pointer. It's the caller's responsibility to ensure it points to a `napi_env` that is valid for the current context.

amilajack: Ah okay that makes sense, thanks! 😃

crates/neon-runtime/src/napi/date.rs
```diff
@@ -0,0 +1,27 @@
...
+    let status = napi::create_date(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::Status::Ok);
+    local.assume_init()
+}
+
+/// Get the value of a date object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
```
kjvalencik, line 21: This one has an additional safety requirement that `Local` be a `napi_value` associated with the `Env`.

src/context/mod.rs
```diff
@@ -313,6 +316,11 @@ pub trait Context<'a>: ContextInternal<'a> {
         JsBuffer::new(self, size)
     }
 
+    #[cfg(feature = "napi-5")]
+    fn date(&mut self, value: impl Into<f64> + PartialOrd) -> Result<Handle<'a, JsDate>, DateError> {
```
kjvalencik, line 320: Why does this need to be `PartialOrd`? Shouldn't we convert to `f64` before performing a comparison?

src/handle/mod.rs
```diff
@@ -29,7 +29,7 @@ pub trait Managed: Copy {
 
 /// A safely rooted _handle_ to a JS value in memory that is managed by the garbage collector.
 #[repr(C)]
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
```
kjvalencik, line 32: This is a good change (https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits), but my preference is to minimize unrelated changes in a PR.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl Value for JsDate { }
+
+impl Managed for JsDate {
+    fn to_raw(self) -> raw::Local { self.0 }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self { JsDate(h) }
+}
+
+#[derive(PartialEq, PartialOrd, Clone, Debug)]
```
kjvalencik, line 27: It seems strange to implement anything besides `Debug` on this type and possibly `Clone`.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl DateError {
+    pub fn kind(&self) -> DateErrorKind {
+        self.0
+    }
+}
+
+impl fmt::Display for DateError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}", self.0.as_str())
```
kjvalencik, line 38: FYI, since there is no additional formatting, this can be written more tersely, although I don't think there is much of an impact on the generated code.

```rust
fmt.write_str(self.0.as_str())
```

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl Error for DateError {}
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
```
kjvalencik, line 44: Why is this deriving `PartialOrd` and `Ord`? thsoe in particular seem strange.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+pub enum DateErrorKind {
+    Overflow,
+    Underflow,
+}
+
+impl DateErrorKind {
+    fn as_str(&self) -> &'static str {
+        match *self {
+            DateErrorKind::Overflow => "Uncaught RangeError: Date overflow",
```
kjvalencik, line 53: These look like strange formatting. Why do they include `Uncaught RangeError`? At this point, they might be caught. It would also result in `Uncaught RangeError: UncaughtRangeError: Date Overflow` when thrown.

Can make this more simply `Date overflow`?

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+            DateErrorKind::Underflow => "Uncaught RangeError: Date underflow",
+        }
+    }
+}
+
+impl<'a> JsResultExt<'a, JsDate> for Result<Handle<'a, JsDate>, DateError> {
```
kjvalencik, line 59: This can be more general. We don't use the `Ok` type, so we should implement it for all `T` in case we want to use it in other places.

Unfortunately, `JsResultExt` unnecessarily requires `T: Value`. I would like to fix that in a future breaking change.

Also, FYI, the [`Result::or_else`](https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else) combinator performs the same thing as this `match`:

```rust
impl<'a, T: Value> JsResultExt<'a, T> for Result<Handle<'a, T>, DateError> {
    fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, T> {
        self.or_else(|e| cx.throw_range_error(e.0.as_str()))
    }
}
```

Lastly, this includes a micro optimization of getting the `&'static str` directly from the error kind. This eliminates allocating a `String`; however, it's likely an optimization pass would remove the allocation anyway because it's trivial to see it isn't mutated.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, JsDate> {
+        match self {
+            Ok(v) => Ok(v),
+            Err(e) => cx.throw_range_error(&e.to_string())
+        }
+    }
+}
+
+impl JsDate {
+    pub const MIN_VALUE: f64 = -8.64e15;
```
kjvalencik, line 70: Nit, it would be nice to add a link to the spec for these values.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+        let time = value.into();
+
+        if time > JsDate::MAX_VALUE {
+            return Err(DateError(DateErrorKind::Overflow))
+        } else if time < JsDate::MIN_VALUE {
+            return Err(DateError(DateErrorKind::Underflow))
+        }
+
+        let local = unsafe {
+            neon_runtime::date::new_date(env, time.clone())
```
kjvalencik, line 84: This `clone` appears unnecessary. `new_date` can take ownership since `time` is not used again. Additionally, `time` is an `f64` which implements `Copy`, explicit cloning wouldn't be necessary.

src/context/mod.rs
kjvalencik, line 320: Please add a doc comment.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+impl Value for JsDate { }
+
+impl Managed for JsDate {
+    fn to_raw(self) -> raw::Local { self.0 }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self { JsDate(h) }
+}
+
+#[derive(PartialEq, PartialOrd, Clone, Debug)]
+pub struct DateError(DateErrorKind);
```
kjvalencik, line 28: Please add a doc comment.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+            Err(e) => cx.throw_range_error(&e.to_string())
+        }
+    }
+}
+
+impl JsDate {
+    pub const MIN_VALUE: f64 = -8.64e15;
+    pub const MAX_VALUE: f64 = 8.64e15;
+
+    pub fn new<'a, C: Context<'a>, T: Into<f64>>(cx: &mut C, value: T) -> Result<Handle<'a, JsDate>, DateError> {
```
kjvalencik, line 73: Please add a doc comment. It should include the behavior of passing an out of bounds value or `NaN`.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        let local = unsafe {
+            neon_runtime::date::new_date(env, time.clone())
+        };
+        let date = Handle::new_internal(JsDate(local));
+        Ok(date)
+    }
+
+    pub fn new_lossy<'a, C: Context<'a>, V: Into<f64> + std::cmp::PartialOrd>(cx: &mut C, value: V) -> Handle<'a, JsDate> {
```
kjvalencik, line 90: Please add a doc comment. It should include the behavior of passing an out of bounds value.

Do we need `PartialOrd`?

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+        let env = cx.env().to_raw();
+        let local = unsafe {
+            neon_runtime::date::new_date(env, value.into())
+        };
+        Handle::new_internal(JsDate(local))
+    }
+
+    pub fn value<'a, C: Context<'a>>(self, cx: &mut C) -> f64 {
```
kjvalencik, line 98: Please add a doc comment.

src/types/date.rs
```diff
@@ -0,0 +1,119 @@
...
+        let env = cx.env().to_raw();
+        unsafe {
+            neon_runtime::date::value(env, self.to_raw())
+        }
+    }
+
+    pub fn is_valid<'a, C: Context<'a>>(self, cx: &mut C) -> bool {
```
kjvalencik, line 105: Please add a doc comment.

src/types/date.rs
amilajack: > Do we need PartialOrd?

Nope, I had this earlier because I was trying to convert the value before calling `into` 🤦

src/handle/mod.rs
amilajack: This was needed to call `date_underflow.unwrap_err` in tests

src/types/date.rs
amilajack: I think `Value` requires that JsDate implement `Clone` and `Copy`.

amilajack: https://github.com/neon-bindings/neon/blob/main/src/types/mod.rs#L72-L73

src/types/date.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl JsDate {
+    /// The smallest possible Date value, defined by ECMAScript. See https://www.ecma-international.org/ecma-262/5.1/#sec-15.7.3.3
+    pub const MIN_VALUE: f64 = -8.64e15;
+    /// The largest possible Date value, defined by ECMAScript. See https://www.ecma-international.org/ecma-262/5.1/#sec-15.7.3.2
+    pub const MAX_VALUE: f64 = 8.64e15;
+
+    /// Create a new Date. It errors when `value` is an out of bounds JavaScript Date value. When `value`
+    /// is `NaN`, an invalid
```
amilajack, line 74: Forgot to finish comment here. Will add in followup commit

kjvalencik (approved):
@amilajack This looks great! Before merging, let's decide whether we want this feature flagged or not prior to merging the RFC.

Issue #584, opened 2020-08-08
Dynamically loading N-API functions


goto-bus-stop:
This issue explores the N-API linking story on Windows and cross-platform.
I tried to make it understandable for people who haven't done any Windows
programming but please ask if anything's unclear 🙇‍♀️ 

## Summary
We could manually load pointers to N-API functions using `GetModuleHandle`
and `dlopen()`. On Windows, this lets us avoid a compile-time dependency on
`node.lib` and makes the `win_delay_load_hook` unnecessary.

## Background
When your Neon addon is loaded, it looks up the address of all the `napi_*`
functions that it needs. Otherwise, it could not call those functions. These
`napi_*` functions are provided by the Node.js executable itself.

On Linux, my understanding is that the process has one big namespace for all the
functions. So when the addon is loaded, the OS's runtime linker can get to work
and finds all the `napi_*` addresses for you, no matter where in the current
process they are defined.

On Windows, things are a bit stricter. You not only need to tell it about the
name of the function, but also which module (exe or dll file) contains it. This
is done at compile time using a `.lib` file (`node.lib` for N-API). This .lib
file contains entries like "find `napi_create_object` in node.exe". When the addon
is loaded, the OS's runtime linker looks up all the `napi_*` addresses inside
`node.exe`.

## The Problems

This setup causes two complications for Neon.

1. We need to have this `node.lib` file available at build time, because we need
   to link to it. This file does not ship with Node.js installations, at least
   not on Windows.
2. Not all Node.js executables are called "node.exe". Node.js can be embedded
   inside other applications. The major one is "electron.exe". If we build a
   Neon addon with Node.js's `node.lib`, Windows will look for N-API functions
   in "node.exe". But, if we are _running_ the node addon inside Electron, there
   is no "node.exe": this causes Windows to look for a node.exe at some
   predefined paths on the system and load it if it exists. Whether "node.exe"
   exists or not, the end result is not good.

   So, what we really want to tell Windows is to look up `napi_*` functions _in
   the host process_, whether it is named "node.exe" or "electron.exe" or
   something else.

The obvious way to address problem 1 is by downloading it in a `build.rs`
script. The more interesting one is problem 2.

<a id=delay-load-hook></a>
## A Solution
node-gyp solves this using a delayed loading hook. MSVC and Windows have a neat
feature where you can specify that a particular module that you link to should
not be loaded at startup, but only once you use its functions. For Neon, that
[...]

PR #645, opened 2020-11-26
Dynamic loading sample

goto-bus-stop:
This PR implements the approach described in #584, and uses it in the functions in `neon_runtime::napi::primitive`.

The N-API symbols are loaded from the host process, which could be Node.js or Electron. We have to provide our own N-API type declarations with this approach. Since N-API is ABI-stable, that should not require a lot of maintenance.

Here, we use `libloading` to retrieve symbols from the host executable. We store raw function pointers, bypassing some `libloading` safety features. This is OK for us because we run as a module inside the host executable: the executable will not be unloaded before we are, so the pointers will not become invalid at any point while we are running. Currently, this sample uses hand-written trampoline functions to call the actual function pointers. The original PR used a macro (described below).

## Original PR description
Some changes were made to the approach. This was the original description:

> The gist of the system is:
> ```rust
> struct Napi<'a> {
>   napi_functionname: Symbol<'a, unsafe extern "C" fn(argtypes...) -> NapiStatus>,
> }
> impl Napi<'_> {
>   pub fn from_host() -> Self {
>     Self {
>       napi_functionname: load_host_library().get("napi_functionname").unwrap(),
>     }
>   }
> }
> ```
> 
> An instance of this `Napi` struct is initialized when the Neon module loads. Calls to N-API functions are made  > like:
> ```rust
> let status = napi!(napi_functionname(args...));
> ```
> Which translates to something like
> ```rust
> let status = ((*crate::bindings::NAPI).napi_functionname)(args...);
> ```

<details>
<summary>
Silly(?) macro idea
</summary>

Maybe we can make a macro like this to ease future maintenance:
```rust
declare_napi_functions! {
    fn napi_get_undefined(env: NapiEnv, out: *mut NapiValue) -> NapiStatus;
    fn napi_get_null(env: NapiEnv, out: *mut NapiValue) -> NapiStatus;

    fn napi_get_boolean(env: NapiEnv, value: bool, out: *mut NapiValue) -> NapiStatus;
    fn napi_get_value_bool(env: NapiEnv, value: NapiValue, out: *mut bool) -> NapiStatus;

    fn napi_create_double(env: NapiEnv, value: f64, out: *mut NapiValue) -> NapiStatus;
    fn napi_get_value_double(env: NapiEnv, value: NapiValue, out: *mut f64) -> NapiStatus;
}
```

I think it would have to be a proc macro, since we need to output those declarations several
times in different shapes. It would allow generating both the property declarations in the Napi
struct, and the `library.get()` calls in the `from_host` constructor. It could also generate
trampoline functions so we don't have to use an `napi!()` macro wrapper for napi function calls:

```rust
pub unsafe fn napi_get_undefined(env: NapiEnv, out: *mut NapiValue) -> NapiStatus {
    ((*NAPI).napi_get_undefined)(env, out)
}
```

Might well be a case where writing the abstraction would be a lot more work than just doing it by hand, though.

</details>

### DIFF
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -9,6 +9,8 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "0.1.9"
+libloading = { version = "0.6.5", optional = true }
+lazy_static = { version = "1.4.0", optional = true }
 neon-sys = { version = "=0.5.3", path = "../neon-sys", optional = true }
 nodejs-sys = { version = "0.7.0", optional = true }
 smallvec = "1.4.2"
@@ -16,6 +18,7 @@ smallvec = "1.4.2"
 [features]
 default = []
 docs-only = ["neon-sys/docs-only"]
+napi = ["nodejs-sys", "lazy_static", "libloading"]
 
 [package.metadata.docs.rs]
 features = ["docs-only"]
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -10,7 +10,6 @@ edition = "2018"
 [dependencies]
 cfg-if = "0.1.9"
 libloading = { version = "0.6.5", optional = true }
-lazy_static = { version = "1.4.0", optional = true }
 neon-sys = { version = "=0.5.3", path = "../neon-sys", optional = true }
 nodejs-sys = { version = "0.7.0", optional = true }
 smallvec = "1.4.2"
@@ -18,7 +17,7 @@ smallvec = "1.4.2"
 [features]
 default = []
 docs-only = ["neon-sys/docs-only"]
-napi = ["nodejs-sys", "lazy_static", "libloading"]
+napi = ["nodejs-sys", "libloading"]
 
 [package.metadata.docs.rs]
 features = ["docs-only"]
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Dave Herman <david.herman@gmail.com>"]
 description = "Bindings to the Node.js native addon API, used by the Neon implementation."
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
+edition = "2018"
 
 [dependencies]
 cfg-if = "0.1.9"
--- a/crates/neon-runtime/src/napi/bindings.rs
+++ b/crates/neon-runtime/src/napi/bindings.rs
@@ -1,0 +1,144 @@
+use std::mem::MaybeUninit;
+use lazy_static::lazy_static;
+use libloading::{Library, Symbol};
+
+/* Later we should do:
+#[repr(C)]
+struct NapiEnvStruct {}
+
+#[repr(C)]
+struct NapiValueStruct {}
+
+pub(crate) type NapiEnv = *mut NapiEnvStruct;
+pub(crate) type NapiValue = *mut NapiValueStruct;
+
+But in this sample we still rely on nodejs_sys's types
+*/
+
+pub(crate) type NapiEnv = nodejs_sys::napi_env;
+pub(crate) type NapiValue = nodejs_sys::napi_value;
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+#[allow(dead_code)]
+pub(crate) enum NapiStatus {
+    napi_ok,
+    napi_invalid_arg,
+    napi_object_expected,
+    napi_string_expected,
+    napi_name_expected,
+    napi_function_expected,
+    napi_number_expected,
+    napi_boolean_expected,
+    napi_array_expected,
+    napi_generic_failure,
+    napi_pending_exception,
+    napi_cancelled,
+    napi_escape_called_twice,
+    napi_handle_scope_mismatch,
+    napi_callback_scope_mismatch,
+    napi_queue_full,
+    napi_closing,
+    napi_bigint_expected,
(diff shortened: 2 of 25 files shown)


crates/neon-runtime/Cargo.toml
```diff
@@ -5,16 +5,20 @@ authors = ["Dave Herman <david.herman@gmail.com>"]
 description = "Bindings to the Node.js native addon API, used by the Neon implementation."
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
+edition = "2018"
```
goto-bus-stop, line 8: Updated to edition 2018 because I only know how to use `macro_rules!` in this edition

crates/neon-runtime/src/napi/bindings.rs
```diff
@@ -0,0 +1,119 @@
...
+*/
+
+pub(crate) type NapiEnv = nodejs_sys::napi_env;
+pub(crate) type NapiValue = nodejs_sys::napi_value;
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+#[allow(dead_code)]
+pub(crate) enum NapiStatus {
```
goto-bus-stop, line 25: Copy-paste from the N-API docs; with `repr(C)` this is fine. N-API will only ever add new values at the end

crates/neon-runtime/src/napi/bindings.rs
```diff
@@ -0,0 +1,119 @@
...
+    napi_closing,
+    napi_bigint_expected,
+    napi_date_expected,
+    napi_arraybuffer_expected,
+    napi_detachable_arraybuffer_expected,
+    napi_would_deadlock,
+}
+
+pub(crate) struct Napi<'a> {
+    pub napi_get_undefined: Symbol<'a, unsafe extern "C" fn(env: NapiEnv, out: *mut NapiValue) -> NapiStatus>,
```
goto-bus-stop, line 51: I think the actual declaration is quite hard to read here, because of the `Symbol<>` wrapper and because it's in the middle of the line instead of at the start. Could we make it simpler somehow?

kjvalencik:
@goto-bus-stop This is really awesome! I think you're right that it would need to be a proc-macro because of the function arguments; I don't think there's a reasonable way to match all the variations there. But, I think a proc-macro might make this a bit too complicated. We could potentially write some python or node.js to generate it offline.

Overall, I really like this strategy. I think it's worth writing the trampoline functions to eliminate the `napi!` macro and hopefully end up with a fast path. Here's a potential option:

```rust
fn must_initialize() -> ! {
    panic!("Must initialize N-API calls");
}

unsafe extern "C" fn _napi_get_undefined(
    _env: NapiEnv,
    _out: *mut NapiValue,
) -> NapiStatus {
    must_initialize()
}

static mut NAPI_GET_UNDEFINED: unsafe extern "C" fn(
    NapiEnv,
    *mut NapiValue,
) -> NapiStatus = _napi_get_undefined;

#[inline]
pub(crate) unsafe fn napi_get_undefined(
    env: NapiEnv,
    out: *mut NapiValue,
) -> NapiStatus {
    NAPI_GET_UNDEFINED(env, out)
}

unsafe extern "C" fn _napi_create_double(
    _env: NapiEnv,
    _value: f64,
    _out: *mut NapiValue,
) -> NapiStatus {
    must_initialize()
}

static mut NAPI_CREATE_DOUBLE: unsafe extern "C" fn(
    NapiEnv,
    f64,
    *mut NapiValue,
) -> NapiStatus = _napi_create_double;

#[inline]
pub(crate) unsafe fn napi_create_double(
    env: NapiEnv,
    value: f64,
    out: *mut NapiValue,
) -> NapiStatus {
    NAPI_CREATE_DOUBLE(env, value, out)
}
```

It's very verbose, but fairly straightforward. Each function has 3 copies of the signature:

* Empty initialization prefixed with `_` (e.g., `_napi_create_double`) that panics
* Static `fn` pointer (e.g., `NAPI_CREATE_DOUBLE`) initialized with the panicking default
* Trampoline `fn` that executes the static. This keeps the call sites simple. These are marked inline for performance. Running dynamically loaded `n-api` methods should be only a single extra pointer lookup.

Unfortunately, the never type (`!`) isn't stabilized so we might need a distinct panic in each. If this blows up the binary size, there's some super unsafe `transmute` magic that can be used to eliminate it.

With the trampoline functions, I don't think we need `lazy_static`. Instead, we can `leak` the `Library`.

goto-bus-stop:
(FWIW the approach in this PR also compiles to a single indirect call in release mode)

goto-bus-stop:
I agree that a proc macro would be too complicated, I guess I just wanted to explore that a bit after the recent `#[neon::main]` addition 😄 

I do like the simplicity of that function pointer approach. Having to also duplicate the arguments list is a bit unfortunate. Maybe it's fine since the compiler will yell at you if they are out of sync, and it should be one of the least-modified parts of the codebase.

The lazy_static isn't strictly necessary for this PR either. I figured that the cost of lazy_static is irrelevant since it's only paid once during initialization. But there is also a cognitive cost to having the lifetimes on the `Napi` struct where really that could just be 'static…

goto-bus-stop:
Last commit removes the `napi!()` macro in favour of manually written trampoline functions. I don't hate it? It definitely is quite verbose…

goto-bus-stop:
https://github.com/neon-bindings/neon/commit/6ce7cf0ba8803cbd1c09fb95ae90cbf925102aab removes the `Symbol<>` wrapping and use of `lazy_static!`. I could go either way on that, I guess

kjvalencik:
I gave it a try. Lots and lots of typing. 😆 

https://github.com/neon-bindings/neon/pull/646

PR #646, opened 2020-11-30
Dynamic Loading

kjvalencik:
As described in https://github.com/neon-bindings/neon/issues/584 and tested in experimented with in https://github.com/neon-bindings/neon/pull/645.

Follow-ups:
* Refactor `neon-build`
* Remove delay load hook
* Convert electron test to N-API

Closes #645

### DIFF
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -9,13 +9,16 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "0.1.9"
+libloading = { version = "0.6.5", optional = true }
 neon-sys = { version = "=0.5.3", path = "../neon-sys", optional = true }
-nodejs-sys = { version = "0.7.0", optional = true }
 smallvec = "1.4.2"
+
+[dev-dependencies]
+nodejs-sys = "0.7.0" # Dev dependency for easy copying
 
 [features]
 default = []
-napi = ["nodejs-sys"]
+napi = ["libloading"]
 docs-only = ["neon-sys/docs-only"]
 
 [package.metadata.docs.rs]
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -15,6 +15,7 @@ smallvec = "1.4.2"
 
 [features]
 default = []
+napi = ["nodejs-sys"]
 docs-only = ["neon-sys/docs-only"]
 
 [package.metadata.docs.rs]
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -14,7 +14,7 @@ neon-sys = { version = "=0.5.3", path = "../neon-sys", optional = true }
 smallvec = "1.4.2"
 
 [dev-dependencies]
-nodejs-sys = "0.7.0" # Dev dependency for easy copying
+nodejs-sys = "0.7.0" # Not strictly needed; just here for easy manual copying
 
 [features]
 default = []
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Dave Herman <david.herman@gmail.com>"]
 description = "Bindings to the Node.js native addon API, used by the Neon implementation."
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
+edition = "2018"
 
 [dependencies]
 cfg-if = "0.1.9"
--- a/crates/neon-runtime/src/napi/array.rs
+++ b/crates/neon-runtime/src/napi/array.rs
@@ -1,6 +1,6 @@
 //! Facilities for working with Array `napi_value`s.
 
-use raw::{Env, Local};
+use crate::raw::{Env, Local};
 
 use nodejs_sys as napi;
 
--- a/crates/neon-runtime/src/napi/array.rs
+++ b/crates/neon-runtime/src/napi/array.rs
@@ -2,12 +2,12 @@
 
 use crate::raw::{Env, Local};
 
-use nodejs_sys as napi;
+use crate::napi::bindings as napi;
 
 pub unsafe extern "C" fn new(out: &mut Local, env: Env, length: u32) {
     assert_eq!(
-        napi::napi_create_array_with_length(env, length as usize, out as *mut _),
-        napi::napi_status::napi_ok,
+        napi::create_array_with_length(env, length as usize, out as *mut _),
+        napi::Status::Ok,
     );
 }
 
@@ -19,8 +19,8 @@ pub unsafe extern "C" fn new(out: &mut Local, env: Env, length: u32) {
 pub unsafe extern "C" fn len(env: Env, array: Local) -> u32 {
     let mut len = 0;
     assert_eq!(
-        napi::napi_get_array_length(env, array, &mut len as *mut _),
-        napi::napi_status::napi_ok
+        napi::get_array_length(env, array, &mut len as *mut _),
+        napi::Status::Ok
     );
     len
 }
--- a/crates/neon-runtime/src/napi/bindings/mod.rs
+++ b/crates/neon-runtime/src/napi/bindings/mod.rs
@@ -1,12 +1,105 @@
+//! # FFI bindings to N-API symbols
+//!
+//! These types are manually copied from bindings generated from `bindgen`. To
+//! update, use the following approach:
+//!
+//! * Run `cargo test --manifest-path=crates/neon-runtime/Cargo.toml --features napi`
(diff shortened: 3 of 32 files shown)


crates/neon-runtime/src/napi/bindings/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    };
+}
+
+mod types;
+mod functions;
+
+pub(crate) use functions::*;
+
+// FIXME: Make this safe!
+pub unsafe fn setup() {
```
goto-bus-stop, line 57: I don't think this _needs_ to be safe since we are the only ones who will ever call it, once during initialization. Is there a way that this would be called more than once?

kjvalencik: I think using the module in multiple contexts might.

kjvalencik: [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) is exactly what I needed. The description even includes:
> Useful for one-time initialization for **FFI** or related functionality.

```rust
static SETUP: Once = Once::new();

pub fn setup() {
    SETUP.call_once(|| unsafe {
        load();
    });
}
```

This is completely safe. It can be called multiple times from multiple threads. The closure will only execute once and if it's called concurrently, the other calls will *block* until it has completed.

goto-bus-stop:
this is looking fantastic 🥰 

We talked about having new N-API versions behind feature flags so we can introduce them in minor releases. How could we support that in this approach?

I don't know if it's feasible to do this with `macro_rules!`:
```rust
#[cfg(feature = "napi-7")]
fn detach_arraybuffer(...) -> Status;
```

If not, perhaps we could generate separate structs for each version.
```rust
mod napi1 {
    generate!(extern "C" {
        fn create_undefined(...) -> Status;
    });
}
mod napi7 {
    generate!(extern "C" {
        fn detach_arraybuffer(...) -> Status;
    });
}

pub fn setup() {
    // wrapped in ONCE.call_once() stuff
    napi1::load().expect("Could not load N-API functions");
    #[cfg(feature = "napi-7")]
    {
        napi7::load().expect("Could not load N-API version 7 functions");
    }
}
```

kjvalencik:
> I don't know if it's feasible to do this with `macro_rules!`:
> 
> ```rust
> #[cfg(feature = "napi-7")]
> fn detach_arraybuffer(...) -> Status;
> ```

I did some reading and learned a couple new things about macros. Yes, this is possible! The `meta` type can be used for capturing attribute contents.

Additionally, doc comments are syntactic sugar for doc attributes. It will catch them as well, so we can add doc comments to these functions.

Are there any functions that we need to feature gate already?

goto-bus-stop:
I think the most recent addition thta we use is napi_get_all_property_names, which was added in N-API 6 in early 2020. So not strictly necessary to gate, but it could be a way to ensure that it all works at least.

Another more useful option could be the type tagging stuff that is not yet in a numbered N-API version, but is released in Node.js 12.x and 14.x. But that would probably best be done in a separate PR that also starts to make use of it in JsBox.

kjvalencik:
@goto-bus-stop After thinking about it a bit more, I think your module suggestion is cleaner. Using attributes, would require making assumptions about those attributes--that they are specifically for conditional compiling--because they would need to be applied in three different places.

Putting them in versioned modules is nice and simple.

crates/neon-runtime/src/napi/bindings/mod.rs
```diff
@@ -0,0 +1,80 @@
+macro_rules! napi_name {
```
kjvalencik, line 1: Sample macro output:

```rust
#[cfg(not(windows))]
use libloading::os::unix::Library;
use std::os::raw::{c_char, c_void};
use super::types::*;

pub(crate) struct Napi {
    get_undefined: unsafe extern "C" fn(env: Env, result: *mut Value) -> Status,
    /* ... repeat for each N-API function */
}

pub(crate) unsafe fn load() -> Result<(), libloading::Error> {
    let host = Library::this();

    NAPI = Napi {
        get_undefined: *host.get("napi_get_undefined".as_bytes())?,
        /* ... repeat for each N-API function */
    };

    Ok(())
}

#[inline]
pub(crate) unsafe fn get_undefined(env: Env, result: *mut Value) -> Status {
    (NAPI.get_undefined)(env, result)
}

fn panic_load<T>() -> T {
    panic!("Must load N-API bindings")
}

static mut NAPI: Napi = {
    unsafe extern "C" fn get_undefined(_: Env, _: *mut Value) -> Status {
        panic_load()
    }
    /* ... repeat for each N-API function */

    Napi {
        get_undefined,
        /* ... repeat for each N-API function */
    }
};
```

crates/neon-runtime/Cargo.toml
```diff
@@ -5,15 +5,20 @@ authors = ["Dave Herman <david.herman@gmail.com>"]
...
 
 [dependencies]
 cfg-if = "0.1.9"
+libloading = { version = "0.6.5", optional = true }
 neon-sys = { version = "=0.5.3", path = "../neon-sys", optional = true }
-nodejs-sys = { version = "0.7.0", optional = true }
 smallvec = "1.4.2"
 
+[dev-dependencies]
+nodejs-sys = "0.7.0" # Dev dependency for easy copying
```
dherman, line 17: Maybe be even more explicit in the comment, like "Not strictly needed; just here for easy manual copying"

crates/neon-runtime/src/napi/array.rs
```diff
@@ -1,13 +1,13 @@
...
+use crate::raw::{Env, Local};
 
-use nodejs_sys as napi;
+use crate::napi::bindings as napi;
 
 pub unsafe extern "C" fn new(out: &mut Local, env: Env, length: u32) {
     assert_eq!(
-        napi::napi_create_array_with_length(env, length as usize, out as *mut _),
-        napi::napi_status::napi_ok,
+        napi::create_array_with_length(env, length as usize, out as *mut _),
```
dherman, line 9: These naming conventions aren't, like, gonna change the world, but they're so nice 😎

crates/neon-runtime/src/napi/bindings/mod.rs
kjvalencik: Need to add this as a comment to more clearly document the output of the macro.

dherman: This example is so helpful. Can you put it into a comment directly in the source file, maybe with a few sentences explaining the overall approach?

crates/neon-runtime/src/napi/bindings/types.rs
```diff
@@ -0,0 +1,154 @@
+use std::ffi::c_void;
+
```
dherman, line 2: I think this file should have a comment at the top explaining the manual process we use for extracting the type declarations from nodejs_sys and pasting them here.

kjvalencik:
@goto-bus-stop I think we can defer the conditional compiling to a different PR since it's not strictly necessary and we still need to iron out the ergonomics of the feature flags. I'm thinking something like:

```toml
# neon/Cargo.toml
[features]
# Default N-API version
napi = ["napi6"]
napi6 = ["napi-runtime/napi6"]
napi7 = ["napi-runtime/napi7"]
napi_experimental = ["napi-runtime/napi-experimental"]

# neon-runtime/Cargo.toml
[features]
napi = ["libloading"]
napi6 = ["napi"]
napi7 = ["napi6"]
napi_experimental = ["napi7"]
```

dherman (changes requested):
This is too exciting. And I can't believe how much easier it was to understand than I expected. My only comments were really some suggestions for comments explaining things maintainers will want to know down the line.

crates/neon-runtime/Cargo.toml
kjvalencik: Done.

crates/neon-runtime/src/napi/bindings/types.rs
kjvalencik: I added this to `bindings/mod.rs` instead because the approach is identical in `types.rs` and `functions.rs`.

crates/neon-runtime/src/napi/array.rs
kjvalencik: What's not to love about repeating the word `napi` 3x in a single identifier. 😆  `s/napi::napi_status::napi_ok/napi::Status::Ok/`

napi, napi.... napi.

...napi

goto-bus-stop:
> @goto-bus-stop I think we can defer the conditional compiling to a different PR since it's not strictly necessary and we still need to iron out the ergonomics of the feature flags. What do you think?

Yeah, if we are OK with the module approach then there is no need to figure the details out now :+1:

dherman (approved):
Looks great! :shipit:

PR #647, opened 2020-12-04
`neon-build` refactor to make use of dynamic loading

kjvalencik:
The PR is a little large, but includes several related commits:

* Re-organize `neon-build`. This commit introduces no functional changes. It re-organizes `napi` and `neon-sys` building so they are entirely separate.
* Re-write the `napi` build to make use of dynamic loading. It only needs to output the `index.node` file
* Update NAPI tests to use a flattened structure to more accurately reflect the pattern outlined in https://github.com/neon-bindings/rfcs/pull/36
* Remove the win delay load hook. This is a little sad because it's very neat, but is also no longer necessary. Let me know if you would like to leave it in for future reference.
* Update electron tests to use NAPI. This was important for testing that win delay load hook is no longer necessary.
* Add a `neon_build::Builder` for customizing the `index.node` location


test/electron/test/index.js
```diff
@@ -5,24 +5,18 @@ const path = require('path');
 
 const electron = require('electron');
 const { Application } = require('spectron');
+const { window } = require('globalthis/implementation');
```
kjvalencik, line 8: Delete this.

kjvalencik: Nevermind. This is required.

crates/neon-build/src/napi/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+///         .output_dir("native.node")
+///         .setup();
+/// }
+#[derive(Debug, Default)]
+pub struct Builder {
+    output_dir: Option<PathBuf>,
+    output_file: Option<PathBuf>,
+}
+
+impl Builder {
```
dherman, line 58: I think `Builder` is a confusing name since it doesn't describe what it does -- and it also could be confusing because this crate is `neon_build` but `Builder` isn't about running a Neon build.

We talked about `Setup::options()` being the way to construct this thing. I think it reads nicely.

dherman (changes requested):
Looks great! I just suggest we rename the `Builder` API.

dherman (approved):
:shipit:

PR #649, opened 2020-12-09
feat(neon-runtime): Add selectable N-API versions

kjvalencik:
Proposed feature flags for selecting a target N-API version.

Ignore the indentation change on symbol definitions: https://github.com/neon-bindings/neon/pull/649/files?w=1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,20 @@ default-panic-hook = []
 # enabled by default.
 legacy-runtime = ["neon-runtime/neon-sys", "neon-build/neon-sys"]
 
-# Feature flag to enable the experimental N-API runtime. For now, this feature
+# Default N-API version. Prefer to select a minimum required version.
+napi-runtime = ["napi-6"]
+
+# Select the N-API version
+# Feature flags to enable the experimental N-API runtime. For now, this feature
 # is disabled by default.
-napi-runtime = ["proc-macros", "neon-macros/napi", "neon-runtime/napi"]
+napi-1 = ["proc-macros", "neon-macros/napi", "neon-runtime/napi"]
+napi-2 = ["napi-1", "neon-runtime/napi-2"]
+napi-3 = ["napi-2", "neon-runtime/napi-3"]
+napi-4 = ["napi-3", "neon-runtime/napi-4"]
+napi-5 = ["napi-4", "neon-runtime/napi-5"]
+napi-6 = ["napi-5", "neon-runtime/napi-6"]
+napi-latest = ["napi-6"]
+napi-experimental = ["napi-6", "neon-runtime/napi-experimental"]
 
 # Feature flag to disable external dependencies on docs build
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -42,9 +42,20 @@ default-panic-hook = []
 # enabled by default.
 legacy-runtime = ["neon-runtime/neon-sys", "neon-build/neon-sys"]
 
-# Feature flag to enable the experimental N-API runtime. For now, this feature
+# Default N-API version. Prefer to select a minimum required version.
```
dherman, line 45: Would this be a good place to link to definitive documentation on the N-API version numbers and what they included?

crates/neon-runtime/src/napi/bindings/mod.rs
```diff
@@ -77,12 +77,17 @@ macro_rules! napi_name {
...
-///     // On Windows, the host process might not be a library
-///     let host = host?;
+/// pub(crate) unsafe fn load(
+///     host: &libloading::Library,
+///     actual_napi_version: u32,
+///     expected_napi_version: u32,
+/// ) -> Result<(), libloading::Error> {
+///     assert!(
+///         actual_napi_version >= expected_napi_version,
+///         "Minimum required N-API Version {}, found {}.",
```
dherman, line 87: Lowercase “version” instead of “Version”?

crates/neon-runtime/src/napi/object.rs
```diff
@@ -8,6 +8,7 @@ pub unsafe extern "C" fn new(out: &mut Local, env: Env) {
     napi::create_object(env, out as *mut _);
 }
 
+#[cfg(feature = "napi-6")]
```
dherman, line 11: My only concern with all these changes is whether it’ll make it intimidating for maintenance, since it’s not obvious what N-API version any given API was introduced in. Is it documented directly in the standard N-API docs in a way that’s easy to discover? Maybe we could explain somewhere in a comment how to know which `napi-<N>` flag to use when adding an API?

dherman (changes requested):
I’m supportive, just a concern about how contributors/maintainers can know what NAPI version any given API comes from. Maybe just a matter of adding a comment. Also one tiny suggested copy edit in an error message.

crates/neon-runtime/src/napi/object.rs
kjvalencik: I left a [comment](https://github.com/neon-bindings/neon/pull/649/files#diff-bf24afa304665f35997079e36185410c84b2a375828d0399f592dce91767fa47R1-R4) in the `functions.rs` file to consult the N-API documentation for version numbers. This should make it very clear when adding a new N-API method.

However, that does not fully address this concern since we only test the latest N-API version. I filed a follow-up issue to fix this in CI: https://github.com/neon-bindings/neon/issues/659

dherman (approved):
:shipit:

PR #655, opened 2020-12-22
feat(cli): Add `--crate` option to specify the Rust crate

demurgos:
This commit adds a new option to the `build` and `clean` commands: `--crate`. This option allows to manually configure the directory containing the Rust crate.

By default, the CLI always looks into the `native` sub-directory to find the Rust crate. Using this flag, you can easily change this directory. In particular, it allows to place the `Cargo.toml` next to the `package.json` and use `neon build --path . --crate .`.

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -167,6 +167,7 @@ const spec: Spec = {
 
   build: {
     args: [{ name: "release", alias: "r", type: Boolean },
+           { name: "crate", alias: "c", type: String, defaultValue: "native" },
            { name: "path", alias: "p", type: Boolean },
            { name: "modules", type: String, multiple: true, defaultOption: true },
            { name: "help", alias: "h", type: Boolean }],
@@ -185,6 +186,11 @@ const spec: Spec = {
         type: Boolean,
         description: "Release build."
       }, {
+        name: "crate",
+        alias: "c",
+        type: String,
+        description: "Specify the path of the Rust crate."
+      }, {
         name: "path",
         alias: "p",
         type: Boolean,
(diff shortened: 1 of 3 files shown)


kjvalencik:
@demurgos Thanks for your contribution! However, I did want to let you know that `neon-cli` will be deprecated when the `n-api` backend is released. Due to this, extending the functionality is not a high priority.

kjvalencik:
FYI, a workaround to this missing feature is using the CLI as a library:

```js
"use strict";

const { default: Project } = require("neon-cli/lib/project");

const TOOLCHAIN = "default";
const RELEASE_FLAG = "--release";

const argv = process.argv.slice(2);
const isRelease = argv.includes(RELEASE_FLAG);
const root = __dirname;

process.on("unhandledRejection", (err) => {
  throw err;
});

(async () => {
  const project = await Project.create(root, {
    // Joined with `root` to find the crate. Overrides the default of `native`.
    crate: "",
  });

  await project.build(TOOLCHAIN, isRelease, []);
})();
```

demurgos:
Hi,
I was using the CLI as a library in other projects but it requires more maintenance. I was configuring another project today and decided to finally submit a PR for this feature.

It's not a big feature so I don't mind if you don't merge because the CLI is being deprecated. I hope that when the `n-api` backend gets released there will be a way to configure the crate path.

kjvalencik:
@demurgos One of the great things about N-API is it doesn't require any special build logic. `cargo build` is sufficient. The only additional step is to copy the build library from `target/` to `[filename].node`.

We currently have a `neon-build` build script that will output directly to `[filename].node` but, it breaks cargo caching causing the file to re-link on every `cargo build`. That might not be acceptable in all use cases, so I'm also proposing a `cargo` wrapper for the copy.

But, really, it's _just_ a copy and could easily be deferred to some other build tool like `package.json`.

demurgos:
Oh, I get it. I enabled the `napi-runtime` and only had to run `cargo build` (loading the module failed, but that's another problem). Thank you very much. I'll close the issue then.

PR #661, opened 2021-01-06
Update documentation URL

Himself65:


README.md
```diff
@@ -21,7 +21,7 @@ Then see the [Hello World guide](https://neon-bindings.com/docs/hello-world/) fo
 
 # Docs
 
-See our [Neon fundamentals docs](https://neon-bindings.com/docs/primitives) and our [API docs](https://neon-bindings.com/api/neon/).
+See our [Neon fundamentals docs](https://neon-bindings.com/docs/primitives) and our [API docs](https://docs.rs/neon/0.7.0/neon).
```
kjvalencik, line 24: Thanks for fixing this link! Can you use the `latest` tag instead of a specific version? Also, it may be better to link to `intro` instead of `primitives`.

```suggestion
See our [Neon fundamentals docs](https://neon-bindings.com/docs/intro) and our [API docs](https://docs.rs/neon/latest/neon).
```

kjvalencik:
Thanks for the fix!

PR #666, opened 2021-01-10
N-API: deprecate PartialEq and add strict_equals method

dherman:
Deprecates `PartialEq` for the N-API backend and adds a `strict_equals` method for handles.

### DIFF
--- a/crates/neon-runtime/src/napi/mem.rs
+++ b/crates/neon-runtime/src/napi/mem.rs
@@ -1,10 +1,6 @@
 use crate::raw::{Env, Local};
 use crate::napi::bindings as napi;
 
-pub unsafe extern "C" fn same_handle(_lhs: Local, _rhs: Local) -> bool {
-    panic!("PartialEq is deprecated with N-API backend");
-}
-
 pub unsafe extern "C" fn strict_equals(env: Env, lhs: Local, rhs: Local) -> bool {
     let mut result = false;
     assert_eq!(napi::strict_equals(env, lhs, rhs, &mut result as *mut _), napi::Status::Ok);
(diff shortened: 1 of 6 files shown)


crates/neon-runtime/src/napi/mem.rs
```diff
@@ -1,3 +1,12 @@
-use crate::raw::Local;
+use crate::raw::{Env, Local};
+use crate::napi::bindings as napi;
 
-pub unsafe extern "C" fn same_handle(_h1: Local, _h2: Local) -> bool { unimplemented!() }
+pub unsafe extern "C" fn same_handle(_lhs: Local, _rhs: Local) -> bool {
+    panic!("PartialEq is deprecated with N-API backend");
```
kjvalencik, line 5: Can we move this method? We removed everything else that we couldn't or weren't going to implement.

src/handle/mod.rs
```diff
@@ -181,6 +181,12 @@ impl<'a, T: Value> Handle<'a, T> {
         self.downcast(cx).or_throw(cx)
     }
 
+    #[cfg(feature = "napi-1")]
+    pub fn strict_equals<'b, U: Value, C: Context<'b>>(&self, cx: &mut C, other: Handle<'b, U>) -> bool {
```
kjvalencik, line 185: Why is this on `Handle`? Shouldn't this be on `trait Value`?

kjvalencik (commented):
@dherman I really like this! I have a couple of small questions. Thanks!

src/handle/mod.rs
dherman: Discussed offline, and we came to this thought:
- It's nice to have it by default without having to import `Value`
- There's not much on `Value`
- Maybe in the future we could consider adding most/all of the `Value` methods on `Handle` and vice versa, so there isn't such a confusing split. (In an ambiguity, rustc will always let the inherent methods take precedence, so as long as the methods are consistent, there is no harm in the duplication.)

kjvalencik (approved):

dherman:
Note: this was the last remaining item in #444! 🎉

PR #673, opened 2021-01-22
CI: Add Windows Arm64 target

EwoutH:
Adds the `aarch64-pc-windows-msvc` target to the build matrix of the Windows job.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 15.x]
+        node-version: [12, 14, 15]
         rust-toolchain: [stable, beta]
         rust-target: [x86_64-pc-windows-msvc, aarch64-pc-windows-msvc]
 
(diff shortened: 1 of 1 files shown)


kjvalencik:
@EwoutH Thanks for the contribution!

Did you encounter issues building for Windows Arm64? Our build matrix is already kind of large and if we haven't experienced a platform specific regression, we might want to hold off adding this.

dennisameling:
Just wanted to chime in here to mention that Neon is working correctly on Windows arm64 with the N-API backend! 🚀  https://github.com/signalapp/ringrtc/pull/12

EwoutH:
Rebased and tuned the Rust install a bit. All Arm64 (Aarch64) builds are still failing, I will investigate a bit more later this week.

.github/workflows/windows.yml
```diff
@@ -15,23 +15,27 @@ jobs:
     runs-on: windows-latest
 
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
```
kjvalencik, line 19: FYI, Neon follows LTS and will support Node 10 until it goes out of maintenance 4/30.

I support removing it from the build matrix, but we may want to defer until after or closer to the date.

EwoutH: Good point. Maybe we could go from a full factorial matrix to a more sparse configuration. Not every node version would have to be tested on every Rust toolchain and every target.

kjvalencik: I would like to do that for the ARM builds. We can do a smaller matrix on PRs and the full matrix on merge to main.

PR #677, opened 2021-01-29
fix(neon): Do not panic when dropping a Root if the VM has shutdown

kjvalencik:
Currently, if a `Root` is inside an `EventQueue` task and the VM is stopped, the user will never have an opportunity to clean up the `Root`. In this case, it's not a leak because the value will be deleted as the VM exits.

This PR introduces a new `IS_RUNNING` thread local value. It is initialized as `true` during module initialization and set to `false` the first time an `EventQueue` callback receives a `None` for `Env`.

If the value is `false`, the `Drop` impl on `Root` will _not_ panic.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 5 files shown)


dherman (approved):
:shipit:

PR #680, opened 2021-02-09
Make Option<T: Finalize> implement Finalize

jrose-signal:
Ran into this recently, though admittedly I was using Finalize outside of its primary purpose of JsBox. I considered whether Result should implement Finalize as well, but decided that was a bit trickier.

Are there any tests I should add here?

### DIFF
--- a/src/types/boxed.rs
+++ b/src/types/boxed.rs
@@ -370,11 +370,19 @@ impl<T: Finalize> Finalize for Vec<T> {
     }
 }
 
-// Smart Pointers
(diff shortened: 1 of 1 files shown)


kjvalencik:
Thanks for the contribution! This code is pretty trivial and doesn't require a test. I agree that `Result` would be trickier because either type (or both) might be `Finalize` and we would end up with overlapping `impl` trying to handle them.

We would only be able to provide the most strict where _both_ `<T: Finalize, E: Finalize>`.

PR #681, opened 2021-02-10
Fix as_slice and as_mut_slice for empty Buffers and ArrayBuffers

jrose-signal:
These buffers may have `NULL` as their address, which `slice::from_raw_parts` does not allow.

### DIFF
--- a/src/types/binary.rs
+++ b/src/types/binary.rs
@@ -138,9 +138,13 @@ impl<'a> BinaryData<'a> {
     /// # }
     /// ```
     pub fn as_slice<T: BinaryViewType>(self) -> &'a [T] {
-        let base = unsafe { mem::transmute(self.base) };
-        let len = self.size / mem::size_of::<T>();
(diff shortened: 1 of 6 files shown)


kjvalencik:
@jrose-signal In what situations will Node have a `null` pointer for the backing buffer? I rolled back the changes to `src/types/binary.rs` and the tests still passed on `v14.15.4`.

kjvalencik:
Hmm, I guess it doesn't matter if we can reproduce or not. In either case it's clearly relying on UB. From the N-API docs:

> If length is 0, this may be NULL or any other pointer value.

kjvalencik (approved):

jrose-signal:
The newly-added tests do fail for me with Node v12.13.0 (at least, that's what I have as `which node`). I suspect there's also an optimization aspect to this; immediately `to_vec()` on an empty buffer's slice worked fine for me but doing more complicated work caused problems.

PR #685, opened 2021-02-15
N-API migration guide

dherman:
Wrote up a complete migration guide with all of the incompatible changes between the legacy backend and N-API backend.

[Rendered](https://github.com/neon-bindings/neon/blob/4fda49d283693fa60af8e0606f3a552d01c66c01/MIGRATION_GUIDE.md)

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -1,5 +1,28 @@
 # N-API Migration Guide
 
+## What is this about?
+
+Since v10, Node.js supports an improved API for building native modules, known as [N-API](https://nodejs.org/api/n-api.html). N-API offers a clearer, more complete, and more stable API layer for writing Node.js plugins than previous Node versions did.
+
+The Neon community has been [hard at work](https://github.com/neon-bindings/neon/issues/444) porting the library to a new backend based on N-API.
+
+### Why N-API?
+
+Some key benefits of the new backend include:
+- Compiled Neon modules will work in all versions of Node _without needing to be recompiled_, guaranteed!
+- You can precompile Neon-based libraries to be completely transparent to downstream consumers.
+- The build process is streamlined, making Neon apps more reliable and easier to debug.
+- The stability guarantees of N-API allow us to avoid risk of incompatible changes to future releases of Neon.
+
+### What does this mean for me?
+
+Porting Neon to N-API has been mostly transparent, but it has required a few backwards-incompatible changes. This guide provides instructions on how to migrate existing apps to the new N-API backend.
+
+Fortunately, the guaranteed stability of N-API means that once Neon users do this migration, we have increased confidence in the stability of Neon. We expect this to be the **last major breaking change before reaching Neon 1.0.**
+
+If you have any trouble porting, **please reach out to us** with a Neon issue or on the community Slack! We want to help everyone upgrade as smoothly and seamlessly as possible.
+
+
 ## Getting started
 
 ### Supported Node versions
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -108,7 +108,7 @@ The N-API backend introduces two categories of significant change:
 
 The `declare_types!` macro is deprecated and replaced by the `JsBox` type.
 
-_Rationale:_ The `declare_types!` macro provides a syntax for defining classes, but requires substantial boilerplate and is unergonomic for simple cases. It's also not flexible enough to express the full range of JavaScript classes syntax and semantics. With the `JsBox` type, it's easy to embed Rust data in JavaScript objects, which can then be nested inside of more feature-rich classes defined in pure JavaScript (or TypeScript).
+_Rationale:_ The `declare_types!` macro provides a syntax for defining classes, but requires substantial boilerplate and is unergonomic for simple cases and tends to interact poorly with IDEs. It's also not flexible enough to express the full range of JavaScript classes syntax and semantics. With the `JsBox` type, it's easy to embed Rust data in JavaScript objects, which can then be nested inside of more feature-rich classes defined in pure JavaScript (or TypeScript).
 
 **Before:**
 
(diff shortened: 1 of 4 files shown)


MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
-# [WIP] N-API Migration Guide
+# N-API Migration Guide
```
amilajack, line 1: Should we name this guide as `1.0 Migration`? There seems to be some things that aren't related to NAPI (from my understanding) like event queue.

dherman: They're all necessary for migrating to the N-API backend, but it's a good idea to be clearer that these changes will be permanent for 1.0.

kjvalencik: Maybe we can leave that note and then rename it the "1.0 Migration Guide" as part of that release?

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
-## Context
+## Getting started
+
+### Supported Node versions
+
+The N-API backend of Neon requires a minimum Node version of 10.0.
+
+### Enabling the N-API backend
+
+To enable the N-API backend, you need to enable the `"napi"` feature flag in your `Cargo.toml`:
```
kjvalencik, line 11: Should the prose also mention`default-features = false`?

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
+The N-API backend introduces two categories of significant change:
+
+1. Embedding Rust data, which is no longer done through the awkward and complex `declare_types!` (i.e. classes) macro, but through a simpler primitive: the `JsBox` API.
+2. Concurrency, which is offered through the Event Queue API instead of the Task API or Event Handlers, both of which are deprecated and removed in the N-API backend.
+
+### Embedding Rust data
+
+The `declare_types!` macro is deprecated and replaced by the `JsBox` type.
+
+_Rationale:_ The `declare_types!` macro provides a syntax for defining classes, but requires substantial boilerplate and is unergonomic for simple cases. It's also not flexible enough to express the full range of JavaScript classes syntax and semantics. With the `JsBox` type, it's easy to embed Rust data in JavaScript objects, which can then be nested inside of more feature-rich classes defined in pure JavaScript (or TypeScript).
```
kjvalencik, line 82: It might also be worth mentioning that IDEs struggle with code inside macros. The `declare_types` macro placed large amounts of code inside a macro reducing IDE ergonomics.

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
+}
+```
+
+**After:**
+
+```rust
+struct User {
+    first_name: String,
+    last_name: String,
+}
```
kjvalencik, line 123: This is missing the `Finalize` impl required by `JsBox`:

```rust
impl Finalize for User {}
```

It would also benefit of explaining why `Finalize` exists and that the empty impl may no longer be necessary after specialization lands in Rust.

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
+    let last_name = cx.argument::<JsString>(1)?.value(&mut cx);
+    Ok(cx.boxed(User { first_name, last_name }))
+}
+
+fn user_full_name(mut cx: FunctionContext) -> JsResult<JsString> {
+    let user = cx.argument::<JsBox<User>>(0)?;
+    let full_name = user.full_name();
+    Ok(cx.string(full_name))
+}
+```
```
kjvalencik, line 136: I think it would be helpful to include the JavaScript glue code as well to make this a fair comparison.

```js
class User {
    constructor(firstName, lastName) {
        self.native = addon.createUser(firstName, lastName);
    }
    
    fullName() {
        return addon.userFullName(this.native);
    }
}
```

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
+
+### Concurrency
+
+The supported mechanism for concurrency is the Event Queue API (`neon::event::EventQueue`). This feature has not yet stabilized, so to use this API, you'll also need to enable the `"event-queue-api"` feature flag as well:
+
+```toml
+[dependencies.neon]
+version = "0.7.1"
+default-features = false
+features = ["napi", "event-queue-api"]
```
kjvalencik, line 146: `napi` isn't a feature flag on `neon`. The previous behavior is `napi-runtime`.

Should we encourage selecting a version? E.g., `napi-1`? Otherwise,`napi-latest` is also available.

MIGRATION_GUIDE.md
```diff
@@ -1,14 +1,264 @@
...
+}
+
+pub fn start_task(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let callback = cx.argument::<JsFunction>(0)?;
+    MyTask.schedule(callback);
+    Ok(cx.undefined())
+}
+```
+
+**After:**
```
kjvalencik, line 188: We might want to include the previous code would execute on the `libuv` thread pool, while this approach requires bringing your own thread pool.

kjvalencik (commented):
This looks fantastic! I'm glad you worked on this! I'm eager for us to write some sugar to make all the "After" examples shorter. 😁

MIGRATION_GUIDE.md
dherman: Let's revisit that as we get close to 1.0 launch.

MIGRATION_GUIDE.md
dherman: Great point, I added some more language around this -- see what you think.

dherman:
@kjvalencik @amilajack I've addressed all your great feedback. Feel free to take another look!

kjvalencik (approved):
This looks great! One last nit that can be left for another PR, but it would be great to link the migration guide in the README.md

PR #690, opened 2021-03-01
create-neon

dherman:
Initial implementation of the `create-neon` CLI, as described in the [Streamlined UX RFC](https://github.com/neon-bindings/rfcs/pull/36).

image


ArturAralin:
Hi @dherman! Look on this PR https://github.com/neon-bindings/create-neon-lib/pull/1. It may be useful for it

create-neon/templates/Cargo.toml.hbs
```diff
@@ -0,0 +1,19 @@
...
+edition = "2018"
+exclude = ["index.node"]
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies.neon]
+version = {{neon.libs.version}}
+default-features = false
+features = ["napi-6"]
```
kjvalencik, line 19: TODO: Should be a variable.

create-neon/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/neon-bindings/neon.git"
+  },
+  "keywords": [
+    "neon"
+  ],
+  "devDependencies": {
```
kjvalencik, line 28: This is missing a dependency on handlebars.

create-neon/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "author": "Dave Herman <david.herman@gmail.com>",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/neon-bindings/neon/issues"
+  },
+  "homepage": "https://github.com/neon-bindings/neon#readme",
+  "bin": "./dist/bin/create-neon.js",
+  "files": [
+    "dist/**/*",
+    "templates/**/*"
```
kjvalencik, line 14: Nit, it would be nice to precompile the templates. It's not necessary because performance of this package isn't critical.

create-neon/src/bin/create-neon.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  let gitignore = (await template('.gitignore.hbs'))(ctx);
+  let manifest = (await template('Cargo.toml.hbs'))(ctx);
+  let readme = (await template('README.md.hbs'))(ctx);
+  let librs = (await template('lib.rs.hbs'))(ctx);
+
+  await mkdir('src');
+
+  await writeFile('.gitignore', gitignore, { flag: 'wx' });
```
kjvalencik, line 44: Why are we overriding flag? More specifically, why with `x`?

create-neon/src/bin/create-neon.ts
```diff
@@ -0,0 +1,52 @@
...
+  await writeFile('Cargo.toml', manifest, { flag: 'wx' });
+  await writeFile('README.md', readme, { flag: 'wx' });
+  await writeFile(path.join('src', 'lib.rs'), librs, { flag: 'wx' });
+
+  console.log(`✨ Created Neon project \`${ctx.project.name}\`. Happy 🦀 hacking! ✨`);
```
kjvalencik, line 49: ❤️

create-neon/src/bin/create-neon.ts
```diff
@@ -0,0 +1,52 @@
...
+  let manifest = (await template('Cargo.toml.hbs'))(ctx);
+  let readme = (await template('README.md.hbs'))(ctx);
+  let librs = (await template('lib.rs.hbs'))(ctx);
+
+  await mkdir('src');
+
+  await writeFile('.gitignore', gitignore, { flag: 'wx' });
+  await writeFile('Cargo.toml', manifest, { flag: 'wx' });
+  await writeFile('README.md', readme, { flag: 'wx' });
+  await writeFile(path.join('src', 'lib.rs'), librs, { flag: 'wx' });
```
kjvalencik, line 47: Does `writeFile` overwrite by default? We might want to be careful not clobber files if they already exist. I think it would be better to abort than to potentially destroy user data.

create-neon/src/template.ts
```diff
@@ -0,0 +1,12 @@
...
+import handlebars from 'handlebars';
+import * as path from 'path';
+
+const TEMPLATES_DIR = path.join(__dirname, '..', 'data', 'templates');
+
+export default async function template(filename: string) {
+  let source = await readFile(path.join(TEMPLATES_DIR, filename), {
+    encoding: 'utf8'
+  });
+  return handlebars.compile(source, { noEscape: true });
```
kjvalencik, line 11: Not necessary, but it would be nice to precompile these and move handlebars to a `devDependency`.

create-neon/src/bin/create-neon.ts
dherman: Copy pasta, I’ll drop that

create-neon/src/template.ts
dherman: I don't think we can remove the dependency -- precompiling still results in a runtime dependency on handlebars. I'll still explore precompiling but I'm not sure there ends up being much payoff.

kjvalencik: It's been so long since I've used handlebars, I didn't recall there being a runtime. Yea, I guess it doesn't save much.

dherman: Yeah I figured out I could do e.g.:

```sh
handlebars -c handlebars -n module.exports -f dist/src/templates/cargo-toml.js data/templates/Cargo.toml.hbs
```

for each template, but then we'd have to add that to the build command for every template, and still need the runtime dependency on `handlebars`. Nice idea but I think it doesn't end up working out.

create-neon/src/bin/create-neon.ts
dherman: Oh ha, `'wx'` means exactly what you asked for: create the file if it doesn't exist, and fail if it already exists.

create-neon/src/bin/create-neon.ts
dherman: The `'wx'` flag does this.

create-neon/data/templates/.gitignore.hbs
```diff
@@ -0,0 +1,4 @@
+target
+index.node
+**/node_modules
+**/.DS_Store
```
kjvalencik, line 4: Should we include anything else from the node `.gitignore` template? https://github.com/github/gitignore/blob/master/Node.gitignore

A lot of them are for specific packages, but `npm-debug.log*` might be nice.

create-neon/src/bin/create-neon.ts
```diff
@@ -0,0 +1,54 @@
+import { mkdir } from 'fs/promises';
```
kjvalencik, line 1: Does this need a hashbang? `#!/usr/bin/env node`

create-neon/src/metadata.ts
```diff
@@ -1,16 +1,66 @@
...
+
+    let test = "cargo test";
+
+    // If the user specifies a non-default test command, use theirs instead.
+    // Ideally there would be better extensibility hooks in `npm init` for
+    // this but there aren't any environment variables or command-line flags
+    // we can use, so we have to guess based on the default value. This also
+    // unfortunately leaks to the user when `npm init` shows the values for
+    // the package.json it's going to use in the final user confirmation.
+    if (!/\s*echo \".*\" && exit 1\s*/.test(json.scripts.test)) {
```
kjvalencik, line 45: If we write a `package.json` _before_ executing `npm init` that contains `{ "scripts": { "test": "cargo test" } }`, it _won't_ prompt for a test command. I think this might be a cleaner workaround than matching on the default text because the user won't even be prompted.

create-neon/src/metadata.ts
```diff
@@ -1,16 +1,66 @@
...
   license: string;
-  description?: FreeText;
+  description: FreeText | undefined;
+
+  static async load(source: string, cargoCpArtifact: string): Promise<Project> {
+    return new Project(JSON.parse(await readFile(source, 'utf8')), cargoCpArtifact);
+  }
+
+  constructor(json: any, cargoCpArtifact: string) {
+    this.json = json;
```
kjvalencik, line 22: Why do we have two copies of the data? Overall, I find the class base approach a little complicated. Alternatively,

```ts
interface PackageJson {
    name: string,
    author: string,
    [k: string]: any,
}
```

There might be a package.json types file we could import.

create-neon/src/bin/create-neon.ts
```diff
@@ -0,0 +1,70 @@
...
+    await expand(source, target, {
+      package: pkg,
+      versions: inferVersions()
+    });
+  }
+
+  console.log(`✨ Created Neon project \`${name}\`. Happy 🦀 hacking! ✨`);
+}
+
+if (process.argv.length !== 3) {
```
kjvalencik, line 60: Is there a particular reason we want to match exactly `3`? Most CLI tools are permissive of additional unnecessary arguments. It's a little more friendly to things like `xargs`.

create-neon/test/create-neon.ts
```diff
@@ -0,0 +1,178 @@
...
+      await(execa(NODE, [CREATE_NEON, 'name', 'ohnoanextraargument']));
+      assert.fail("should fail when too many arguments are supplied");
+    } catch (expected) {
+      assert.isTrue(true);
+    }
+  });
+
+  it('fails if the directory already exists', async () => {
+    try {
+      await execa(NODE, [CREATE_NEON, 'dist']);
```
kjvalencik, line 107: `dist` might not exist. Should we use a different directory?

create-neon/test/create-neon.ts
```diff
@@ -0,0 +1,178 @@
...
+  'description:':    '',
+  'git repository:': '',
+  'keywords:':       '',
+  'author:':         '',
+  'license:':        '',
+  'Is this OK?':     ''
+};
+
+describe('Project creation', () => {
+  it('succeeds with all default answers', async () => {
```
kjvalencik, line 152: This test fails if `npm run build` has not run. Should we add `"pretest": "npm run build"` to the package.json scripts?

kjvalencik (commented):
This is looking awesome! I'm really excited! ✨

create-neon/test/create-neon.ts
```diff
@@ -0,0 +1,178 @@
...
+
+function splitLines(s: string): string[] {
+  return s.split(/([^\n]*\r?\n)/).filter(x => x);
+}
+
+function isCompleteLine(s: string): boolean {
+  return s.endsWith('\n');
+}
+
+class LinesBuffer {
```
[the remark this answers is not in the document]
dherman: Unfortunately no—I tried that at first, but `readline` blocks forever waiting on a prompt that doesn't end with a newline. That's why I bit the bullet and wrote my own line reader that takes a predicate that will even match on an incomplete line.

kjvalencik (approved):
@dherman This looks really good! I like that we start with a fully functional `package.json`. This way even if the user cancels the prompts, they still end up with a fully functional project.

dherman:
> I don't know enough about node on windows to know if this is always `npm.cmd`. If we spawn a shell (by passing `shell: true` to spawn), does it work on Windows without this?

Good thought, I'm running that approach through CI to see if it works on Windows. Seems a bit more robust if so.

PR #691, opened 2021-03-07
Add "napi-embedding" feature to allow statically linked N-API symbols

patr0nus:
This PR adds a new feature `napi-embedding`. When enabled, it loads N-API symbols using `extern "C" { ... }` declarations instead of `libloading`.

## Why?

I am working on [rust-nodejs](https://github.com/patr0nus/rust-nodejs) which embeds Node.js in Rust. The general idea is to build and link Node.js as a static library, and use N-API to communicate between Rust and Node.js.

Neon fits perfectly for the latter part, expect that it tries to load N-API symbols dynamically (https://github.com/neon-bindings/neon/pull/646). This doesn't work in the case of rust-nodejs, because all N-API symbols are statically linked.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ napi-3 = ["napi-2", "neon-runtime/napi-3"]
 napi-4 = ["napi-3", "neon-runtime/napi-4"]
 napi-5 = ["napi-4", "neon-runtime/napi-5"]
 napi-6 = ["napi-5", "neon-runtime/napi-6"]
+napi-embedding = [ "neon-runtime/napi-embedding" ]
 napi-latest = ["napi-6"]
 napi-experimental = ["napi-6", "neon-runtime/napi-experimental"]
 
--- a/crates/neon-runtime/Cargo.toml
+++ b/crates/neon-runtime/Cargo.toml
@@ -12,6 +12,7 @@ cfg-if = "1.0.0"
 libloading = { version = "0.6.5", optional = true }
 neon-sys = { version = "=0.7.1", path = "../neon-sys", optional = true }
 smallvec = "1.4.2"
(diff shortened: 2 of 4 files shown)


kjvalencik:
@patr0nus Can you explain a little bit more about your use case? While the symbols are being loaded dynamically, they are being loaded from the main process and not from a shared library.

Is there possibly a simpler solution to your problem? For example, if the library that the symbols were loaded from was configurable.

The reason I ask is that I'm hesitant to merge this as is because it will not work on Windows. It also might not work on macOS without additional RUSTC flags.

I am also generally curious about the `rust-nodejs` project. It seems a like a really neat idea! However, I am a little curious why a user might want to embed Node in Rust instead of Rust in Node.js. Is it about packaging a single binary?

patr0nus:
@kjvalencik 

> While the symbols are being loaded dynamically, they are being loaded from the main process and not from a shared library.

You are right! It turns out it's because all unused N-API symbols are stripped out by the linker. I'm closing this PR in favor of ["-rdynamic" linker flag](https://github.com/patr0nus/rust-nodejs/commit/e5a15e0e9f620ba3a3374b3fe05f359ca03a6a3e).

> However, I am a little curious why a user might want to embed Node in Rust instead of Rust in Node.js. Is it about packaging a single binary?

Here're my two use cases: 
1. Build an Electron-like framework,
2. Allow user script in an Rust application;

Also I think the `neon` project itself could benefit: with `rust-nodejs` some tests can run directly in-process, without the need of setting up npm packages. Just an idea though. I'm no expert.

kjvalencik:
Thanks @patr0nus! It looks like an interesting project. I'll keep an eye on your progress. Cheers!

PR #692, opened 2021-03-07
Add binding to napi_run_script

patr0nus:


crates/neon-runtime/src/napi/string.rs
```diff
@@ -44,3 +44,13 @@ pub unsafe extern "C" fn data(env: Env, out: *mut u8, len: isize, value: Local)
...
 }
+
+pub unsafe extern "C" fn run_script(out: &mut Local, env: Env, value: Local) -> bool {
+    let status = napi::run_script(
+        env,
+        value,
+        out as *mut _,
+    );
+
+    status == napi::Status::Ok
```
kjvalencik, line 55: What are the failure conditions on this? Do we need to be concerned with other errors besides `Ok` and `Throw`?

src/types/mod.rs
```diff
@@ -360,6 +360,16 @@ impl JsString {
         }
     }
 
+    #[cfg(feature = "napi-1")]
+    pub fn run_as_script<'a, C: Context<'a>>(self, cx: &mut C) -> JsResult<'a, JsValue> {
```
kjvalencik, line 364: @dherman What do you think about marking this as `unsafe`? Even though it's not unsafe by Rust's definition,`eval` is a super very exploit vector and it could be helpful to signal to users that only trusted input should be used.

kjvalencik (commented):
Thanks for the contribution! This seems very straight forward, but I would like to discuss the API a bit further. Specifically if this method should be marked `unsafe`.

crates/neon-runtime/src/napi/string.rs
patr0nus: According to [the implementation of `napi_run_script `](https://github.com/nodejs/node/blob/03806a0bb2f3336bcbf3a668e156993cc2582c71/src/js_native_api_v8.cc#L3036) all code paths that don't return `Ok` have called `napi_set_last_error`. The docs don't state clearly though.
For a peace of mind I've added a test for invalid syntax.

src/types/mod.rs
patr0nus: rustls has a similar case: they want to add an option that disables cert verification, which is very unsafe, but not rust `unsafe`. [What they do](https://github.com/ctz/rustls/issues/68#issuecomment-300912735) is put it behind a feature flag and explicitly say "dangerous" on the api.

crates/neon-runtime/src/napi/string.rs
kjvalencik: Thanks for checking!

src/types/mod.rs
kjvalencik: I like the idea of adding `dangerous` to the function name itself. For example, React has a `dangerouslySetInnerHTML` function. We could call this, `dangerously_run_script`.

Adding the feature flag seems like additional complexity without any benefit. Unlike with `rustls`, the `run_script` API _will still exist_ even if we feature flag it because it's part of Node.

dherman: I think marking `unsafe` when it's not about Rust's safety guarantees is generally considered poor form in the Rust ecosystem, so I agree that's not the best approach. Rather than a prefix, how about a module? Some possibilities:

- `neon::danger::run_script`
- `neon::advanced::run_script`
- `neon::unchecked::run_script`
- `neon::reflect::run_script`

dherman: I should mention, I also agree the feature flag is the wrong approach because it's viral. If a library wants to use the feature they should be able to do it without requiring downstream clients to have to enable the feature flag, and then those clients end up enabling it for everyone anyway, which isn't great either.

dherman: I like `advanced` because it’s a little less paternalistic than “dangerous.” Lets people know they should know what they’re doing but respects their decision if they believe they do.

kjvalencik: I prefer `danger` to `advanced`. I think `advanced` understates the risk while also overstating the complexity of the API.

When I hear advanced, I think, "this API is low level and unergonomic to use." Neither of which is true.

I'm okay with putting it in a module, but I think that signal is a little weak since it's only observed when importing and not on every usage. However, in my opinion, every single usage should be scrutinized.

I tried finding some other examples and the only one I was able to find so far is in `ring` where `less_safe` is suffixed.

https://briansmith.org/rustdoc/ring/?search=safe

dherman: I'm just not sure we really have an actual threat model that we're clearly sticking to (e.g. we don't protect against dangerous APIs accessible from the global object), nor do we really have any kind of enforcement (e.g., users can easily use `use dangerous_eval as eval;`, they can pass `eval` in from JS, etc.). I don't really know what principle we're using for distinguishing dangerous from benign APIs.

kjvalencik: Sorry for the verbosity!

tl;dr-- I like the idea of putting this in `neon::danger` best and removing the `Context` shorthand entirely.

```
use neon::danger;

danger::run_script(&mut cx, "console.log('Hello, World!')");
```

---

@dherman That's a good point that there's other dangerous functionality accessible. Although, most of those aren't accessible as one liners.

For example, accessing `eval` would require getting the global, downcasting to a `JsObject`, getting the `eval` key, downcasting to a `JsFunction`, creating a `JsString` and calling the function.

Your `use dangerous_eval as eval` example has me coming around to the simplicity of putting it in a `danger` module. I would avoid `unchecked` because that has more specific meaning in Rust and typically indicates `unsafe`.

One thing I'll add is that since this _isn't_ called `eval`, it might not be clear that many of the same dangers exist. If I understand correctly, `run_script` is closer to `new Function` than `eval` in accessibility.

dherman (commented):
*Edit:* (moved this to the appropriate sub-thread)

src/types/mod.rs
dherman: Update: @kjvalencik and I spoke this morning, and we came to this suggestion for the naming and documentation:

- It shouldn't be available in any convenience forms (not on `Context` or anywhere else)
- `neon::reflect::eval` is a good name for it: it shows that it's a reflection API, not something to reach for casually, and it connects it to JS's `eval`, which is in fact what it is (technically speaking, it's JS's "indirect eval" but you can't implement direct eval as anything other that a syntax in pure JavaScript source code) and also brings along the mindshare from the JS ecosystem about the dangers of `eval`
- The API docs should clearly explain that use of `eval` can be a security risk and should not be used casually

@patr0nus How does that sound?

patr0nus: @dherman  @kjvalencik 

Putting it under `neon::reflect` sounds good to me. 

However I prefer the name `run_script` rather than `eval`, since it's a direct binding to `napi_run_script`, and as you mentioned, it's not quite the same as `eval` in JS. For example, one might assume `neon::reflect::eval` can access local variables like JS `eval` can.

Also I would suggest the function should accept `JsString` instead of `AsRef<str>`. If a user wants to be run a script repeatedly, he/she wouldn't want a `JsString` to be created every time.

**TL;DR**: I would prefer the function signature to be:

 `fn neon::reflect::run_script<'a, 'b, C: Context<'a>>(cx: &mut C, script: Handle<'b, JsString>) -> JsResult<'a, JsValue>`.

kjvalencik: In my opinion, the N-API naming is not optimal. It _is_ identical to indirect eval (as defined by the ECMA spec).

Considering direct eval is not possible in a native context, eval seems more appropriate.

Neon already diverges from n-api in a number of places. For example, Root instead of reference.

Additionally, `eval` carries the stigma of danger associated with the term, while `run_script` does not. I think this is very important.

I like accepting `JsString`. This is somewhat outside the scope of this PR, but we could improve the API by adding a `ToJsString` trait implemented for `Handle<JsString>`, `&JsString`, and anything that's `AsRef<are>` and making `eval` generic.

patr0nus: That seems reasonable. I have pushed the changes.
Though I didn't include the documentation which explains the security risk, as I am not a native speaker, so not really suitable for explaining this kind of serious stuff.

dherman: No problem! I am going to be working on API docs next so I can follow up in a future PR with some documentation. Thanks so much for your work!

kjvalencik (approved):

kjvalencik:
@patr0nus Thanks for the contribution!

PR #698, opened 2021-03-12
Lint CI enforcement and cargo fmt

kjvalencik:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+        rust-toolchain: [nightly]
+
+    steps:
```
dherman, line 22: I think it's worth adding a comment explaining why we have to do such an explosion of different clippy invocations here.

dherman (approved):
Just one suggestion for adding a comment.

Feels so good to get all these lints and standardizations enforced! I've wanted to standardize the codebase for years. Thank you!

.github/workflows/lint.yml
kjvalencik: I found the cargo package features RFC which provides a way to selectively enable feature flags in a workspace on some packages.

It's working really well. I was able to add `cargo clippy-legacy` and `cargo clippy-core` aliases.

https://doc.rust-lang.org/cargo/reference/unstable.html#package-features

dherman (approved):
Great improvement! :shipit:

PR #700, opened 2021-03-16
Global drop queue for Root

kjvalencik:
# Why

Currently, dropping a `Root` will cause a `panic` instead of leaking the data. This can be problematic for users since there are _many_ ways to accidentally leak a `Root`. Consider the following example:

* A channel accepts closures to execute
* A `Root` is created and captured by a closure
* The closure is sent across the channel
* The channel is closed

In this case, even though the user _knows_ the `Root` won't be used, there's no easy way to get it back out to safely drop it.

# What

This feature is flagged with `napi-6`. The existing `panic` behavior continues for older versions.

## `InstanceData`

A private `InstanceData` type is added that uses `napi_get_instance_data` and `napi_set_instance_data`. This provides a place to stash module specific context for Neon.

## Global drop queue

A global drop queue is added as a `ThreadsafeFunction`.

* `ThreadsafeFunction` is used instead of `EventQueue` since we only need a static callback. Closures would require unnecessary overhead.
* `ThreadsafeFunction` is maintained in an `Arc` to ensure it can outlive an unloaded module
* The queue is unreferenced to prevent it from keeping the event loop running. Hanging references will be freed when the VM drops.

## Additional

Added `impl Object for JsBox` that was missing.

## Testing

It was really fun finding a creative way to test this behavior!

* Create a `struct Wrapper` for holding a callback and an event queue
* The wrapper is put in a `JsBox` which is then rooted and immediately leaked (dropped)
* Instead of an idiomatic `Finalize` implementation, `Wrapper` has a `Drop` implementation that calls the callback. If this is called, we know that the `Root` reference count was dropped, allowing our struct to be dropped.
* We force run the garbage collector after giving the drop queue an opportunity to execute

### DIFF
--- a/crates/neon-runtime/src/napi/lifecycle.rs
+++ b/crates/neon-runtime/src/napi/lifecycle.rs
@@ -1,0 +1,39 @@
+use std::mem::MaybeUninit;
+use std::os::raw::c_void;
+use std::ptr;
+
+use crate::napi::bindings as napi;
+use crate::raw::Env;
+
+/// # Safety
+/// `env` must point to a valid `napi_env` for this thread
+pub unsafe fn set_instance_data<T: Send + 'static>(env: Env, data: T) -> *mut T {
+    let data = Box::into_raw(Box::new(data));
+
+    assert_eq!(
+        napi::set_instance_data(env, data.cast(), Some(drop_box::<T>), ptr::null_mut(),),
+        napi::Status::Ok,
+    );
+
+    data
+}
+
+/// # Safety
+/// * `T` must be the same type used in `set_instance_data`
+/// * Caller must ensure reference does not outlive `Env`
+/// * Return value may be `null`
+/// * `env` must point to a valid `napi_env` for this thread
+pub unsafe fn get_instance_data<T: Send + 'static>(env: Env) -> *mut T {
+    let mut data = MaybeUninit::uninit();
+
+    assert_eq!(
+        napi::get_instance_data(env, data.as_mut_ptr(),),
+        napi::Status::Ok,
+    );
+
+    data.assume_init().cast()
+}
+
+unsafe extern "C" fn drop_box<T>(_env: Env, data: *mut c_void, _hint: *mut c_void) {
+    Box::<T>::from_raw(data.cast());
+}
--- a/crates/neon-runtime/src/napi/lifecycle.rs
+++ b/crates/neon-runtime/src/napi/lifecycle.rs
@@ -1,3 +1,13 @@
+//! # Environment life cycle APIs
+//!
+//! These APIs map to the life cycle of a specific "Agent" or self-contained
+//! environment. If a Neon module is loaded multiple times (Web Workers, worker
+//! threads), these API will be handle data associated with a specific instance.
+//!
+//! See the [N-API Lifecycle][npai-docs] documentation for more details.
+//!
+//! [napi-docs]: https://nodejs.org/api/n-api.html#n_api_environment_life_cycle_apis
+
 use std::mem::MaybeUninit;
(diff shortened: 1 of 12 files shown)


src/handle/root.rs
```diff
@@ -3,23 +3,33 @@ use std::marker::PhantomData;
...
 use object::Object;
 use types::boxed::Finalize;
 
 #[repr(transparent)]
 #[derive(Clone)]
-struct NapiRef(*mut c_void);
+pub(crate) struct NapiRef(*mut c_void);
+
+// # Safety
+// `NapiRef` are intended to be `Send` and `Sync`
```
dherman, line 21: Can you explain why in the doc comment?

src/types/boxed.rs
```diff
@@ -159,6 +160,8 @@ impl<T: Send + 'static> Clone for JsBox<T> {
     }
 }
 
+impl<T: Send + 'static> Object for JsBox<T> {}
```
dherman, line 163: What is this about? Does this mean you can get and set properties on a `JsBox`? Does that work?

test/napi/src/js/threads.rs
```diff
@@ -144,3 +144,44 @@ pub fn leak_event_queue(mut cx: FunctionContext) -> JsResult<JsUndefined> {
 
     Ok(cx.undefined())
 }
+
+pub fn drop_global_queue(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    struct Wrapper {
```
dherman, line 149: Neat use of nested struct types. If you asked me in an interview question if this was legal I wouldn't have known! 😂 But it's very readable.

src/lifecycle.rs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        let drop_queue = Box::new(unsafe {
+            let mut queue = ThreadsafeFunction::new(env, drop_napi_ref);
+            queue.unref(env);
+            queue
+        });
+
+        let data = InstanceData {
+            drop_queue: Box::leak(drop_queue),
```
dherman, line 39: We talked about this a couple days ago. I think I prefer the alternative approach you mentioned of using an `Arc`, and we can consider optimizing in the future.

crates/neon-runtime/src/napi/lifecycle.rs
```diff
@@ -0,0 +1,39 @@
+use std::mem::MaybeUninit;
```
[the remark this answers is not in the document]
dherman: I think it would be helpful for the docs to talk explicitly about these entities:
- "agents" (i.e. the main thread + workers)
- modules
- Roots
and how their lifecycles relate to each other.

dherman (changes requested):
I think this is a great improvement. The code is just subtle enough that my main comments are about comments and docs.

src/types/boxed.rs
kjvalencik: Yes! It was an omission from a prior PR. `napi_external` (which `JsBox` wrap) inherit from `Object`.

I added it in this PR because my test needed to "root" a `JsBox`.

src/lifecycle.rs
kjvalencik: I will make that change and add some documentation on the design in the form of a code comment. Hopefully it will help hint a possible solution if it _does_ pop up as a bottleneck for a user.

test/napi/src/js/threads.rs
kjvalencik: I agree! I prefer this locality for structs and functions that are for a very specific purpose that aren't intended to be generally usable externally.

If I understand correctly, closures are essentially sugar for this.

dherman (approved):
LGTM!

PR #703, opened 2021-03-21
Expanding on Root and a potential common-case

bitemyapp:
Extracted from a real problem I had. I believe this is the only runtime error I encountered. Please let me know what you think.

### DIFF
--- a/src/handle/root.rs
+++ b/src/handle/root.rs
@@ -43,6 +43,21 @@ impl<T: Object> Root<T> {
     /// The caller _must_ ensure `Root::into_inner` or `Root::drop` is called
     /// to properly dispose of the `Root<T>`. If the value is dropped without
     /// calling one of these methods, it will *panic*.
+    ///
+    /// Be careful that you aren't short-circuiting with an early return before
+    /// your Root value gets into_inner'd or dropped. The following will cause
+    /// a runtime panic when your error case gets triggered:
+    /// ```
+    /// let callback = cx.argument::<JsFunction>(1)?.root(&mut cx);
+    /// let my_log = match (create_log_entry(&id_generator, "log-emitter")) {
+    ///                Err(_err) => {
+    ///                  return cx.throw_error("Couldn't construct log");
+    ///                },
+    ///                Ok(log) => log,
+    /// };
+    /// ```
+    /// The solution in the original case for this was to bind the callback
+    /// after the fallible code, right before spawning an async task.
     pub fn new<'a, C: Context<'a>>(cx: &mut C, value: &T) -> Self {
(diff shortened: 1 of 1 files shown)


src/handle/root.rs
```diff
@@ -43,6 +43,21 @@ impl<T: Object> Root<T> {
     /// The caller _must_ ensure `Root::into_inner` or `Root::drop` is called
     /// to properly dispose of the `Root<T>`. If the value is dropped without
     /// calling one of these methods, it will *panic*.
+    ///
+    /// Be careful that you aren't short-circuiting with an early return before
+    /// your Root value gets into_inner'd or dropped. The following will cause
+    /// a runtime panic when your error case gets triggered:
+    /// ```
```
kjvalencik, line 50: This is failing doc tests for me locally. I'm not sure why it's not failing in CI. 😕

src/handle/root.rs
```diff
@@ -43,6 +43,21 @@ impl<T: Object> Root<T> {
     /// The caller _must_ ensure `Root::into_inner` or `Root::drop` is called
     /// to properly dispose of the `Root<T>`. If the value is dropped without
     /// calling one of these methods, it will *panic*.
+    ///
+    /// Be careful that you aren't short-circuiting with an early return before
```
kjvalencik, line 47: Since this communicates the same as above, can it be reworded to present it as an example?

> For example, early return with a `?` may cause a function to return before freeing a `Root`. ...

src/handle/root.rs
```diff
@@ -43,6 +43,21 @@ impl<T: Object> Root<T> {
...
+    /// your Root value gets into_inner'd or dropped. The following will cause
+    /// a runtime panic when your error case gets triggered:
```
kjvalencik, line 49: This is no longer strictly accurate after merging https://github.com/neon-bindings/neon/pull/700

kjvalencik (commented):
Thanks for your contribution! I have a few suggestions on wording. Cheers!

src/handle/root.rs
bitemyapp: I think I have fixed this, I wasn't even thinking about doc tests when I wrote the code.

src/handle/root.rs
bitemyapp: I've changed it to delineate panic prior to 0.8, slow path otherwise.

src/handle/root.rs
bitemyapp: I've attempted a rewording along the lines suggested here.

bitemyapp:
I merged the doc update with what I did here.

PR #705, opened 2021-03-24
API docs: Neon root, module summaries, neon::borrow

dherman:
- main Neon root doc
- all module summaries
- `neon::borrow` module

[preview](https://neon-apidocs-preview.surge.sh/neon/)


src/borrow/mod.rs
```diff
@@ -1,4 +1,33 @@
-//! Types and traits for obtaining temporary access to the internals of JavaScript values.
+//! Provides temporary access to JavaScript typed arrays.
+//!
+//! ## Typed Arrays
+//!
+//! JavaScript's [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays)
```
kjvalencik, line 5: FYI, links support anchors which can be useful for limiting the impact to flow within the next. For example:

```
This is a link to [Rust Lang][rust-lang]. It's really cool and you should check it out.

[rust-lang]: https://www.rust-lang.org/
```

This is a link to [Rust Lang][rust-lang]. It's really cool and you should check it out.

[rust-lang]: https://www.rust-lang.org/

src/borrow/mod.rs
```diff
@@ -1,4 +1,33 @@
...
+//! and several typed views for managing access to the buffer.
+//! Neon provides access to the `ArrayBuffer` class with the
+//! [`JsArrayBuffer`](../../neon/types/struct.JsArrayBuffer.html) type.
+//!
+//! Node also provides a [`Buffer`](https://nodejs.org/api/buffer.html) type,
+//! which is built on top of `ArrayBuffer` and provides additional functionality.
+//! Neon provides access to the `Buffer` class with the
+//! [`JsBuffer`](../../neon/types/struct.JsBuffer.html) type.
+//!
+//! Many of Node's I/O APIs work with these types, and they can also be used for
```
kjvalencik, line 20: It may be beneficial to mention that they can be used to efficiently share data between JavaScript and Rust _without_ copying.

src/lib.rs
```diff
@@ -1,4 +1,77 @@
-//! The [Neon](https://www.neon-bindings.com/) crate provides bindings for writing Node.js plugins with a safe and fast Rust API.
+//! The [Neon](https://www.neon-bindings.com/) crate provides bindings for
+//! writing [Node.js addons](https://nodejs.org/api/addons.html) (i.e.,
+//! dynamically-linked binary modules) with a safe and fast Rust API.
```
kjvalencik, line 3:
```suggestion
//! dynamically-loaded binary modules) with a safe and fast Rust API.
```

src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+//!
+//! ## Getting Started
+//!
+//! You can conveniently bootstrap a new Neon project with the Neon project
+//! generator. You don't need to install anything special on your machine as
+//! long as you have a supported version of Node on your system.
```
kjvalencik, line 9: Please also mention that Rust/Cargo is necessary.

src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+//! ```
+//!
+//! You can look in the project's generated `README.md` for more details on
+//! the project structure.
+//!
+//! ## Example
+//!
+//! The generated `src/lib.rs` contains a function annotated with the
+//! `#[neon::main]` attribute, marking it as the module's main entry point
+//! to be executed when the module is first loaded. This function can have
```
kjvalencik, line 43: Remove the word `first` since a module may be loaded multiple times.

```suggestion
//! to be executed when the module is loaded. This function can have
```

src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+//!
+//! ```
+//! # use neon::prelude::*;
+//! #
+//! fn hello(mut cx: FunctionContext) -> JsResult<JsString> {
+//!     Ok(cx.string("hello node"))
+//! }
+//! ```
+//!
+//! The `hello` function takes a [`FunctionContext`](./context/type.FunctionContext.html) and
```
kjvalencik, line 72: Rust doc was recently updated to support linking by name.

```suggestion
//! The `hello` function takes a [`FunctionContext`](context::FunctionContext) and
```

This is more robust against HTML structure changes. There are also shorthand versions when the name and display text are the same.

https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html

*Note*: This feedback applies to multiple places in these docs, but those are omitted for brevity.

src/prelude.rs
```diff
@@ -1,4 +1,4 @@
-//! A convenience module that re-exports the most commonly-used Neon APIs.
+//! Convenience module for the most common Neon imports.
```
kjvalencik, line 1: Should we consider removing this or refactoring as part of `1.0`? It includes nearly everything in `neon`.

dherman: I kind of like the prelude idiom but I'm happy to discuss it! Let's discuss on Slack or in a separate issue thread.

src/lib.rs
dherman: Oh I was hoping that would be the case, awesome!

kjvalencik (approved):

PR #715, opened 2021-04-27
Fix build with node 16

ogoffart:
After archlinux upgraded to the newer version of node, neon-sys stoped building with this error:

```
warning: ar: target/debug/build/neon-sys-e960e2d326235d28/out/native/build/Debug/obj.target/neon/src/neon.o: No such file or directory
error: failed to run custom build command for `neon-sys v0.8.0`
```
After investigation, the real error is: (this PR also make sure to print compilation errors, if any)
```
../src/neon.cc: In function 'void Neon_Class_SetClassMap(v8::Isolate*, void*, Neon_DropCallback)':
  ../src/neon.cc:335:16: error: cannot convert 'void (*)(void*)' to 'node::Environment*'
    335 |   node::AtExit(cleanup_class_map, holder);
        |                ^~~~~~~~~~~~~~~~~
        |                |
        |                void (*)(void*)
```

It turns out that the overload taking 2 arguments was removed in the latest version of node: https://github.com/nodejs/node/commit/dfc288e7fdbe888f4e134772de421684247f0bea#diff-4acdf5a5c11188763423b746a9a43917c90546f3b48b9aa6f6eeb60e157ae2a7

This PR calls the remaining overload which have been there before.

I'd appreciate a release on crates.io with this change, thanks!

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
+        node-version: [10.x, 12.x, 14.x, 16.x]
         rust-toolchain: [stable, beta, nightly]
 
     steps:
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
+        node-version: [10.x, 12.x, 14.x, 16.x]
         rust-toolchain: [stable, beta, nightly]
 
     steps:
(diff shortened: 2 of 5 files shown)


kjvalencik:
@ogoffart Thanks for the contribution! Do you mind updating the Github CI workflows to reflect the current support Node versions by _removing_ Node 15 and *add* Node 16?

ogoffart:
> Do you mind updating the Github CI workflows to reflect the current supported Node versions by removing Node 15 and adding Node 16?

I can do that. 
Sure you want to remove Node 15?

kjvalencik:
@ogoffart Yes, we officially only support latest and active LTS. As soon as 16 was released, we no longer supported 15. It will also be dropped by Node support in June.

kjvalencik:
@ogoffart Thanks again for this fix! This has been published as `0.8.1`.

PR #717, opened 2021-04-27
Update README to mention Node 16 support

### LINKED ISSUE #716: Mention Node 16 support in the README
I recently updated my system to use the new Node 16.0.0 which was released a few days ago, and I'm no longer able to `npm install` a library I'm using with Neon. I filed an issue with a traceback [here](https://github.com/brave/adblock-rust/issues/167) as I assumed I just needed to update Neon to the latest version, but it appears to still have issues with Neon 8.0 and even with the newer N-API feature. I verified that it still builds correctly when switching back to Node 15 using `nvm`.

It would be great to see support for Node 16 if possible, considering that it is now officially the "Current" release.

Also, the README [only lists](https://github.com/neon-bindings/neon#nodejs) Node 10, Node 12, and Node 14 compatibility - perhaps it would be worth updating with an extra column for Node 16.

antonok-edm:
Closes #716 :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ We've ported Neon to a new backend based on [N-API](https://nodejs.org/api/n-api
 
 ### Node.js
 
-| Node 10 | Node 12 | Node 14 |
-| ------- | ------- | ------- |
-| ✓       | ✓       | ✓       |
+| Node 10 | Node 12 | Node 14 | Node 16 |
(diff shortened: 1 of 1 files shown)


amilajack (approved):

kjvalencik:
Thanks for docs update!

PR #734, opened 2021-05-15
Rename `EventQueue` to `Channel`

dherman:
This PR experiments with renaming the event queue API:

- `neon::event::EventQueue` -> `neon::event::Channel`
- `neon::event::EventQueueError` -> `neon::event::SendError`

This is based on an insight @kjvalencik made recently: a clone/copy/sync-able data structure that gives access across multiple threads to the Node.js runtime's internal event queue by a `send` operation that communicates across threads is much closer conceptually to Rust's idiomatic concept of channels than it is to being an abstraction of the JS event queue itself.

The PR attempts to make the change conservatively, preserving the previous names as aliases for the new names but adding deprecation attributes and hiding them from API docs. This allows us to ship the new name without breaking existing customers and eventually remove the deprecated names in a future release.

Here's a [preview of updated `neon::event` docs with code examples](https://neon-event-apidocs.surge.sh/neon/event/index.html) where you can see how the code looks with the renamed API.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -152,7 +152,7 @@ use crate::borrow::internal::Ledger;
 use crate::borrow::{Borrow, BorrowMut, Ref, RefMut};
 use crate::context::internal::Env;
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-use crate::event::EventQueue;
+use crate::event::Channel;
 use crate::handle::{Handle, Managed};
 #[cfg(feature = "legacy-runtime")]
 use crate::object::class::Class;
@@ -550,9 +550,19 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-    /// Creates an unbounded queue of events to be executed on a JavaScript thread
-    fn queue(&mut self) -> EventQueue {
-        EventQueue::new(self)
+    /// Creates an unbounded channel for scheduling events to be executed on the JavaScript thread.
+    fn channel(&mut self) -> Channel {
+        Channel::new(self)
+    }
+
+    #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
+    #[deprecated(
+        since = "0.9.0",
+        note = "Please use the channel() method instead"
+    )]
+    #[doc(hidden)]
+    fn queue(&mut self) -> Channel {
+        self.channel()
     }
 }
 
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -556,10 +556,7 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-    #[deprecated(
-        since = "0.9.0",
-        note = "Please use the channel() method instead"
-    )]
+    #[deprecated(since = "0.9.0", note = "Please use the channel() method instead")]
     #[doc(hidden)]
     fn queue(&mut self) -> Channel {
         self.channel()
--- a/src/event/event_queue.rs
+++ b/src/event/event_queue.rs
@@ -21,7 +21,7 @@ type Callback = Box<dyn FnOnce(Env) + Send + 'static>;
 ///     // across threads.
 ///     let n = cx.argument::<JsNumber>(0)?.value(&mut cx);
 ///     let callback = cx.argument::<JsFunction>(1)?.root(&mut cx);
-///     let queue = cx.queue();
+///     let channel = cx.channel();
 ///
 ///     // Spawn a thread to complete the execution. This will _not_ block the
 ///     // JavaScript event loop.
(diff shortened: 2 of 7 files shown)


src/context/mod.rs
```diff
@@ -550,9 +550,16 @@ pub trait Context<'a>: ContextInternal<'a> {
...
-    /// Creates an unbounded queue of events to be executed on a JavaScript thread
-    fn queue(&mut self) -> EventQueue {
-        EventQueue::new(self)
+    /// Creates an unbounded channel for scheduling events to be executed on the JavaScript thread.
+    fn channel(&mut self) -> Channel {
+        Channel::new(self)
+    }
+
+    #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
+    #[deprecated(since = "0.9.0", note = "Please use the channel() method instead")]
```
dherman, line 559: @kjvalencik Is 0.9.0 the right version to choose here? I was guessing that would be the next version we ship.

src/event/event_queue.rs
```diff
@@ -138,19 +138,19 @@ impl EventQueue {
     }
 }
 
-/// Error indicating that a closure was unable to be scheduled to execute on the event queue
-pub struct EventQueueError;
+/// Error indicating that a closure was unable to be scheduled to execute on the event loop.
+pub struct SendError;
```
dherman, line 142: I chose `SendError` but we could also potentially choose `TrySendError`? The `mpsc` module has both types where we only have one. I thought `TrySendError` seemed unnecessarily pedantic, but I'm open to alternatives.

src/prelude.rs
```diff
@@ -10,6 +10,9 @@ pub use crate::declare_types;
 #[cfg(all(not(feature = "napi-1"), feature = "event-handler-api"))]
 pub use crate::event::EventHandler;
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
+pub use crate::event::{Channel, SendError};
+#[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
+#[allow(deprecated)]
```
dherman, line 15: Exposing these in the prelude means that people using `use neon::prelude::*;` won't get deprecation warnings, but it seemed better not to break backwards compatibility. Seem like the right call?

src/event/event_queue.rs
kjvalencik: `SendError` looks like the correct one. `TrySendError` is for a non-blocking send on a bounded queue. Currently, we only have an unbounded version of `EventQueue`.

src/context/mod.rs
kjvalencik: Yes, we can ship that next. 👍  I'll keep that in mind for the next release.

src/prelude.rs
kjvalencik: Since these are only type aliases and aliases are identical to the type they refer, maybe we can _duplicate_ the aliases here instead of importing them?

Will that cause _all_ users to get the deprecation warning even if they aren't using it?

kjvalencik (commented):
I really, _really_ like this naming change. I think it clarifies a lot. It also resolves some naming questions for when we implement bounded queues. We can call it `cx.sync_channel()`.

src/event/event_queue.rs
```diff
@@ -49,19 +49,19 @@ type Callback = Box<dyn FnOnce(Env) + Send + 'static>;
...
+impl std::fmt::Debug for Channel {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str("EventQueue")
+        f.write_str("Channel")
     }
 }
 
-impl EventQueue {
-    /// Creates an unbounded queue for scheduling closures on the JavaScript
+impl Channel {
```
kjvalencik, line 63: What do you think of `Sender` to make it align even more with `std`? I think I still favor `Channel`.

dherman: I think `Sender` only makes sense when there's also a `Receiver`. I think I prefer `Channel` as well.

dherman:
> I really, _really_ like this naming change. I think it clarifies a lot.

Right?? I'm so glad you brought this idea up.

> It also resolves some naming questions for when we implement bounded queues. We can call it `cx.sync_channel()`.

Oh, I like that. I know it's FCP, but we should probably update neon-bindings/rfcs#32 with the new name and a note about `sync_channel()` listed as a possible future extension.

src/prelude.rs
dherman: Oh, I was wrong! I did an experiment and it turns out that when you import `use neon::prelude::*;` it only warns when you actually use the deprecated names, so that's actually perfect!

[example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3e92e2161229edbc62bbb16936ca34e2)

dherman:
@kjvalencik Do you think this is ready to merge?

src/prelude.rs
kjvalencik: @dherman Did you need to make a change for this?

dherman: No, what I figured out was that this is perfect:

- the `#[allow(deprecated)]` inside the prelude means that Neon itself builds without warnings
- if anyone does `use neon::prelude::*;` without referencing the deprecated names, they get zero warnings
- if anyone does `use neon::prelude::*;` and references the deprecated names, they get a warning at the place where they reference the deprecated names

So it's exactly what we'd want, the way it is.

kjvalencik (approved):

PR #735, opened 2021-05-16
Types diagram for docs

dherman:
Small PR to add a type hierarchy diagram to the repo, which we can link to from the docs.

I'm putting this in a PR by itself as a workaround for rust-lang/rust#32104. If we merge the image into the repo then it'll have a URL I can use in the API docs, and that will work in any environment (locally, in a temporary CDN-hosted preview, and/or on docs.rs).


kjvalencik (approved):

kjvalencik:
Looks great @dherman !

Issue #727, opened 2021-05-07
Optimization: share a single napi_threadsafe_function for all EventQueues


jrose-signal:
Today each EventQueue has its own napi_threadsafe_function to send tasks to the Node event loop. However, nodejs/node#38506 (by @indutny-signal) makes it faster to run multiple calls on the same napi_threadsafe_function than to hop between them. If Neon saves a single napi_threadsafe_function in its instance data, it can be shared among all EventQueues for a particular instance.

One complication with this is the `reference`/`unref`/`has_ref` API on EventQueue, which uses the lifetime of the napi_threadsafe_function to keep the main event loop alive. This would have to be implemented with refcounting alongside the single napi_threadsafe_function:
1. The napi_threadsafe_function starts as unref with a refcount of 0.
2. Creating an EventQueue adds one to the refcount.
3. Going from 0 to 1 calls `napi_ref_threadsafe_function`.
4. `reference` and `unref` modify the refcount only if they change the local `has_ref` flag. 
5. Going from 1 to 0 calls `napi_unref_threadsafe_function`.

kjvalencik:
This is an interesting idea and I could see how it could be beneficial in many use cases. I think if this feature were added, I would want to maintain both behaviors because there are valid use cases for maintaining multiple queues (for example, pre-empting events in other queues).

One ergonomic way to handle this might be `cx.queue()` returns the shared queue (more common and better in most cases), while `EventQueue::new()` creates an independent queue.

The design space is probably large enough that an RFC would be beneficial.

indutny-signal:
(Copy-pasting my response from Slack for posterity)

@kjvalencik  I see what you mean, but the behavior of threadsafe function in n-api is intentionally underspecified. It provides you with a way to invoke a function on JS thread from another non-JS thread and that's it. The fact that they use different uv_async_t under the hood is an implementation detail, not a feature!

On a more technical side, though, users won't preempt event queue by managing several different instances. What would happen is in fact the opposite, call one event queue fast enough and the other one is going to wait for callbacks of the first one to complete.

With all of this in mind, I would say that [that PR](https://github.com/indutny/neon/commit/76853a9b6cb10c2b5500d4076103309736b020c5) might be best viewed as a performance optimization and not as a behavior or API change. FWIW, it even creates new EventQueue instances so the API surface is exactly the same before and after this change.

Let me know if I can elaborate more on it.

kjvalencik:
[...]

PR #739, opened 2021-05-18
Reuse ThreadsafeFunction in EventQueue

indutny:
Node.js optimizes subsequent ThreadsafeFunction invocations to happen
during the same event loop tick, but only if the same instance of
ThreadsafeFunction is used. The performance improvement is most
noticeable when used in Electron, because scheduling a new UV tick in
Electron is very costly.

With this change EventQueue will use an
existing instance of ThreadsafeTrampoline (wrapper around
ThreadsafeFunction) if compiled with napi-6 feature, or it will fallback
to creating a new ThreadsafeFunction per EventQueue instance.

Fix: #727

### DIFF
--- a/crates/neon-runtime/src/napi/tsfn.rs
+++ b/crates/neon-runtime/src/napi/tsfn.rs
@@ -153,20 +153,26 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
 
     /// References a threadsafe function to prevent exiting the event loop until it has been dropped. (Default)
     /// Safety: `Env` must be valid for the current thread
-    pub unsafe fn reference(&mut self, env: Env) {
-        assert_eq!(
-            napi::ref_threadsafe_function(env, self.tsfn.0,),
-            napi::Status::Ok,
-        );
+    pub unsafe fn reference(&mut self, env: Env) -> Result<(), napi::Status> {
+        let status = napi::ref_threadsafe_function(env, self.tsfn.0);
+
+        if status == napi::Status::Ok {
+            Ok(())
+        } else {
+            Err(status)
+        }
     }
 
     /// Unreferences a threadsafe function to allow exiting the event loop before it has been dropped.
     /// Safety: `Env` must be valid for the current thread
-    pub unsafe fn unref(&mut self, env: Env) {
-        assert_eq!(
-            napi::unref_threadsafe_function(env, self.tsfn.0,),
-            napi::Status::Ok,
-        );
+    pub unsafe fn unref(&mut self, env: Env) -> Result<(), napi::Status> {
+        let status = napi::unref_threadsafe_function(env, self.tsfn.0);
+
+        if status == napi::Status::Ok {
+            Ok(())
+        } else {
+            Err(status)
+        }
     }
 
     // Provides a C ABI wrapper for a napi callback notifying us about tsfn
--- a/crates/neon-runtime/src/napi/tsfn.rs
+++ b/crates/neon-runtime/src/napi/tsfn.rs
@@ -156,7 +156,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
     pub unsafe fn reference(&self, env: Env) {
         assert_eq!(
             napi::ref_threadsafe_function(env, self.tsfn.0),
-            napi::Status::Ok
+            napi::Status::Ok,
         );
     }
 
@@ -165,7 +165,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
     pub unsafe fn unref(&self, env: Env) {
         assert_eq!(
             napi::unref_threadsafe_function(env, self.tsfn.0),
-            napi::Status::Ok
+            napi::Status::Ok,
         );
     }
 
--- a/crates/neon-runtime/src/napi/tsfn.rs
+++ b/crates/neon-runtime/src/napi/tsfn.rs
@@ -155,7 +155,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
     /// Safety: `Env` must be valid for the current thread
     pub unsafe fn reference(&self, env: Env) {
         assert_eq!(
-            napi::ref_threadsafe_function(env, self.tsfn.0),
+            napi::ref_threadsafe_function(env, self.tsfn.0,),
             napi::Status::Ok,
         );
     }
@@ -164,7 +164,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
     /// Safety: `Env` must be valid for the current thread
     pub unsafe fn unref(&self, env: Env) {
         assert_eq!(
-            napi::unref_threadsafe_function(env, self.tsfn.0),
+            napi::unref_threadsafe_function(env, self.tsfn.0,),
             napi::Status::Ok,
         );
     }
--- a/crates/neon-runtime/src/napi/tsfn.rs
+++ b/crates/neon-runtime/src/napi/tsfn.rs
@@ -153,26 +153,20 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
 
     /// References a threadsafe function to prevent exiting the event loop until it has been dropped. (Default)
     /// Safety: `Env` must be valid for the current thread
-    pub unsafe fn reference(&mut self, env: Env) -> Result<(), napi::Status> {
-        let status = napi::ref_threadsafe_function(env, self.tsfn.0);
-
-        if status == napi::Status::Ok {
-            Ok(())
-        } else {
-            Err(status)
-        }
+    pub unsafe fn reference(&self, env: Env) {
+        assert_eq!(
+            napi::ref_threadsafe_function(env, self.tsfn.0),
+            napi::Status::Ok
+        );
     }
 
     /// Unreferences a threadsafe function to allow exiting the event loop before it has been dropped.
     /// Safety: `Env` must be valid for the current thread
-    pub unsafe fn unref(&mut self, env: Env) -> Result<(), napi::Status> {
-        let status = napi::unref_threadsafe_function(env, self.tsfn.0);
-
-        if status == napi::Status::Ok {
-            Ok(())
-        } else {
-            Err(status)
-        }
+    pub unsafe fn unref(&self, env: Env) {
+        assert_eq!(
+            napi::unref_threadsafe_function(env, self.tsfn.0),
+            napi::Status::Ok
+        );
     }
 
     // Provides a C ABI wrapper for a napi callback notifying us about tsfn
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -154,6 +154,8 @@ use crate::context::internal::Env;
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
 use crate::event::Channel;
 use crate::handle::{Handle, Managed};
+#[cfg(feature = "napi-6")]
+use crate::lifecycle::InstanceData;
 #[cfg(feature = "legacy-runtime")]
 use crate::object::class::Class;
 use crate::object::{Object, This};
@@ -551,8 +553,17 @@ pub trait Context<'a>: ContextInternal<'a> {
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
     /// Creates an unbounded channel for scheduling events to be executed on the JavaScript thread.
+    ///
+    /// When using N-API >= 6,the channel returned by this method is backed by a shared queue.
+    /// To create a channel backed by a _new_ queue see [`Channel`](crate::event::Channel).
     fn channel(&mut self) -> Channel {
-        Channel::new(self)
+        #[cfg(feature = "napi-6")]
+        let channel = InstanceData::channel(self);
+
+        #[cfg(not(feature = "napi-6"))]
+        let channel = Channel::new(self);
+
+        channel
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -154,7 +154,7 @@ use crate::context::internal::Env;
 #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
 use crate::event::Channel;
 use crate::handle::{Handle, Managed};
-#[cfg(feature = "napi-6")]
+#[cfg(all(feature = "napi-6", feature = "event-queue-api"))]
 use crate::lifecycle::InstanceData;
 #[cfg(feature = "legacy-runtime")]
 use crate::object::class::Class;
@@ -552,7 +552,7 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-    /// Creates an unbounded channel for scheduling events to be executed on the JavaScript thread.
+    /// Clones a shared unbounded channel for scheduling events to be executed on the JavaScript thread.
     ///
     /// When using N-API >= 6,the channel returned by this method is backed by a shared queue.
     /// To create a channel backed by a _new_ queue see [`Channel`](crate::event::Channel).
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -552,7 +552,7 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-    /// Clones a shared unbounded channel for scheduling events to be executed on the JavaScript thread.
+    /// Returns an unbounded channel for scheduling events to be executed on the JavaScript thread.
     ///
     /// When using N-API >= 6,the channel returned by this method is backed by a shared queue.
     /// To create a channel backed by a _new_ queue see [`Channel`](crate::event::Channel).
--- a/src/event/event_queue.rs
+++ b/src/event/event_queue.rs
@@ -1,3 +1,5 @@
+use std::sync::{Arc, RwLock};
+
 use neon_runtime::raw::Env;
 use neon_runtime::tsfn::ThreadsafeFunction;
 
@@ -7,6 +9,9 @@ use crate::result::NeonResult;
 type Callback = Box<dyn FnOnce(Env) + Send + 'static>;
 
 /// Channel for scheduling Rust closures to execute on the JavaScript main thread.
+///
+/// Cloning a `Channel` will create a new channel that shares a backing queue for
+/// events.
 ///
 /// # Example
 ///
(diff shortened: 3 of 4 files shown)


indutny:
Converted to draft because of the reference counting issues. Will keep you posted.

indutny-signal:
(Force pushed to fix lint errors)

kjvalencik:
I left a comment here mentioning a couple of my thoughts: https://github.com/neon-bindings/neon/issues/727#issuecomment-844070242

indutny-signal:
Replied.

src/trampoline.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// Safety: `Env` must be valid for the current thread
+    pub(crate) fn reference(&mut self, env: Env) {
+        unsafe {
+            self.tsfn.reference(env);
+        }
+    }
+
+    /// Unreferences a trampoline to allow exiting the event loop before it has been dropped.
+    /// Safety: `Env` must be valid for the current thread
+    pub(crate) fn unref(&mut self, env: Env) {
```
kjvalencik, line 52: `unreference` will not operate correctly in the presence of mixed usage. It will immediately `unref` the shared queue even if referenced copies still exist.

src/event/event_queue.rs
```diff
@@ -64,30 +65,48 @@ impl EventQueue {
...
+
+        // Start referenced
+        queue.reference(cx);
+
+        queue
     }
 
     /// Allow the Node event loop to exit while this `EventQueue` exists.
     /// _Idempotent_
     pub fn unref<'a, C: Context<'a>>(&mut self, cx: &mut C) -> &mut Self {
```
kjvalencik, line 87: `unref` will not operate correctly in the presence of mixed usage. It will immediately `unref` the shared queue even if referenced copies still exist.

indutny-signal: I think it does the checks here to make sure that global instance is referenced/unreferenced properly. I'll move the logic into the `trampoline.rs`.

indutny-signal: I believe I've addressed this by making `reference()`/`unref()` not idempotent in `ThreadsafeTrampoline` in the latest commit in this PR. The way it works is that we enforce idempotency on the level of `EventQueue` and only when the value of `has_ref` actually changes - we call `reference()`/`unref()` on the parent `ThreadsafeTrampoline`. It in turn will track the total number of references and invoke `ref_threadsafe_function`/`unref_threadsafe_function` whenever the internal reference count becomes zero or non-zero.

src/trampoline.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    /// Unreferences a trampoline to allow exiting the event loop before it has been dropped.
+    /// Safety: `Env` must be valid for the current thread
+    /// _Not idempotent_
+    pub(crate) fn unref(&mut self, env: Env) {
+        if self.ref_count == 0 {
+            return;
+        }
```
jrose-signal, line 65: This should probably panic, or however else Neon detects internal errors.

src/lifecycle.rs
```diff
@@ -73,4 +91,12 @@ impl InstanceData {
     pub(crate) fn drop_queue<'a, C: Context<'a>>(cx: &mut C) -> Arc<ThreadsafeFunction<NapiRef>> {
         Arc::clone(&InstanceData::get(cx).drop_queue)
     }
+
+    /// Helper to return a reference to the `invoke_callback` field of `InstanceData`
+    #[cfg(feature = "event-queue-api")]
+    pub(crate) fn threadsafe_trampoline<'a, C: Context<'a>>(
+        cx: &mut C,
+    ) -> Arc<RwLock<ThreadsafeTrampoline>> {
+        Arc::clone(&InstanceData::get(cx).threadsafe_trampoline)
```
jrose-signal, line 100: Nitpick: it might be nice to lazily initialize this for programs that don't use EventQueues.

src/trampoline.rs
```diff
@@ -0,0 +1,86 @@
...
+use neon_runtime::tsfn::{CallError, ThreadsafeFunction};
+
+use crate::context::TaskContext;
+use crate::result::NeonResult;
+
+pub(crate) type Callback = Box<dyn FnOnce(Env) + Send + 'static>;
+
+pub(crate) struct ThreadsafeTrampoline {
+    tsfn: ThreadsafeFunction<Callback>,
+    ref_count: u32,
```
jrose-signal, line 11: Nitpick: if this were AtomicU32, you could avoid the lock. (You'd be able to use Relaxed operations safely because you're only checking the refcount, not relying on any other data to have been synchronized.)

src/trampoline.rs
```diff
@@ -0,0 +1,86 @@
...
+            });
+        });
+
+        self.tsfn.call(callback, None)
+    }
+
+    /// References a trampoline to prevent exiting the event loop until it has been dropped. (Default)
+    /// Safety: `Env` must be valid for the current thread
+    /// _Not idempotent_
+    pub(crate) fn reference(&mut self, env: Env) {
```
jrose-signal, line 49: Nitpick: Maybe we should make this more obviously distinct from the idempotent nature of `ref` by calling it `increment_reference_count` or something.

src/trampoline.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        if self.ref_count != 0 {
+            return;
+        }
+        unsafe {
+            self.tsfn.unref(env);
+        }
+    }
+
+    // Monomorphized trampoline funciton for calling the user provided closure
```
jrose-signal, line 76: Typo: function

src/event/event_queue.rs
```diff
@@ -64,30 +63,51 @@ impl EventQueue {
...
     /// main thread
     pub fn new<'a, C: Context<'a>>(cx: &mut C) -> Self {
-        let tsfn = unsafe { ThreadsafeFunction::new(cx.env().to_raw(), Self::callback) };
+        let trampoline = Arc::new(RwLock::new(ThreadsafeTrampoline::new(cx.env().to_raw())));
 
-        Self {
-            tsfn,
-            has_ref: true,
-        }
+        Self::with_trampoline(cx, trampoline)
```
jrose-signal, line 68: `new` returns a referenced trampoline, but `with_trampoline` expects an unreferenced one. I think the best answer here is to have `with_trampoline` take an already-referenced trampoline, which is consistent with fresh ones and not hard to implement for the shared one. Having it take a not-already-referenced one and immediately referencing it would also work, but would mean extra calls into Node for the fresh case.

jrose-signal: In that case it should probably be called `with_referenced_trampoline` or something so people don't forget.

src/trampoline.rs
indutny-signal: Sadly, the lock is needed because `.reference()`/`.unref()` calls need mutable access to the `ThreadsafeFunction`.

src/lifecycle.rs
indutny-signal: It sounds like in order to that I'll have to store it in an `Arc<RwLock<Option<Arc<RwLock<ThreadsafeTrampoline>>>>>` which doesn't look to nice to me. Is there any better way to do this?

indutny-signal:
@jrose-signal I believe I've addressed your feedback except for one suggestion. PTAL

src/context/mod.rs
```diff
@@ -561,12 +561,16 @@ pub trait Context<'a>: ContextInternal<'a> {
...
-        let trampoline = InstanceData::threadsafe_trampoline(self);
+        let shared_trampoline = InstanceData::threadsafe_trampoline(self);
 
         #[cfg(not(feature = "napi-6"))]
-        let trampoline = ThreadsafeTrampoline::new(self.env());
+        let shared_trampoline = {
+            let shared_trampoline = ThreadsafeTrampoline::new(self.env());
+            shared_trampoline.unref(self.env().to_raw());
+            shared_trampoline
+        };
```
jrose-signal, line 571: This is still a little less efficient but it's harder to get wrong, so maybe that's worth it.

src/trampoline.rs
```diff
@@ -59,9 +59,9 @@ impl ThreadsafeTrampoline {
     /// Unreferences a trampoline to allow exiting the event loop before it has been dropped.
     /// Safety: `Env` must be valid for the current thread
     /// _Not idempotent_
-    pub(crate) fn unref(&mut self, env: Env) {
+    pub(crate) fn decrement_references(&mut self, env: Env) {
         if self.ref_count == 0 {
-            return;
+            panic!("ThreadsafeTrampoline reference underflow");
         }
```
jrose-signal: Ah, Rust tip: this is probably better written as `assert!(self.ref_count > 0, "…")`, which is exactly equivalent.

jrose-signal (commented):
👍

src/trampoline.rs
indutny-signal: Good point. Fixed!

src/lifecycle.rs
jrose-signal: You have an exclusive (`&mut`) reference to InstanceData, so you don't need extra locking. `Option<Arc<_>>` is sufficient.

indutny-signal:
The [latest commit](https://github.com/neon-bindings/neon/pull/739/commits/c22f9321e242145cf3f098fdbaa855cc54490556) on this PR actually fixes a bug, but it doesn't look like it could be reproduced without the proposed changes so I'm not sure if we should split this commit into a separate PR or handle here as well. Any suggestions are welcome!

indutny:
Opened a separate PR to track it: https://github.com/neon-bindings/neon/pull/744

crates/neon-runtime/src/napi/tsfn.rs
```diff
@@ -167,6 +179,12 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
 
 impl<T> Drop for ThreadsafeFunction<T> {
     fn drop(&mut self) {
+        // tsfn was already finalized by `Environment::CleanupHandles()` in
+        // Node.js
+        if *self.is_finalized.lock().unwrap() {
```
indutny-signal, line 184: I'm a bit worried about the cases where we might end up running `release` concurrently to `finalize` callback. Maybe this has to be fixed in Node.js instead...

kjvalencik:
I like the way this API is headed. One thing that I'm still not sure about is the asymmetry on `EventQueue` you get from `EventQueue::new` and ones you get from `cx.queue()`. The later shares a backing queue and the former doesn't, but it's not visible from the type signature and isn't super clear from usage.

Any ideas of how to make this more clear? I have a couple thoughts:
* `EventQueue` could implement `Clone` in a way that shares the backing queue. This could be a way for users to leverage sharing outside of the global queue and maybe make things feel more symmetrical.
* Introduce a new type for sharing. This is what [tokio does](https://docs.rs/tokio/1.6.1/tokio/runtime/struct.Handle.html), but increasing the API surface area might be confusing.

kjvalencik:
@indutny Don't feel the need to jump on an implementation. Right now I'm trying to get a discussion going. I don't have a clear picture of what the API _should_ be.

But, I also don't want to bikeshed too long because this optimization is very cool! 🚲  🏠

indutny-signal:
To confirm, nothing actionable for me right now? If yes - what's the plan for this PR?

kjvalencik:
Correct, nothing actionable. I'm going to think about the API a bit more and talk with @dherman while working on getting the other changes released.

Thanks!

dherman:
I like the idea of using `Clone` as a way to connect sharing to a user-visible operation, which then makes it a little easier to explain that the `cx` method always returns a clone of the same shared queue.

In #734 we experimented with renaming `EventQueue` to `Channel`, which maybe also helps here. `EventQueue` is a little mysterious since the queue is internal and the user of the API isn't actually directly interacting with it. I like the idea that you can create new channels or you can clone an existing one to share its internal queue, and `cx.channel()` gives you a cloned channel so in the common case you get the sharing by default.

dherman:
One more thought: I think just the one `Channel` type is sufficient even with cloning. It's reminiscent of `Rc` and `Arc`, where you have a single type with interesting connections under the hood between the original and its clones, but the user doesn't generally need to worry about the difference; they all act as the same type.

src/trampoline.rs
kjvalencik: Since they require `&mut Context`, it implies the operations are serialized. We could get away with a `Cell` instead.

But, that might be an optimization better left for when it's demonstrated to be a performance bottleneck.

indutny: Yeah, I don't imagine this being a bottleneck, and it is way too simple to do it with a mutex. I'm open to changing it, though.

kjvalencik:
@indutny Thanks again so much for this PR. I spent a good amount of time thinking about this and also discussing it with @dherman.

For both of us, what makes this all fit is `impl Clone for EventQueue`. This makes it semantically identical to things in Rust `std` that share a backing store when cloned (e.g., `std::sync::Arc`, `std::mpsc::Sender`). I think it also makes `cx.queue()` easy to understand--it's a *clone* of the globally cached queue.

There are a few changes I would like to make to this PR:

* We [renamed `EventQueue` to `Channel`](https://github.com/neon-bindings/neon/pull/752) (with deprecated aliases, to minimize the user impact). This (hopefully) more clearly illustrates what it is. This PR needs to be updated to use `Channel` everywhere.
* Implement `Clone` for `Channel`. This can't be derived, but should be fairly straight forward. Clone the fields and increment the reference count (if and only if the channel being cloned is referenced).
* Replace the `cx.channel()` implementation with a `clone()` + `reference(&mut cx)` of the `InstanceData` queue.
* Remove the `ThreadsafeTrampoline` abstraction. In my opinion, this makes things more complicated than necessary. It's a little easier to understand with the functionality inlined into `Channel`. Possibly with a `ChannelState` helper struct to store the data inside the `RwLock`.
* Add some additional documentation explaining the new behavior
* Document that if needed for performance the `RwLock` may be replaced with an `AtomicUsize` and a small amount of `unsafe` since `&mut cx` already enforces the necessary invariants.

I know this is asking a lot and you have already be incredibly generous with your time, so I am happy to take over this PR if you would like me to, but I can also help guide it through if you prefer.

Cheers!

indutny-signal:
@kjvalencik I'll happily look into it. Will let you know how it goes!

src/event/event_queue.rs
```diff
@@ -138,6 +162,57 @@ impl Channel {
...
+    fn drop(&mut self) {
+        // Not a referenced event queue
+        if !self.has_ref {
+            return;
+        }
+
+        // It was only us who kept the `ChannelState` alive. No need to unref
+        // the `tsfn`, because it is going to be dropped once this function
+        // returns.
+        if Arc::strong_count(&self.state) == 1 {
```
kjvalencik, line 203: This is a really great trick that's much lighter weight than locking the mutex!

There is a scenario where we *could* return early that this won't catch. If there are only unreferenced clones, the `Arc::strong_count` may be `> 0`, while `ref_count == 0`. This may be a likely scenario if a user creates one unreferenced `queue` that lives a long time, but mostly uses `cx.queue()` as a one-off.

Do you think this is a likely enough scenario that it's worth the extra cost of a `read` lock on the internal state?

indutny-signal: If `ref_count == 0` - we'll return early at the start of the function where we check `has_ref`. I've added another optimization for the case where there are more than one references to the shared state. PTAL.

kjvalencik:
I performed some benchmarks with this and unfortunately it performs worse in most scenarios and only marginally better in the ideal situation.

Poking around a bit, I think the primary issue is acquiring a lock to perform `Channel::try_send` contending with the lock in `Drop`.

Unfortunately, I think the best fix is a more complicated solution that uses an atomic for reference counting instead of a `RwLock`. It will share a very similar design to the internals of `Arc`. It would be nice if we could simply _use_ an `Arc` over a sentinel with a clever `Drop` implementation, but I haven't been able to think of a way to do that.

src/event/event_queue.rs
```diff
@@ -154,3 +173,140 @@ impl std::fmt::Debug for SendError {
...
 
 impl std::error::Error for SendError {}
+
+struct ChannelState {
+    shared: Arc<ChannelSharedState>,
+    has_ref: AtomicBool,
+}
+
+impl ChannelState {
+    fn reference<'a, C: Context<'a>>(self: &Self, cx: &mut C) {
```
jrose-signal, line 183: Nitpick: `self: &Self` is idiomatically written `&self`.

indutny-signal: Ack!

src/event/event_queue.rs
```diff
@@ -117,27 +127,68 @@ impl Channel {
...
+    fn drop(&mut self) {
+        // It was only us who kept the `ChannelState` alive. No need to unref
+        // the `tsfn`, because it is going to be dropped once this function
+        // returns.
+        if Arc::strong_count(&self.state) == 1 {
+            return;
+        }
+
+        // Not a referenced event queue
+        if !self.has_ref {
```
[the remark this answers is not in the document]
indutny-signal: Ack.

src/context/mod.rs
```diff
@@ -550,9 +552,18 @@ pub trait Context<'a>: ContextInternal<'a> {
     }
 
     #[cfg(all(feature = "napi-4", feature = "event-queue-api"))]
-    /// Creates an unbounded channel for scheduling events to be executed on the JavaScript thread.
+    /// Clones a shared unbounded channel for scheduling events to be executed on the JavaScript thread.
```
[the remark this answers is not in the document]
indutny-signal: Ack.

src/event/event_queue.rs
```diff
@@ -117,27 +127,68 @@ impl Channel {
...
+            return;
+        }
+
+        // The ChannelState is dropped on a worker thread. We have to `unref`
+        // the tsfn on the UV thread after all pending closures.
+        let state = Arc::clone(&self.state);
+
+        self.send(move |mut cx| {
```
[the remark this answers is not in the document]
indutny-signal: As discussed over Slack, there is a race condition that don't let us use this optimization here.

src/event/event_queue.rs
```diff
@@ -154,3 +205,49 @@ impl std::fmt::Debug for SendError {
...
+    fn new<'a, C: Context<'a>>(cx: &mut C) -> Self {
+        let tsfn = unsafe { ThreadsafeFunction::new(cx.env().to_raw(), Self::callback) };
+        Self {
+            tsfn,
+            ref_count: AtomicUsize::new(1),
+        }
+    }
+
+    fn reference<'a, C: Context<'a>>(&self, cx: &mut C) {
+        if self.ref_count.fetch_add(1, Ordering::Relaxed) != 0 {
```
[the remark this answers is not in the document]
indutny-signal: Added comment.

src/event/event_queue.rs
```diff
@@ -154,3 +205,49 @@ impl std::fmt::Debug for SendError {
...
+            return;
+        }
+
+        unsafe {
+            self.tsfn.reference(cx.env().to_raw());
+        }
+    }
+
+    fn unref<'a, C: Context<'a>>(&self, cx: &mut C) {
+        if self.ref_count.fetch_sub(1, Ordering::Relaxed) != 1 {
```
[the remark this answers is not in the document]
indutny-signal: Ditto.

PR #740, opened 2021-05-19
Top-level API docs for `neon::object`

dherman:
Top-level API docs for the `neon::object` module.

[preview](https://neon-object-apidocs.surge.sh/neon/object/index.html)

### DIFF
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -1,4 +1,36 @@
 //! Traits for working with JavaScript objects.
+//!
+//! This module defines the [`Object`](Object) trait, which is implemented
+//! by all object types in the [JavaScript type hierarchy][hierarchy]. This
+//! trait provides key operations in the semantics of JavaScript objects,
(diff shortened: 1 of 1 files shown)


src/object/mod.rs
```diff
@@ -1,4 +1,36 @@
...
+//! ## Property Keys
+//!
+//! Object properties are accessed by a _property key_, which in JavaScript
+//! can be a string or [symbol][symbol]. (Neon does not yet have support for
+//! symbols.) For convenience, the [`PropertyKey`](PropertyKey) trait allows
+//! Neon programs to use various Rust string types, as well as numeric types,
+//! as keys when accessing object properties, converting the keys to strings
+//! as necessary:
+//!
+//! ```
```
kjvalencik, line 17: This needs a `cfg` to skip it in non-N-API and the `get` and `set` functions need a context added.

dherman: Fixed and CI is passing again!

PR #744, opened 2021-05-20
Handle early finalization of TSFN by Node.js

indutny:
Thread-safe functions can get released early by Node.js itself during
the teardown of the Environment that holds them. When this happens -
`finalize_cb` of the thread-safe function is invoked to notify us. This
change makes us remember if the thread-safe function was finalized
before we got into the `Drop` implementation for it so that we don't
deallocate it twice.

Additionally, we should not invoke `call_threadsafe_function` after
either `finalize_cb` call or `napi_closing` return value from previous
`call_threadsafe_function`. Handle this by holding a mutex around the
call, which doesn't introduce any additional locking because
`call_threadsafe_function` itself holds another mutex while active.

See: https://github.com/neon-bindings/neon/pull/739#issuecomment-844588950

### DIFF
--- a/crates/neon-runtime/src/napi/tsfn.rs
+++ b/crates/neon-runtime/src/napi/tsfn.rs
@@ -2,6 +2,7 @@
 
 use std::ffi::c_void;
 use std::mem::MaybeUninit;
+use std::sync::{Arc, Mutex};
 
 use crate::napi::bindings as napi;
 use crate::raw::{Env, Local};
@@ -34,6 +35,7 @@ unsafe impl Sync for Tsfn {}
 /// function for scheduling tasks to execute on a JavaScript thread.
 pub struct ThreadsafeFunction<T> {
     tsfn: Tsfn,
+    is_finalized: Arc<Mutex<bool>>,
     callback: fn(Option<Env>, T),
 }
 
@@ -76,6 +78,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
         callback: fn(Option<Env>, T),
     ) -> Self {
         let mut result = MaybeUninit::uninit();
+        let is_finalized = Arc::new(Mutex::new(false));
 
         assert_eq!(
             napi::create_threadsafe_function(
@@ -87,8 +90,8 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
                 // Always set the reference count to 1. Prefer using
                 // Rust `Arc` to maintain the struct.
                 1,
-                std::ptr::null_mut(),
-                None,
+                Arc::into_raw(is_finalized.clone()) as *mut _,
+                Some(Self::finalize),
                 std::ptr::null_mut(),
                 Some(Self::callback),
                 result.as_mut_ptr(),
@@ -98,6 +101,7 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
 
         Self {
             tsfn: Tsfn(result.assume_init()),
+            is_finalized: is_finalized,
             callback,
         }
     }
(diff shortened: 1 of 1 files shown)


crates/neon-runtime/src/napi/tsfn.rs
```diff
@@ -34,6 +35,7 @@ unsafe impl Sync for Tsfn {}
 /// function for scheduling tasks to execute on a JavaScript thread.
 pub struct ThreadsafeFunction<T> {
     tsfn: Tsfn,
+    is_finalized: Arc<Mutex<bool>>,
```
kjvalencik, line 38: Can this be an [`AtomicBool`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html) instead of a `Mutex<bool>`? I'm also not clear why it needs to be an `Arc`.

indutny-signal: `Arc` provides a nice way of passing the pointer to the C function, because unlike `Box` you can clone it before turning into a raw pointer, and then decrement the reference count when turning it back from the raw pointer to the `Arc` instance and dropping it.

I think it could be `AtomicBool`. Let me try that in a moment.

indutny-signal: Moved to `AtomicBool`. @jrose-signal should be really happy to see this happen :wink:

jrose-signal:
I think you're right about the race in https://github.com/neon-bindings/neon/pull/739#discussion_r636366731. :-( A 99% multithreading fix can be worse than no fix, because it makes it harder to track down further issues.

crates/neon-runtime/src/napi/tsfn.rs
jrose-signal: Yeah, it has to be *something* indirect because ThreadsafeFunction is passed around by value. Box is *probably* safe because the finalizer callback is guaranteed to be invoked before `Drop::drop` completes (either Node invokes it on shutdown or we invoke it in `drop`), but it's a lot easier to mess that up somehow.

indutny:
Nevermind, I'm a fool. ThreadsafeFunction can only be destroyed from the main loop because it might hold a reference to a JS function. The [documentation](https://nodejs.org/api/n-api.html#n_api_asynchronous_thread_safe_function_calls) explicitly says:

> Aside from the main loop thread, no threads should be using the thread-safe function after the finalize callback completes.

Now, I've checked all code paths and it cannot crash as I previously thought. To prevent the race condition in this patch, however, we'll have to hold the mutex for the duration of the `drop()` method. I'll modify this PR accordingly.

indutny:
Alright, pushed an update. Sadly `AtomicBool` had to go. When `finalize()` and `drop()` are called concurrently - two different orderings are possible:

- `finalize()` gets to the mutex first (and returns) and thus `drop()` will see `is_finalized = true` and won't call release
- `drop()` holds the mutex and observes `is_finalized = false`, `finalize()` is waiting on mutex on the UV loop thread and does not return until `drop()` calls the `release()`. In this case `release()` will attempt to send a `uv_async_t` event to the UV loop, which won't ever trigger because right after `finalize()` returns Node.js will close the `uv_async_t` handle

In both of the cases we get the desired result. This PR is ready for a review again.

indutny:
Had to add a mutex to the `call()` method as well because we shouldn't be calling it after either of:

* getting `napi_closed` from a previous call
* `finalize_cb` was called

This is quite inconvenient for Neon, but would likely be needed until we'll get some resolution in https://github.com/nodejs/node/issues/38775

indutny:
Force pushed PR with a rebase over main branch and all changes squashed.

crates/neon-runtime/src/napi/tsfn.rs
```diff
@@ -167,6 +195,14 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
 
 impl<T> Drop for ThreadsafeFunction<T> {
     fn drop(&mut self) {
+        let is_finalized = self.is_finalized.lock().unwrap();
+
+        // tsfn was already finalized by `Environment::CleanupHandles()` in
+        // Node.js
+        if *is_finalized {
+            return;
+        }
```
jrose-signal, line 204: Don't you need to drop the lock before calling `napi::release_threadsafe_function`?

indutny-signal: Nope, actually I don't the idea here is that we want `finalize_cb` to block until `release_threadsafe_function` completes, because otherwise `finalize_cb` might finish first and delete the tsfn in C++.

jrose-signal: Mutex isn't re-entrant, though, so in the *non*-early-exit case you'll lock, call `release_threadsafe_function`, and fail when it's time to lock again.

indutny-signal: `release_threadsafe_function` doesn't call `finalize_cb` directly. It schedules an event on the loop thread and executes `finalize_cb` in another tick.

indutny-signal:
See: https://github.com/napi-rs/napi-rs/pull/522 . We are not the first to run into it! napi-rs took a milder approach of just leaking the function and letting Node.js reclaim it :joy:

kjvalencik:
After fixing a leak in https://github.com/neon-bindings/neon/pull/750, this started happening every time with the global queue.

I think I fully understand this fix now and appreciate it. Thanks!

indutny-signal:
No worries! Let's land the thing! :wink:

kjvalencik:
Thanks so much for this fix! I'm going to release it along with this PR: https://github.com/neon-bindings/neon/pull/750

Issue #746, opened 2021-05-27
A possible memory leak in callbacks/root


pimeys:
Hi,

I was evaluating the new N-API codebase, and found quite fast greatly leaking piece of code. Here's the test repo that will show the leak:

https://github.com/pimeys/hello-neon

Now, I did some digging with Valgrind:
![leak](https://user-images.githubusercontent.com/34967/119855009-da22e000-bf11-11eb-93c4-ee0ddc7a4c99.png)

As you can see the culprit is in `napi_create_reference`, called from `Root::new`. You can dig into the results from the included `massif` dump! I was not able to collect this data, even by forcing `global.gc()`, meaning we probably do something wrong in the native side.

I'd expect this loop to have a steady flat memory profile, the data should be collected in the young-space already...

pimeys:
```
n7: 6550942 (heap allocation functions) malloc/new/new[], --alloc-fns, etc.
      
      
         n1: 4627840 0x91FF79: napi_create_reference (in /nix/store/ncn3gm18lzqv4w3jrfpk5yqbw4swdz03-nodejs-slim-14.17.0/bin/node)
      
      
          n1: 4627840 0x9E84B63: neon_runtime::napi::bindings::functions::napi1::create_reference (mod.rs:163)
      
      
           n1: 4627840 0x9E7C52D: neon_runtime::napi::reference::new (reference.rs:11)
      
      
            n1: 4627840 0x9DD8727: neon::handle::root::Root&lt;T&gt;::new (root.rs:65)
      
      
             n1: 4627840 0x9DDD60F: neon::object::traits::Object::root (mod.rs:238)
      
      
              n1: 4627840 0x9DCF08D: hello_neon::QueryEngine::js_query (lib.rs:60)
      
      
               n1: 4627840 0x9DCB95A: &lt;neon::types::internal::FunctionCallback&lt;T&gt; as neon::types::internal::Callback&lt;*mut neon_runtime::napi::bindings::types::Value__&gt;&gt;::invoke::{{closure}}::{{closure}} (internal.rs:62)
      
      
                n1: 4627840 0x9DC5E82: neon::types::error::convert_panics::{{closure}} (error.rs:86)
      
      
                 n1: 4627840 0x9DE3404: std::panicking::try::do_call (panicking.rs:379)
      
      
                  n1: 4627840 0x9DE385C: __rust_try (in /home/pimeys/code/hello-neon/index.node)
      
      
[...]

PR #750, opened 2021-06-01
Memory leak in `Root`

kjvalencik:
Fixes a memory leak in `Root` reported in https://github.com/neon-bindings/neon/issues/746.

This PR includes two commits:

* Delete N-API references when the count hits `0`.
  This is the bug referenced above. It stems from a misunderstanding of N-API. It's not sufficient to decrement the reference count to `0`, the reference must also be deleted.
* Use an `Option` to store the reference instead of using `ManuallyDrop`.
   While not as extreme, the global event queue reference counter was always incrementing. It would never be freed, event after the context ended.

### DIFF
--- a/src/handle/root.rs
+++ b/src/handle/root.rs
@@ -1,6 +1,6 @@
 use std::ffi::c_void;
 use std::marker::PhantomData;
-use std::mem::ManuallyDrop;
+#[cfg(feature = "napi-6")]
 use std::sync::Arc;
 
 use neon_runtime::reference;
@@ -32,7 +32,9 @@ unsafe impl Sync for NapiRef {}
 /// A `Root<T>` may be sent across threads, but the referenced object may
 /// only be accessed on the JavaScript thread that created it.
 pub struct Root<T> {
-    internal: NapiRef,
+    // `Option` is used to skip `Drop` when `Root::drop` or `Root::into_inner` is used.
+    // It will *always* be `Some` when a user is interacting with `Root`.
+    internal: Option<NapiRef>,
     #[cfg(feature = "napi-6")]
     drop_queue: Arc<ThreadsafeFunction<NapiRef>>,
     _phantom: PhantomData<T>,
@@ -65,7 +67,7 @@ impl<T: Object> Root<T> {
         let internal = unsafe { reference::new(env, value.to_raw()) };
 
         Self {
-            internal: NapiRef(internal as *mut _),
+            internal: Some(NapiRef(internal as *mut _)),
             #[cfg(feature = "napi-6")]
             drop_queue: InstanceData::drop_queue(cx),
             _phantom: PhantomData,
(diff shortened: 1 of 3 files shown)


src/handle/root.rs
```diff
@@ -32,7 +32,7 @@ unsafe impl Sync for NapiRef {}
 /// A `Root<T>` may be sent across threads, but the referenced object may
 /// only be accessed on the JavaScript thread that created it.
 pub struct Root<T> {
-    internal: NapiRef,
+    internal: Option<NapiRef>,
```
dherman, line 35: It took me a while to understand the use of `Option` and convince myself the `unwrap()` is safe. It might help to just document the `Option` is only used to track whether the root has been explicitly dropped via `drop(cx)` /`into_inner()`.

src/handle/root.rs
```diff
@@ -130,10 +130,18 @@ impl<T: Object> Root<T> {
...
     pub fn to_inner<'a, C: Context<'a>>(&self, cx: &mut C) -> Handle<'a, T> {
         let env = cx.env();
-        let local = unsafe { reference::get(env.to_raw(), self.internal.0 as *mut _) };
+        let local = unsafe { reference::get(env.to_raw(), self.as_napi_ref().0 as *mut _) };
 
         Handle::new_internal(T::from_raw(env, local))
     }
+
+    fn as_napi_ref(&self) -> &NapiRef {
+        self.internal.as_ref().unwrap()
```
dherman, line 139: Probably worth a comment explaining that this method is only ever called when we know the root hasn't been explicitly dropped, so the `Option` is `Some`.

dherman (changes requested):
This looks great, it's just a little bit subtle and I don't 100% understand yet. How did the `ManuallyDrop` logic prevent the queue reference count from decrementing, and how does the `Option` fix that?

src/handle/root.rs
kjvalencik: That's a good idea. I kept trying to find a good way to implement this, but no matter what I did, it was complicated. `ManuallyDrop` require `unsafe` for _partial_ dropping, which is what we need.

Do you think it would be better with a `bool`? That would remove the `unwrap`.

dherman: No, I think the `unwrap` is fine with a comment.

kjvalencik:
> This looks great, it's just a little bit subtle and I don't 100% understand yet. How did the `ManuallyDrop` logic prevent the queue reference count from decrementing, and how does the `Option` fix that?

@dherman Since `Drop` is recursive, `ManuallyDrop` prevents the _fields_ from executing their `Drop` implementation as well. To drop the `Arc` would require some `unsafe`.

This changes the strategy so `ManuallyDrop` is _no longer used_. Instead, a `None` signals to the `Drop` implementation that nothing needs to be done. Since these fields are only used internally, the value will _always_ be `Some` when a user interacts with it. The methods that `take` the value and replace it with `None` *consume* the struct.

dherman (approved):
Looks great!

PR #753, opened 2021-06-05
docs: account for n-api rename to node-api, suggest new headline structure

janpio:
- https://nodejs.medium.com/renaming-n-api-to-node-api-27aa8ca30ed8
- Rework headline hierarchy


kjvalencik:
Sorry that I missed this! Thanks for the update!

Issue #502, opened 2020-03-23
Support the primitive type: `JsSymbol`(NEON) as `Symbol`(JS)


usagi:
I found a lacks to supporting to these primitive types in my working at the https://github.com/neon-bindings/examples/issues/51 :

1. `Symbol` -> This issue
2. `BigInt` -> #376 

---

Note: `Symbol` is not a frequently used primitive type. So this issue may have a lower priority. However, this is a necessary task for NEON as a language binding core feature.

kjvalencik:
As a workaround for first class support, since `Symbol` is not a primitive, it can be used by accessing it on the global object.

```rs
fn symbol(mut cx: FunctionContext) -> JsResult<JsValue> {
    let global = cx.global();
    let symbol_ctor = global
        .get(&mut cx, "Symbol")?
        .downcast::<JsFunction>()
        .or_throw(&mut cx)?;
    
    let symbol_label = cx.string("MySymbol");
    let my_symbol = symbol_ctor.call(&mut cx, global, vec![symbol_label])?;

    Ok(my_symbol)
}
```

usagi:
Thank you @kjvalencik , your alternative technique is useful for the current implementation level of `NEON`. 👍

But, I have one disagreement. **`Symbol` is a primitive type.** My souce is ECMA-262 and MDN. And supplementary Node.js:

- ECMA-262 "§4.2 ECMAScript Overview", "§19.4 Symbol Objects"
    - <https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-262.pdf>
    - <https://tc39.es/ecma262/#sec-symbol-objects>
- MDN "Primitive" <https://developer.mozilla.org/en-US/docs/Glossary/Primitive>
- Node.js N-API "napi_create_symbol" and etc. <https://nodejs.org/api/n-api.html#n_api_napi_create_symbol>

I think, `JsSymbol` supporting is good things for a API structure unifying and completeness. Of cource, I wrote "So this issue may have a lower priority" in the previous note. However, it should be at least including in the implementation plan even if the priority is actually very low.

kjvalencik:
![](https://media1.giphy.com/media/14bDMRUYVrzOIo/giphy.gif)

I stand corrected, that Symbol _is_ a primitive type. That's good to know! Also, I didn't mean to indicate that this _shouldn't_ be done, I only wanted to provide a workaround in the feature's absense.

PR #761, opened 2021-06-25
JsSymbol primitive

chrisbajorin:
Addresses #502 and [JsSymbol RFC](https://github.com/neon-bindings/rfcs/pull/38). 

A few thoughts from playing with the implementation:

In JS, the `Symbol()` [function spec](https://tc39.es/ecma262/#sec-symbol-description) allows for non-string values that can be coerced to string, e.g. 
`Symbol(null).description === 'null'`.

Should `JsSymbol::with_description()` accept some trait that allows it to accept both `Handle` and `&str` (along the lines of `PropertyKey`)? I don't know how other folks utilize the description property. When using them in JS, I personally only use string literals since I use the description to convey meaning to other developers in the source code. I don't think I've ever used the description itself as a value, but I could see someone else using it as such.

Along that same line of thought, if people do use the description as a value, should `JsSymbol::description()` return an `Option<Handle<'a, JsString>>` instead of `Option<String>` (or even `Handle<'a, JsValue>` that could be either `JsUndefined` or `JsString`)?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,10 @@ event-queue-api = []
 # Feature flag to include procedural macros
 proc-macros = ["neon-macros"]
 
+# Feature flag to enable the `JsSymbol` API of RFC 38
+# https://github.com/neon-bindings/rfcs/pull/38
+symbol-primitive-api = []
+
 [package.metadata.docs.rs]
 features = ["docs-only", "event-handler-api", "proc-macros", "try-catch-api"]
 
--- a/crates/neon-runtime/src/napi/primitive.rs
+++ b/crates/neon-runtime/src/napi/primitive.rs
@@ -43,3 +43,9 @@ pub unsafe fn number_value(env: Env, p: Local) -> f64 {
     );
     value
 }
+
+/// Mutates the `out` argument provided to refer to a newly created `Local` containing a
+/// JavaScript symbol.
+pub unsafe fn symbol(out: &mut Local, env: Env, desc: Local) {
+    napi::create_symbol(env, desc, out as *mut Local);
+}
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -67,7 +67,8 @@
 //!     of custom objects that own Rust data structures.
 //! - **Primitive types:** These are the built-in JavaScript datatypes that are not
 //!   object types: [`JsNumber`](JsNumber), [`JsBoolean`](JsBoolean),
-//!   [`JsString`](JsString), [`JsNull`](JsNull), and [`JsUndefined`](JsUndefined).
+//!   [`JsString`](JsString), [`JsNull`](JsNull), [`JsSymbol`](JsSymbol),
+//!    and [`JsUndefined`](JsUndefined).
 //!
 //! [types]: https://raw.githubusercontent.com/neon-bindings/neon/main/doc/types.jpg
 //! [unknown]: https://mariusschulz.com/blog/the-unknown-type-in-typescript#the-unknown-type
@@ -80,6 +81,8 @@ pub(crate) mod date;
 pub(crate) mod error;
 
 pub(crate) mod internal;
+#[cfg(all(feature = "napi-1", feature = "symbol-primitive-api"))]
+pub(crate) mod symbol;
 pub(crate) mod utf8;
 
 use self::internal::{FunctionCallback, ValueInternal};
@@ -434,16 +437,19 @@ impl JsString {
 
     #[cfg(feature = "napi-1")]
     pub fn value<'a, C: Context<'a>>(self, cx: &mut C) -> String {
-        let env = cx.env().to_raw();
+        unsafe { JsString::value_internal(cx.env().to_raw(), self.to_raw()) }
+    }
 
-        unsafe {
-            let capacity = neon_runtime::string::utf8_len(env, self.to_raw()) + 1;
-            let mut buffer: Vec<u8> = Vec::with_capacity(capacity as usize);
-            let p = buffer.as_mut_ptr();
-            std::mem::forget(buffer);
-            let len = neon_runtime::string::data(env, p, capacity, self.to_raw());
-            String::from_raw_parts(p, len as usize, capacity as usize)
-        }
+    /// Invariants:
+    /// - `local` has an underlying value type of `napi::ValueType::String`
+    #[cfg(feature = "napi-1")]
+    pub(crate) unsafe fn value_internal(env: raw::Env, local: raw::Local) -> String {
+        let capacity = neon_runtime::string::utf8_len(env, local) + 1;
+        let mut buffer: Vec<u8> = Vec::with_capacity(capacity as usize);
+        let p = buffer.as_mut_ptr();
+        std::mem::forget(buffer);
+        let len = neon_runtime::string::data(env, p, capacity, local);
+        String::from_raw_parts(p, len as usize, capacity as usize)
     }
 
     pub fn new<'a, C: Context<'a>, S: AsRef<str>>(cx: &mut C, val: S) -> Handle<'a, JsString> {
--- a/src/types/symbol.rs
+++ b/src/types/symbol.rs
@@ -1,0 +1,86 @@
+use crate::context::Context;
+use crate::handle::{Handle, Managed};
+use crate::types::internal::ValueInternal;
+use crate::types::utf8::Utf8;
+use crate::types::{Env, JsString, Value};
+
+use neon_runtime::raw;
+
+/// A JavaScript symbol primitive value.
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct JsSymbol(raw::Local);
+
+impl JsSymbol {
+    /// Create a new symbol.
+    /// Equivalent to calling `Symbol()` in JavaScript
+    pub fn new<'a, C: Context<'a>>(cx: &mut C) -> Handle<'a, JsSymbol> {
+        JsSymbol::new_internal(cx.env(), None::<&str>)
+    }
+
+    /// Create a new symbol with a description.
+    /// Equivalent to calling `Symbol(desc)` in JavaScript
+    pub fn with_description<'a, C: Context<'a>, S: AsRef<str>>(
+        cx: &mut C,
+        desc: S,
+    ) -> Handle<'a, JsSymbol> {
+        JsSymbol::new_internal(cx.env(), Some(desc))
+    }
+
+    /// Get the optional symbol description, where `None` represents an undefined description.
+    pub fn description<'a, C: Context<'a>>(self, cx: &mut C) -> Option<String> {
+        let env = cx.env().to_raw();
+        let (desc_ptr, desc_len) = Utf8::from("description").into_small_unwrap().lower();
+
+        unsafe {
+            let mut local = std::mem::zeroed();
+            if !neon_runtime::object::get_string(env, &mut local, self.to_raw(), desc_ptr, desc_len)
+            {
+                return None;
+            }
+
+            if neon_runtime::tag::is_string(env, local) {
+                Some(JsString::value_internal(env, local))
+            } else {
+                None
+            }
+        }
+    }
+
+    pub(crate) fn new_internal<'a, S: AsRef<str>>(
+        env: Env,
+        desc: Option<S>,
+    ) -> Handle<'a, JsSymbol> {
+        unsafe {
+            let desc_local = desc
+                .and_then(|d| JsString::new_internal(env, d.as_ref()))
+                .map_or_else(|| std::mem::zeroed(), |h| h.to_raw());
+
+            let mut sym_local = std::mem::zeroed();
+            neon_runtime::primitive::symbol(&mut sym_local, env.to_raw(), desc_local);
+            Handle::new_internal(JsSymbol(sym_local))
+        }
+    }
+}
+
+impl Value for JsSymbol {}
+
+impl Managed for JsSymbol {
+    fn to_raw(self) -> raw::Local {
+        self.0
+    }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self {
+        JsSymbol(h)
+    }
+}
+
+impl ValueInternal for JsSymbol {
+    fn name() -> String {
+        "symbol".to_string()
+    }
+
+    fn is_typeof<Other: Value>(env: Env, other: Other) -> bool {
+        unsafe { neon_runtime::tag::is_symbol(env.to_raw(), other.to_raw()) }
+    }
+}
--- a/test/napi/lib/objects.js
+++ b/test/napi/lib/objects.js
@@ -22,6 +22,15 @@ describe('JsObject', function() {
     assert.deepEqual({number: 9000, string: 'hello node'}, addon.return_js_object_with_mixed_content());
   });
 
+  it('return a JsObject with a symbol property key', function () {
+    const obj = addon.return_js_object_with_symbol_property_key();
+    const propertySymbols = Object.getOwnPropertySymbols(obj);
+    assert.equal(propertySymbols.length, 1);
+    const sym = propertySymbols[0];
+    assert.equal(typeof sym, "symbol");
+    assert.equal(obj[sym], sym);
+  })
+
   it('gets a 16-byte, zeroed ArrayBuffer', function() {
     var b = addon.return_array_buffer();
     assert.equal(b.byteLength, 16);
--- a/test/napi/lib/types.js
+++ b/test/napi/lib/types.js
@@ -70,6 +70,15 @@ describe('type checks', function() {
     assert(!addon.is_string(new String('1')));
   });
 
+  it('is_symbol', function () {
+    assert(addon.is_symbol(Symbol()));
+    assert(addon.is_symbol(Symbol("unique symbol")));
+    assert(addon.is_symbol(Symbol.for('neon:description')));
+    assert(addon.is_symbol(Symbol.toStringTag));
+    assert(!addon.is_symbol(undefined));
+    assert(!addon.is_symbol("anything other than symbol"));
+  });
+
   it('is_undefined', function () {
     assert(addon.is_undefined(undefined));
     assert(!addon.is_undefined(null));
@@ -84,5 +93,9 @@ describe('type checks', function() {
     assert(addon.strict_equals(o1, o1));
     assert(!addon.strict_equals(o1, o2));
     assert(!addon.strict_equals(o1, 17));
+    let s1 = Symbol();
+    let s2 = Symbol();
+    assert(addon.strict_equals(s1, s1));
+    assert(!addon.strict_equals(s1, s2));
   });
 });
--- a/test/napi/src/js/objects.rs
+++ b/test/napi/src/js/objects.rs
@@ -31,6 +31,13 @@ pub fn return_js_object_with_string(mut cx: FunctionContext) -> JsResult<JsObjec
     Ok(js_object)
 }
 
+pub fn return_js_object_with_symbol_property_key(mut cx: FunctionContext) -> JsResult<JsObject> {
+    let js_object: Handle<JsObject> = cx.empty_object();
+    let s = cx.symbol("neon:description");
+    js_object.set(&mut cx, s, s)?;
+    Ok(js_object)
+}
+
 pub fn return_array_buffer(mut cx: FunctionContext) -> JsResult<JsArrayBuffer> {
     let b: Handle<JsArrayBuffer> = cx.array_buffer(16)?;
     Ok(b)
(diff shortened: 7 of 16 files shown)


kjvalencik:
@chrisbajorin I might not have a chance to review the code right away, but I wanted to answer some of your questions with my thoughts quickly:

> Should `JsSymbol::with_description()` accept some trait that allows it to accept both Handle and `&str `(along the lines of `PropertyKey`)?

This is a good idea. It would be nice if a user could directly use either a `AsRef<str>` or a `JsString`. If we only pick one, I would probably favor `JsString` since it's slightly more flexible. The downside of this is that because of ownership on `Context` it would require two lines of code instead of a single line and that's not very ergonomic.

```rust
// Compiler error because `Context` is still borrowed by `cx.string` when calling `with_description`
JsSymbol::with_description(&mut cx, cx.string("MySymbol"));

let description = cx.string("MySymbol");
JsSymbol::with_description(&mut cx, description);
```

But, since I can't think of a use case for frequently creating symbols or using very large descriptions, the overhead of converting to `&str` and back, might be fine. What do you think?

If we do have an API that can take a handle, I *do* think it should be restricted to just `JsString` and not all `JsValue` for a couple of reasons:

* Users can always manually perform the coercion: https://docs.rs/neon/0.8.3/neon/types/trait.Value.html#method.to_string
* Coercion may throw an exception and that would require making this API failable. This would be awkward ergonomics since in most cases it will never fail.

> Along that same line of thought, if people do use the description as a value, should `JsSymbol::description()` return an `Option<Handle<'a, JsString>>` instead of `Option<String>` (or even `Handle<'a, JsValue>` that could be either `JsUndefined` or `JsString`)?

I like returning `Option<Handle<'a, JsString>>`. It's lower overhead if the conversion isn't necessary and can still be chained with `.map(|v| v.value(&mut cx))` to get an `Option<String>`. I think it's valuable to perform the `undefined` check upfront to give a more constrained type than `JsValue` that needs to be downcast.

chrisbajorin:
> It would be nice if a user could directly use either a `AsRef<str>` or a `JsString`.

One solution is to have `JsSymbol::with_description()` accept `Handle<JsString>`, but have `cx.symbol()` take `AsRef<str>`. The two-liner could then just be the implementation for `cx.symbol()` itself:
```rust
    fn symbol<S: AsRef<str>>(&mut self, desc: S) -> Handle<'a, JsSymbol> {
        let description = self.string(desc);
        JsSymbol::with_description(self, description)
    }
```
However, this would be a deviation from the rest of the context helper methods as direct wrappers around the underlying type constructor signature. That being said, because of the optional nature of the description property, it's already a deviation anyway.

Alternatively, have both functions accept something that implements `SymbolDescription`:
```rust
pub trait SymbolDescription {
    fn to_description_local(self, env: Env) -> Option<raw::Local>;
}

impl <'a>SymbolDescription for &'a str {
    fn to_description_local(self, env: Env) -> Option<raw::Local> {
        JsString::new_internal(env, self).map(|s| s.to_raw())
    }
}

impl <'a>SymbolDescription for Handle<'a, JsString> {
    fn to_description_local(self, _: Env) -> Option<raw::Local> {
        Some(self.to_raw())
    }
}
```
Although after writing it up, I'm not sure if it's needless complexity just to avoid writing that extra line of code (and this naive implementation would silently ignore a `StringOverflow` by converting it to an undefined description).

Glancing through my own JS code, I can't find a single instance where I'm dynamically generating unique symbols. Given that symbols can't be converted to `NapiRef` ([although this may change](https://github.com/nodejs/node/issues/39131)), I suspect that I'll likely still be declaring my library symbols in index.js, and passing them into a neon function whenever they are needed. Within that context, my primary motivation for the PR is definitely for the typeguard, so I can't say I have strong opinions on the constructor interface.

> If we do have an API that can take a handle, I do think it should be restricted to just `JsString` and not all `JsValue`

That makes sense.

> I like returning `Option<Handle<'a, JsString>>`

As do I. It would also allow me to revert the my edit to `JsString`, making the feature purely additive.

kjvalencik:
I really like your suggestion of having some asymmetry between `cx.symbol(..)` and `JsSymbol`. In general, the `Context` methods are meant to be the most _common_ constructors and not necessarily one-to-one with `new`.

That seems like a really good balance between flexibility and convenience.

Not being able to put a symbol in a persistent reference is unfortunate. That does limit the usability quite a bit but, I can't think of any way around it without it changing upstream.

src/types/mod.rs
```diff
@@ -80,6 +81,8 @@ pub(crate) mod date;
 pub(crate) mod error;
 
 pub(crate) mod internal;
+#[cfg(all(feature = "napi-1", feature = "symbol-primitive-api"))]
+pub(crate) mod symbol;
```
kjvalencik, line 85: Can you add a `use symbol::Symbol`? Currently it's not part of the `types` module.

src/types/symbol.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    pub(crate) fn new_internal<'a>(
+        env: Env,
+        desc: Option<Handle<'a, JsString>>,
+    ) -> Handle<'a, JsSymbol> {
+        unsafe {
+            let desc_local = match desc {
+                None => std::mem::zeroed(),
```
kjvalencik, line 56: Technically this is valid because it's only a wrapper over a raw pointer, but this wouldn't be robust against future changes.

Instead can we use `ptr::null()` or similar?

crates/neon-runtime/src/napi/primitive.rs
```diff
@@ -43,3 +43,9 @@ pub unsafe fn number_value(env: Env, p: Local) -> f64 {
     );
     value
 }
+
+/// Mutates the `out` argument provided to refer to a newly created `Local` containing a
+/// JavaScript symbol.
+pub unsafe fn symbol(out: &mut Local, env: Env, desc: Local) {
```
kjvalencik, line 49: Please remove the out pointer. That pattern is only used for exiting methods that require compatibility between the two backends. Since this is a new API, it can follow the new pattern of returning the type directly.

crates/neon-runtime/src/napi/primitive.rs
```diff
@@ -43,3 +43,9 @@ pub unsafe fn number_value(env: Env, p: Local) -> f64 {
...
+    napi::create_symbol(env, desc, out as *mut Local);
+}
```
kjvalencik, line 51: This method needs to validate `napi_status`.

```suggestion
pub unsafe fn symbol(env: Env, desc: Local) -> Local {
    let local = MaybeUninit::uninit();
    
    assert_eq!(napi::create_symbol(env, desc, local.as_mut_ptr()), napi::Status::Ok);
    
    local.assume_init();
}
```

src/types/symbol.rs
```diff
@@ -0,0 +1,86 @@
...
+        cx: &mut C,
+        desc: Handle<'a, JsString>,
+    ) -> Handle<'a, JsSymbol> {
+        JsSymbol::new_internal(cx.env(), Some(desc))
+    }
+
+    /// Get the optional symbol description, where `None` represents an undefined description.
+    pub fn description<'a, C: Context<'a>>(self, cx: &mut C) -> Option<Handle<'a, JsString>> {
+        let env = cx.env().to_raw();
+        let (desc_ptr, desc_len) = Utf8::from("description").into_small_unwrap().lower();
```
kjvalencik, line 33: This would probably be a little simpler if we added `napi_get_named_property` to `neon-runtime`. Usually strings are user provided so we can't be guaranteed they don't contain null, but this one could be a constant.

```rust
const DESCRIPTION_KEY: &[u8] = b"description\0";
```

Cargo.toml
```diff
@@ -75,6 +75,10 @@ event-queue-api = []
 # Feature flag to include procedural macros
 proc-macros = ["neon-macros"]
 
+# Feature flag to enable the `JsSymbol` API of RFC 38
+# https://github.com/neon-bindings/rfcs/pull/38
+symbol-primitive-api = []
```
kjvalencik, line 80: I think this could be `symbol-api`.

test/napi/lib/objects.js
```diff
@@ -22,6 +22,15 @@ describe('JsObject', function() {
     assert.deepEqual({number: 9000, string: 'hello node'}, addon.return_js_object_with_mixed_content());
   });
 
+  it('return a JsObject with a symbol property key', function () {
+    const obj = addon.return_js_object_with_symbol_property_key();
+    const propertySymbols = Object.getOwnPropertySymbols(obj);
+    assert.equal(propertySymbols.length, 1);
+    const sym = propertySymbols[0];
```
kjvalencik, line 29:
```suggestion
    const [sym] = propertySymbols;
```

test/napi/lib/objects.js
```diff
@@ -22,6 +22,15 @@ describe('JsObject', function() {
...
+    assert.equal(typeof sym, "symbol");
```
kjvalencik, line 30:
```suggestion
    assert.typeOf(sym, "symbol");
```

test/napi/lib/types.js
```diff
@@ -70,6 +70,15 @@ describe('type checks', function() {
     assert(!addon.is_string(new String('1')));
   });
 
+  it('is_symbol', function () {
+    assert(addon.is_symbol(Symbol()));
+    assert(addon.is_symbol(Symbol("unique symbol")));
+    assert(addon.is_symbol(Symbol.for('neon:description')));
+    assert(addon.is_symbol(Symbol.toStringTag));
```
kjvalencik, line 77: Can you add a test for a built in symbol like `Symbol.iterator`?

src/types/symbol.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Equivalent to calling `Symbol(description)` in JavaScript
+    pub fn with_description<'a, C: Context<'a>>(
+        cx: &mut C,
+        desc: Handle<'a, JsString>,
+    ) -> Handle<'a, JsSymbol> {
+        JsSymbol::new_internal(cx.env(), Some(desc))
+    }
+
+    /// Get the optional symbol description, where `None` represents an undefined description.
+    pub fn description<'a, C: Context<'a>>(self, cx: &mut C) -> Option<Handle<'a, JsString>> {
```
kjvalencik, line 31: I'd prefer if this were built into Node-API, but looks like it isn't. The best I can tell reading the ECMAScript spec, this property is guaranteed to exist and it's impossible to overwrite. @dherman is that correct?

test/napi/src/js/objects.rs
```diff
@@ -31,6 +31,13 @@ pub fn return_js_object_with_string(mut cx: FunctionContext) -> JsResult<JsObjec
     Ok(js_object)
 }
 
+pub fn return_js_object_with_symbol_property_key(mut cx: FunctionContext) -> JsResult<JsObject> {
+    let js_object: Handle<JsObject> = cx.empty_object();
```
kjvalencik, line 35: The type is unnecessary.

```suggestion
    let js_object = cx.empty_object();
```

test/napi/src/js/symbols.rs
```diff
@@ -0,0 +1,22 @@
+use neon::prelude::*;
+
+pub fn return_js_symbol_with_description(mut cx: FunctionContext) -> JsResult<JsSymbol> {
+    Ok(cx.symbol("neon:description"))
+}
+
+pub fn return_js_symbol(mut cx: FunctionContext) -> JsResult<JsSymbol> {
```
kjvalencik, line 7: Can you add a test for `JsSymbol::with_description` as well? Probably taking a `JsString` as an argument.

test/napi/src/js/symbols.rs
```diff
@@ -0,0 +1,22 @@
...
+    Ok(JsSymbol::new(&mut cx))
+}
+
+pub fn read_js_symbol_description(mut cx: FunctionContext) -> JsResult<JsValue> {
+    let symbol: Handle<JsSymbol> = cx.argument(0)?;
+    match symbol.description(&mut cx) {
```
kjvalencik, line 13: FYI, this can also be done with combinators.

```suggestion
    symbol
        .description()
        .map(|v| Ok(v.upcast()))
        .unwrap_or_else(|| Ok(cx.undefined()))
```

kjvalencik (commented):
Thanks so much for working on this feature! I've left some comments, but if you want to wait for the design to solidify before making changes, that's totally reasonable!

Cheers!

dherman:
Maybe a workaround (at least until/unless they fix `NapiRef` to work with symbols) would be for `impl<T: JsSymbol> Root<T>` to have specialized logic to wrap the symbol in a wrapper object before putting it in the `NapiRef`? It's maybe a little tacky since it just duck-types the methods of `impl<T: Object> Root<T>` as a kind of ad-hoc specialization.

kjvalencik:
@dherman I really like that idea and it should work. We can always optimize it later without making any breaking changes to the Neon API.

test/napi/lib/types.js
chrisbajorin: [Symbol.toStringTag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) is one of the well-knowns, but I'm happy to swap it for one that is more obvious at-a-glance.

src/types/symbol.rs
chrisbajorin: FYI, when I first wrote up the RFC, node@10 was still in maintenance LTS, and symbol.prototype.description was [unsupported](https://node.green/#ES2019-misc-Symbol-prototype-description).  With this implementation, if I downgrade to 10.24, the description tests fail both the node and rust getter with:

```
  1) JsSymbol
       should return a JsSymbol with a description built in Rust:
     AssertionError: expected undefined to equal 'neon:description'
      at Context.<anonymous> (lib/symbols.js:8:16)

  2) JsSymbol
       should read the description property in Rust:
     AssertionError: expected undefined to equal 'neon:description'
      at Context.<anonymous> (lib/symbols.js:18:16)
```
 `neon_runtime::tag::is_string(env, local)` returns false in the description function, so it returns `None`.

crates/neon-runtime/src/napi/primitive.rs
chrisbajorin: For this part I was definitely playing the follow-the-pattern game. Is there a reason why the other create primitives aren't validated? I can understand `null`/`undefined`/`boolean` as they are effectively just getters for global values, but `number` doesn't have any validation.

chrisbajorin:
In order to keep this syntax consistent:
```rust
let root1 = cx.argument::<JsObject>(0)?.root(&mut cx);
let root2 = cx.argument::<JsSymbol>(1)?.root(&mut cx);
```

the `Object`/`PropertyKey` API is fallible, so the `Root::new()` function would have to panic if `set_from` fails on the object wrapper:

```rust
impl Root<JsSymbol> {

    const PROPERTY_KEY: &'static str = "internal";
    
    pub fn new<'a, C: Context<'a>>(cx: &mut C, symbol: &JsSymbol) -> Self {
        let value = JsObject::new_internal(cx.env());
        unsafe {
            let mut result = false;
            if !Root::PROPERTY_KEY.set_from(cx, &mut result, value.to_raw(), symbol.to_raw()) {
                // panic?
            }
        }

        let env = cx.env().to_raw();
        let internal = unsafe { reference::new(env, value.to_raw()) };

        Self {
            internal: Some(NapiRef(internal as *mut _)),
            #[cfg(feature = "napi-6")]
            drop_queue: InstanceData::drop_queue(cx),
            _phantom: PhantomData,
        }
    }
}
```
Alternatively, we could make the JsSymbol root implementations all fallible, but this expands the API, and I suspect you'd have to point folks to this discussion just to make sense of it:
```rust
impl JsSymbol {
    // ...
    pub fn try_root<'a, C: Context<'a>>(&self, cx: &mut C) -> NeonResult<Root<Self>>
}

impl Root<JsSymbol> {

    pub fn try_new<'a, C: Context<'a>>(cx: &mut C, symbol: &JsSymbol) -> NeonResult<Self>
  
    pub fn clone<'a, C: Context<'a>>(&self, cx: &mut C) -> Self
    
    pub fn drop<'a, C: Context<'a>>(self, cx: &mut C)

    pub fn try_into_inner<'a, C: Context<'a>>(self, cx: &mut C) -> JsResult<'a, JsSymbol>

    pub fn try_to_inner<'a, C: Context<'a>>(&self, cx: &mut C) -> JsResult<'a, JsSymbol>
}
```

crates/neon-runtime/src/napi/primitive.rs
kjvalencik: JsNumber should be validating it since it's effectively infallible due to protections on the Rust side.

Thinking about this one a bit more, it should probably return something to indicate an error since there are valid cases where it could `throw`.

Maybe check if it's throwing and return None and _then_ assert that it's ok.

test/napi/lib/types.js
kjvalencik: Sorry just missed this!

src/types/symbol.rs
kjvalencik: The good news is we no longer support Node 10 and can make that simplification.

kjvalencik:
It should be safe to panic there. The only way these API can fail is if the key can't be allocated because it exceeds the max length (we statically know the length and this won't happen) or if there is a getter defined that throws. We also know that can't happen.

crates/neon-runtime/src/napi/primitive.rs
chrisbajorin: I'm not following. I understand this part:
> JsNumber should be validating it since it's effectively infallible due to protections on the Rust side.

In `neon_runtime::primitive::number()`, the value passed in has to be an `f64`, so it's inherently validated by the Rust type system.

While in `neon_runtime::primitive::symbol()`, `desc` has to be either a null pointer or a `Local` representing a `ValueType::String`. Adding the status check makes sense to me given that someone could pass in a different `Local` type, but I don't understand this part:
> Thinking about this one a bit more, it should probably return something to indicate an error since there are valid cases where it could throw.

What does this have to do with `throw`? If we're in a throwing state, and this function returns `Option<Local>` We'd have to expect/unwrap it in `JsSymbol::new_internal()` in order to return `Handle<JsSymbol>`. If we then unwrap the `None`, that panic would override the current throw state which seems to defeat the point. 

If it's the case that you're mistaking this function for the description getter, that one currently returns `None` when in a throwing state:
```rust
    let sym = cx.symbol("description");
    let opt1 = sym.description(&mut cx);
    assert!(opt1.is_some());
    cx.throw_type_error::<_, ()>("entering throw state with TypeError").unwrap_err();
    let opt2 = sym.description(&mut cx);
    assert!(opt2.is_none());
```
Although this is a side effect of `napi::get_property()` failing in `neon_runtime::object::get_string()` rather than explicitly checking for the throw state in `JsSymbol::description()`.

kjvalencik: You're right! I was thinking this method accepts a `str` as a description and converting to `JsString` could fail, but it's already a `napi_value` here. 🤦 

I agree, checking the status shouldn't be necessary. It might not hurt to check in a `debug_assert`.

chrisbajorin:
Relating to our discussion about throwing, It seems like `napi::create_reference()` succeeds while in a throw state, whereas the object property setters return `Status::PendingException`. So if your code is:
```rust
    let obj: Handle<JsObject> = cx.argument(0)?;
    cx.throw_type_error::<_, ()>("entering a throwing state with TypeError").unwrap_err();
    let root = obj.root(&mut cx);
```
this would throw a TypeError, but if you replace that `JsObject` with `JsSymbol`, then the TypeError would be overridden by whatever panic is in that property setter in `Root::<JsSymbol>::new()`.

kjvalencik:
@chrisbajorin This is because most things will `panic` if in a throwing state and the unwind gets caught and converted back to an exception. It's an unfortunate compromise in the design due to limitations in Rust.

Ideally, it would be statically impossible to call methods that can't be called while in a throwing state. However, to do that gets really unergonomic because `Context` would need to be threaded through the `Ok` results:

```rust
let (cx, obj) = cx.argument::<JsObject>(0)?;
```

Instead, Neon simply suggests that users are careful never to try to use `Context` when in a throwing state. Another option would be to pre-emptively return `Err(Throw)` whenever we are already throwing, but then that makes nearly every single API return `Result` when most are infallible.

In retrospect, it may have been better for all APIs to return `Result`, but that's not an API change we would take likely.

tl;dr -- What you shared is expected.

kjvalencik:
@chrisbajorin Sorry for letting this sit for so long! It had slipped my mind. Are you still interested in finishing this? I think it's pretty close.

chrisbajorin:
I'm terribly sorry about that, I was moving last year and my computer sat in a box for a few months. This completely slipped my mind. I'll take a look this weekend and see if I can get it finished up.

kjvalencik:
Let me know if you have any questions about the conflicts. A bunch of files moved and were changed when I deleted the legacy backend. Cheers!

PR #763, opened 2021-07-01
Hide deprecated APIs from the API docs.

dherman:


kjvalencik:
@dherman I'm not sure declare_types and task should be hidden on legacy since they are the only ways ro accomplish certain things on legacy.

dherman:
I think I was just building the docs with the wrong command-line flags. Closing this.

PR #770, opened 2021-07-23
v0.9.0

kjvalencik:
This includes a few odds and ends in addition to the v0.9.0 tag:

* Changes the default backend for `docs.rs` to Node-API
* Add `docsrs` attributes to mark items that require feature flags (e.g., `try_catch`)
* Rename the `event-queue-api` feature flag to `channel-api` (keeping the former as an alias)


dherman (approved):
Looks good! I’m excited to update docs.rs and get 0.9 out. One tiny nitpick but it’s up to you.

Issue #782, opened 2021-08-30
Building with Electron 13 does not work on Windows


andiliu-gh:
It seems as if building with Electron 13 is broken on Windows due to this: https://github.com/electron/electron/issues/29893

Error from neon:
```
    neon.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: class std::shared_ptr<class v8::BackingStore> __thiscall v8::ArrayBuffer::GetBackingStore(void)" (__imp_?GetBackingStore@ArrayBuffer@v8@@QAE?AV?$shared_ptr@VBackingStore@v8@@@std@@XZ) referenced in function _Neon_ArrayBuffer_Data [C:\Users\andyl\-\target\release\build\neon-sys-ab7d83299aae5b27\out\native\build\neon.vcxproj]
    C:\Users\-\target\release\build\neon-sys-ab7d83299aae5b27\out\native\build\Release\neon.node : fatal error LNK1120: 1 unresolved externals [C:\Users\-\target\release\build\neon-sys-ab7d83299aae5b27\out\native\build\neon.vcxproj]
```

I think it's cause neon uses nan or that it builds with msvc...? So the header signature is mismatched. The Electron issue link above mentions building with libc++ to avoid this issue or to replace nan.h w N-API?

kjvalencik:
I would be willing to merge a fix for this, but none of the core devs are Windows developers which would make this difficult.

However, Neon with the `napi` backend (the default when creating a project with `npm init neon my-project`) does not have this issue. That backend already uses Node-API (formerly N-API).

Since the `legacy` backend (`neon-sys`) is no longer being actively developed, it is unlikely that this will be fixed.

I'm happy to chat in our [Slack channel](https://github.com/neon-bindings/neon#get-involved) about possible fixes, but I most likely won't have chance to work on it myself.

andiliu-gh:
Oh I was totally unaware that we were still using the legacy backend. I assumed that since we were updated that it would be backed by the `napi` backend. I also asked my co-worker regarding that and that's what he concluded as well. I'll get back to you if we're still facing issues w the `napi` backend!

kjvalencik:
@andiliu-gh Let me know! We chose *not* to change the default backend on existing projects since it would needlessly cause breaking changes. We _may_ add a compiler warning in the future (or default to no backend). Instead, new projects will default to the `napi` backend.

[...]

PR #785, opened 2021-08-31
fix(neon-sys): Add a hack for recent electron versions on windows

kjvalencik:
Fixes https://github.com/neon-bindings/neon/issues/782

### DIFF
--- a/crates/neon-sys/native/src/neon.cc
+++ b/crates/neon-sys/native/src/neon.cc
@@ -221,15 +221,27 @@ extern "C" bool Neon_ArrayBuffer_New(v8::Local<v8::ArrayBuffer> *out, v8::Isolat
 
 
 extern "C" size_t Neon_ArrayBuffer_Data(v8::Isolate *isolate, void **base_out, v8::Local<v8::ArrayBuffer> buffer) {
-#if (V8_MAJOR_VERSION >= 8)
-    auto contents = buffer->GetBackingStore();
-    *base_out = contents->Data();
-    return contents->ByteLength();
-#else
-    v8::ArrayBuffer::Contents contents = buffer->GetContents();
-    *base_out = contents.Data();
-    return contents.ByteLength();
-#endif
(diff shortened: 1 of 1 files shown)


crates/neon-sys/native/src/neon.cc
```diff
@@ -221,15 +221,21 @@ extern "C" bool Neon_ArrayBuffer_New(v8::Local<v8::ArrayBuffer> *out, v8::Isolat
...
-#if (V8_MAJOR_VERSION >= 8)
-    auto contents = buffer->GetBackingStore();
-    *base_out = contents->Data();
-    return contents->ByteLength();
-#else
-    v8::ArrayBuffer::Contents contents = buffer->GetContents();
-    *base_out = contents.Data();
-    return contents.ByteLength();
-#endif
+  #if _MSC_VER && NODE_RUNTIME_ELECTRON && NODE_MODULE_VERSION >= 89
```
dherman, line 224: If there's no way to reliably detect we're in electron (since `NODE_RUNTIME_ELECTRON` is only set by `electron-build-env` and so not super reliable), we can just unconditionally use this hack for all Windows builds with the legacy backend. It's probably a performance hit but only for the legacy backend, and we'll encourage people to upgrade to the new backend anyway.

crates/neon-sys/native/src/neon.cc
```diff
@@ -221,15 +221,21 @@ extern "C" bool Neon_ArrayBuffer_New(v8::Local<v8::ArrayBuffer> *out, v8::Isolat
...
+    v8::Local<v8::Object> local;
+    node::Buffer::New(isolate, buffer, 0, buffer->ByteLength()).ToLocal(&local);
```
dherman, line 226: I like this hack, I think it quietly fixes a painful build issue so users don't need to worry about it. I'd just add some comments explaining why it's needed, for internal maintainability.

PR #787, opened 2021-09-03
feat(neon): Add `JoinHandle` as a result of `Channel::send`

kjvalencik:
Part of https://github.com/neon-bindings/rfcs/pull/32

### DIFF
--- a/src/event/event_queue.rs
+++ b/src/event/event_queue.rs
@@ -1,5 +1,5 @@
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{mpsc, Arc};
 
 use neon_runtime::raw::Env;
 use neon_runtime::tsfn::ThreadsafeFunction;
@@ -104,9 +104,10 @@ impl Channel {
 
     /// Schedules a closure to execute on the JavaScript thread that created this Channel
     /// Panics if there is a libuv error
-    pub fn send<F>(&self, f: F)
+    pub fn send<T, F>(&self, f: F) -> JoinHandle<T>
     where
-        F: FnOnce(TaskContext) -> NeonResult<()> + Send + 'static,
+        T: Send + 'static,
+        F: FnOnce(TaskContext) -> NeonResult<T> + Send + 'static,
     {
         self.try_send(f).unwrap()
     }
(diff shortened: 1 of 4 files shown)


test/napi/lib/threads.js
```diff
@@ -71,4 +71,17 @@ const assert = require('chai').assert;
     // Asynchronously GC to give the task queue a chance to execute
     setTimeout(() => global.gc(), 10);
   });
+
+  it('should be able to join on the result of a channel', function (cb) {
+    let msg = "Uninitialized";
```
dherman, line 76: This and the Rust logic that goes with it took me a few minutes to follow 😂 

So `msg` is a shared mutable cell, we're modifying it after a 10ms timeout, and we're creating a Rust thread that waits 100ms and sends one event back to the main thread with the updated message, joins on that message, and sends another event back to the main thread that sends the message to an async callback. And then that callback tests that the message was the updated message.

Did I get that right? A comment might help 😅

dherman (approved):
Looks good to me! Maybe some comments in the test, which has kind of intricate control flow, would help. But it's up to you.

src/event/event_queue.rs
```diff
@@ -115,21 +116,29 @@ impl Channel {
...
     ///
     /// See [`SendError`] for additional details on failure causes.
-    pub fn try_send<F>(&self, f: F) -> Result<(), SendError>
+    pub fn try_send<T, F>(&self, f: F) -> Result<JoinHandle<T>, SendError>
     where
-        F: FnOnce(TaskContext) -> NeonResult<()> + Send + 'static,
+        T: Send + 'static,
+        F: FnOnce(TaskContext) -> NeonResult<T> + Send + 'static,
     {
+        let (tx, rx) = mpsc::sync_channel(1);
```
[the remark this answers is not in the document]
kjvalencik: Yes, that's common naming. It comes from telecomm.

https://doc.rust-lang.org/rust-by-example/std_misc/channels.html

test/napi/lib/threads.js
kjvalencik: Yes, that's right! I'll write a more detailed comment that also explains _what_ we're trying to prove here. Specifically, we want to confirm that we can join on the result of `Channel::send` without impacting the JavaScript main thread.

kjvalencik: Added a bunch of comments. 😄

PR #789, opened 2021-09-13
Initial implementation for JsPromise and TaskBuilder

kjvalencik:
Implements the [`JsPromise` and `TaskBuilder` RFC](https://github.com/neon-bindings/rfcs/pull/35).

Adds two new feature flags:
* `promise-api`
* `task-api`

[Rust Docs](https://deploy-preview-789--neon-bindings-docs.netlify.app/neon/)

```rust
pub fn sum(mut cx: FunctionContext) -> JsResult<JsPromise> {
    let nums = cx.argument::<JsTypedArray<f64>>(0)?.as_slice(&cx).to_vec();
    let promise = cx
        .task(move || nums.into_iter().sum())
        .promise(|cx, n: f64| Ok(cx.number(n)));

    Ok(promise)
}
```

### DIFF
--- a/crates/neon-runtime/src/napi/async_work.rs
+++ b/crates/neon-runtime/src/napi/async_work.rs
@@ -1,0 +1,138 @@
+//! Rust wrappers for Node-API simple asynchronous operations
+//!
+//! Unlike `napi_async_work` which threads a single mutable pointer to a data
+//! struct to both the `execute` and `complete` callbacks, the wrapper follows
+//! a more idiomatic Rust ownership pattern by passing the output of `execute`
+//! into the input of `complete`.
+//!
+//! https://nodejs.org/api/n-api.html#n_api_simple_asynchronous_operations
+
+use std::ffi::c_void;
+use std::mem;
+use std::ptr;
+
+use crate::napi::bindings as napi;
+use crate::raw::Env;
+
+type Execute<T, O> = fn(input: T) -> O;
+type Complete<O> = fn(env: Env, output: O);
+
+/// Schedule work to execute on the libuv thread pool
+///
+/// # Safety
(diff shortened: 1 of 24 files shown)


crates/neon-runtime/src/napi/async_work.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Return the input if `State::Input`, replacing with `State::Empty`
+    fn take_input(&mut self) -> Option<T> {
+        match mem::replace(self, Self::Empty) {
+            Self::Input(input) => Some(input),
+            _ => None,
+        }
+    }
+
+    /// Return the output if `State::Output`, replacing with `State::Empty`
+    fn take_output(&mut self) -> Option<O> {
```
dherman, line 94: I'd suggest making this consume `self` so you don't need to synthesize an `Empty` state for this case.

crates/neon-runtime/src/napi/async_work.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/// State of the task that is transitioned by `execute` and `complete`
+enum State<T, O> {
+    /// Initial data input passed to `execute`
+    Input(T),
+    /// Return data of `execute` passed to `complete`
+    Output(O),
+    /// Transient state between calls
+    Empty,
```
dherman, line 81: If we make `take_output` consume `self`, we could call the `Empty` case `Executing` so it's a tiny bit clearer it's for the state when we've consumed the input but don't yet have the output.

dherman (approved):
It's remarkable how clean the implementation is -- I bet the hardest part was figuring out the API, and then the primitives match the Node-API design enough that the implementation is pretty clean!

I made a couple tiny suggestions in `async_work` but I'll leave it to you.

PR #794, opened 2021-09-17
Remove unused cslice dependency

complexspaces:
This dependency appears to be leftover from a very old version of the Neon project. I couldn't find any places it is used.


kjvalencik:
Great catch! Thanks!

PR #795, opened 2021-09-18
Overload result type of `Object::get()`

dherman:
This PR changes the `Object::get()` API to return an overloaded result type `V: Value` instead of `JsValue`, and automatically performs the downcasting on behalf of the caller. Since it's already a fallible API, this should likely end up being generally more ergonomic.

This is a backwards-incompatible change, however, and will require migration notes if we decide to go through with it.

I decided not to implement the same change for `JsFunction::call()` and `JsFunction::construct()` because we may not need the backwards-incompatible change if an ergonomic new `Arguments` builder API ends up being nicer than `call()` and `construct()` anyway (and maybe we'd just deprecate those?). I'll implement that in a separate PR.

### DIFF
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -95,7 +95,8 @@ mod traits {
             cx: &mut C,
             key: K,
         ) -> NeonResult<Handle<'a, V>> {
-            let v: Handle<JsValue> = build(cx.env(), |out| unsafe { key.get_from(out, self.to_raw()) })?;
+            let v: Handle<JsValue> =
+                build(cx.env(), |out| unsafe { key.get_from(out, self.to_raw()) })?;
             v.downcast_or_throw(cx)
         }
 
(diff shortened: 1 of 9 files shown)


src/object/mod.rs
```diff
@@ -90,12 +90,14 @@ mod traits {
 
     /// The trait of all object types.
     pub trait Object: Value {
-        fn get<'a, C: Context<'a>, K: PropertyKey>(
+        fn get<'a, C: Context<'a>, K: PropertyKey, V: Value>(
```
kjvalencik, line 93: In other places (e.g., `downcast`), we've put the return type generic in the first position to make turbo fish a little easier.

```rust
// Currently:
let emit = emitter.get::<_, _, JsFunction>(&mut cx, "emit")?;

// Proposed
let emit = emitter.get::<JsFunction, _, _>(&mut cx, "emit")?;
```

kjvalencik (changes requested):
@dherman Was it in intentional to only make this change for the `legacy` backend? The trait is duplicated for Node-API further down in the file.

dherman:
Replaced by #839.

PR #796, opened 2021-09-18
Arguments builder API

dherman:
~First~ ~Second~ N'th (I lost count) draft of a function arguments builder API, which tries to provide a more ergonomic experience for invoking JS functions.

## Benefits

- Arguments don't need to be explicitly wrapped in a vector
- No need to store `this` binding in a temp variable before invoking the function
- Default `this` binding of `undefined` for common case
- Heterogeneous arguments without explicit upcasting
- Overloaded result type so callers don't need to explicitly downcast

Before:

```rust
let this = cx.undefined();
let args = vec![
    cx.number(17).upcast(),
    cx.string("hello").upcast()
];
let v: Handle<JsArray> = f.call(&mut cx, this, args)?
    .downcast_or_throw()?;
```

After:

```rust
let v: Handle<JsArray> = f
    .call_with(&cx)
    .arg(cx.number(17))
    .arg(cx.string("hello"))
    .apply(&mut cx)?;
```

## Open questions

- ~Since `.call()` is generic in two types (the context and the result type), turbofish instantiation is a little ugly (e.g. `.call::<_, JsArray>(&mut cx)`). You can usually do the instantiation in a let binding instead (see the above example), and `.exec()` avoids having to instantiate in the case where you're ignoring the result. So maybe this is OK, but I'm just calling it out.~
- ~I made `Arguments` implement `Clone` so you can reuse and call an instance of the builder multiple times. We could possibly make `call()` etc not consume `self` but would that require forcing an extra vector copy?~
- ~Are we OK with requiring `.args(())` for the empty arguments case? It seems like the least-annoying option we've come up with, compared to:~
  - ~`f.with().call(&mut cx)`~
  - ~`f.args::<JsValue>().call(&mut cx)`~
  - ~The one other option I can think of is backwards-incompatibly removing the existing `.call()` method from `JsFunction` and having `.call()`, `.new()`, and `.exec()` methods directly on `JsFunction`. But this could be quite a disruptive change to our existing users.~


src/types/mod.rs
```diff
@@ -774,3 +808,171 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+///     .arg(cx.string("42"))
+///     .call(&mut cx)?;
+/// # Ok(x)
+/// # }
+/// ```
+#[derive(Clone)]
+pub struct FunctionCall<'a> {
+    callee: Handle<'a, JsFunction>,
+    this: Handle<'a, JsValue>,
+    args: Vec<Handle<'a, JsValue>>,
```
kjvalencik, line 831: We could consider [`tinyvec`](https://docs.rs/tinyvec/1.4.0/tinyvec/index.html) to try to avoid heap allocations.

src/types/mod.rs
```diff
@@ -774,3 +808,171 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn call<'b, C: Context<'b>, V: Value>(self, cx: &mut C) -> JsResult<'b, V> {
+        let v = self.callee.call(cx, self.this, self.args)?;
+        v.downcast_or_throw(cx)
+    }
+
+    /// Make the function call for side effect, discarding the result value. This method is
+    /// preferable to [`call()`](crate::types::FunctionCall::call) when the result value is
+    /// not needed, since it does not require specifying a result type.
+    pub fn exec<'b, C: Context<'b>>(self, cx: &mut C) -> NeonResult<()> {
```
kjvalencik, line 885: I really, _really_ like this! When testing out I immediately hit the issue of needing to specify a return type when I didn't care and this made it so much easier. :)

src/types/mod.rs
```diff
@@ -774,3 +808,171 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+
+    /// Add multiple arguments to the arguments list.
+    pub fn args<A: Arguments<'a>>(mut self, args: A) -> Self {
+        args.append(&mut self.args);
+        self
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn call<'b, C: Context<'b>, V: Value>(self, cx: &mut C) -> JsResult<'b, V> {
```
kjvalencik, line 877: It would be really nice to make this take `&self` since it doesn't need to consume `self` (for the Node-API call). In turn that would allow the other methods to take `&mut self` instead of `self` which makes for better optional chaining.

src/types/mod.rs
```diff
@@ -774,3 +808,171 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+    pub fn exec<'b, C: Context<'b>>(self, cx: &mut C) -> NeonResult<()> {
+        let undefined: Handle<JsValue> = cx.undefined().upcast();
+        let _ = self.callee.call(cx, undefined, self.args)?;
+        Ok(())
+    }
+}
+
+/// The trait for specifying arguments in a [`Call`](crate::types::Call) or
+/// [`FunctionCall`](crate::types::FunctionCall).
+pub trait Arguments<'a> {
```
kjvalencik, line 940: Do we want to allow users to implement this `trait`? It _could_ be nice to allow users to do it, but if we don't make this `Sealed`, I would like to bikeshed the shape a bit more.

kjvalencik (commented):
I really like this API. This is a big improvement. I think it can be optimized further if we don't tie ourselves to using the existing function call implementation and instead go directly to Node-API.

Goals
* Avoid allocating multiple vecs
* Avoid iterating multiple times
* Non-consuming builder

Since both `Handle` and `JsValue` are `#[repr(C)]` with only a single field, these types are equivalent to `napi_value`.

It should be safe to transmute directly from `&mut [Handle<JsValue>]` to `&mut [napi_value]`. This does not require consuming the `Vec` or making any copies.

src/types/mod.rs
dherman: No I agree, it should be sealed at least for now. I thought I remembered you had done some sealed traits before but my greps aren't turning up anything -- what conventions do you like to use for sealed traits?

src/types/mod.rs
dherman: Pushed a change that implements this!

src/types/mod.rs
dherman: I've stuck with SmallVec for the moment but I can migrate to TinyVec a little later. Getting CI to pass again ATM…

dherman:
I just realized `Call::new()` isn't generic in the result type the way `JsFunction::construct()` is. I'll address that soon.

dherman:
Ready for re-review when you have a chance, @kjvalencik 🙏

src/context/mod.rs
```diff
@@ -94,9 +94,10 @@
 //!
 //!     while !done {
 //!         done = cx.execute_scoped(|mut cx| {                   // temporary scope
-//!             let args: Vec<Handle<JsValue>> = vec![];
-//!             let obj = next.call(&mut cx, iterator, args)?     // temporary object
-//!                 .downcast_or_throw::<JsObject, _>(&mut cx)?;
+//!             let obj: Handle<JsObject> = next                  // temporary object
+//!                 .args(())
```
kjvalencik, line 98: This line can be deleted and use `JsFunction::this`.

```suggestion
```

src/types/internal.rs
```diff
@@ -113,3 +113,6 @@ pub(crate) trait Callback<T: Clone + Copy + Sized>: Sized {
         }
     }
 }
+
+/// This type marks the `Arguments` trait as sealed.
+pub trait ArgumentsInternal {}
```
kjvalencik, line 118: My preference is a `trait Sealed` in a `mod private`. This is a common pattern and it makes the docs more clear since the user will still see the super trait.

```rust
pub trait Arguments<'a>: ArgumentsInternal { }

// vs

pub trait Arguments<'a>: private::Sealed { }
```

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+    };
+}
+
+impl_arguments! {
+    (V1, V2, V3, V4, V5, V6, V7, V8,
+     V9, V10, V11, V12, V13, V14, V15, V16,
+     V17, V18, V19, V20, V21, V22, V23, V24,
+     V25, V26, V27, V28, V29, V30, V31, V32,);
+
+    (v1, v2, v3, v4, v5, v6, v7, v8,
```
kjvalencik, line 1075: It would be nice to mark most of these as `#[doc(hidden)]` to keep the docs page more manageable.

https://deploy-preview-796--neon-bindings-docs.netlify.app/neon/types/trait.arguments#foreign-impls

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+
+    /// Add multiple arguments to the arguments list.
+    pub fn args<A: Arguments<'a>>(&mut self, args: A) -> &mut Self {
+        args.append(&mut self.args);
+        self
+    }
+
+    /// Call the function as a constructor (like a JavaScript `new` expression).
+    /// If the function returns without throwing, returns the resulting object.
+    pub fn new<'b, C: Context<'b>>(&self, cx: &mut C) -> JsResult<'b, T> {
```
kjvalencik, line 1012: Should we avoid the `new` keyword in case it's added as a keyword in future Rust versions?

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+
+    /// Add multiple arguments to the arguments list.
+    pub fn args<A: Arguments<'a>>(&mut self, args: A) -> &mut Self {
+        args.append(&mut self.args);
+        self
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn call<'b, C: Context<'b>, V: Value>(&self, cx: &mut C) -> JsResult<'b, V> {
```
kjvalencik, line 973: Can you reverse the order of `C` and `V` in the generics? That way it's consistent with other usages and also a little nicer for turbofish:

```rust
.call::<JsNumber, _>(&mut cx)
```

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+
+    /// Call the function as a constructor (like a JavaScript `new` expression).
+    /// If the function returns without throwing, returns the resulting object.
+    pub fn new<'b, C: Context<'b>>(&self, cx: &mut C) -> JsResult<'b, T> {
+        self.callee.do_construct(cx, &self.args)
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn call<'b: 'a, C: Context<'b>, V: Value>(&self, cx: &mut C) -> JsResult<'b, V> {
```
kjvalencik, line 1018: Same as above. Reverse `C` and `V` in generics list.

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn call<'b, C: Context<'b>, V: Value>(&self, cx: &mut C) -> JsResult<'b, V> {
+        let v: Handle<JsValue> = self.callee.do_call(cx, self.this, &self.args)?;
+        v.downcast_or_throw(cx)
+    }
+
+    /// Make the function call for side effect, discarding the result value. This method is
+    /// preferable to [`call()`](crate::types::FunctionCall::call) when the result value is
+    /// not needed, since it does not require specifying a result type.
+    pub fn exec<'b, C: Context<'b>>(&self, cx: &mut C) -> NeonResult<()> {
```
kjvalencik, line 981: Should this completely discard or return `JsValue`? I _think_ favor completely discard, but I could go either way.

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+        let _: Handle<JsValue> = self.callee.do_call(cx, self.this, &self.args)?;
```
kjvalencik, line 982: The best I can tell, `do_call` isn't generic over the return type, so the binding isn't necessary. I recommend removing the binding to ensure we get a clippy warning if for some reason the `?` and explicit typing get dropped through refactoring.

src/types/mod.rs
```diff
@@ -774,3 +902,178 @@ impl<T: Object> ValueInternal for JsFunction<T> {
...
+        let _: Handle<JsValue> = self.callee.do_call(cx, undefined, &self.args)?;
+        Ok(())
+    }
+}
+
+/// The trait for specifying arguments in a [`Call`](crate::types::Call) or
+/// [`FunctionCall`](crate::types::FunctionCall). This trait is sealed and cannot
+/// be implemented by types outside of the Neon crate.
+pub trait Arguments<'a>: ArgumentsInternal {
+    #[doc(hidden)]
```
kjvalencik, line 1038: Could this method be moved to `ArgumentsInternal` so that it doesn't need to be exposed to users (even doc hidden)?

test/napi/src/js/functions.rs
```diff
@@ -124,11 +119,7 @@ pub fn throw_and_catch(mut cx: FunctionContext) -> JsResult<JsValue> {
 pub fn call_and_catch(mut cx: FunctionContext) -> JsResult<JsValue> {
     let f: Handle<JsFunction> = cx.argument(0)?;
     Ok(cx
-        .try_catch(|cx| {
-            let global = cx.global();
-            let args: Vec<Handle<JsValue>> = vec![];
-            f.call(cx, global, args)
-        })
+        .try_catch(|cx| f.args(()).this(cx.global()).call(cx))
```
kjvalencik, line 122: I think `.args` can be dropped.

```suggestion
        .try_catch(|cx| f.this(cx.global()).call(cx))
```

src/types/mod.rs
dherman: I don't think it'll be made into a Rust keyword, but I think we should use `.construct` for consistency.

dherman:
I'm going to close this for now since I submitted #817 to implement part of this functionality (in particular the `Arguments` trait) first. When that merges I'll follow up with a builder API PR.

PR #797, opened 2021-09-18
Make it harder to forge `Throw` tokens

dherman:
This PR makes it harder (though not technically impossible) to forge `Throw` tokens. We already have safety mechanisms in place in case of forged `Throw` tokens, leading to well-defined panics (as opposed to something unsafe, like undefined behavior). But this should help prevent programs from _accidentally_ saving and reusing a `Throw` token.

There is also a test case that still does show how it's technically possible to forge a token, and demonstrates that it still has predictable behavior.

An alternative that would be even safer would be to place a lifetime on a `Throw` token, making it impossible to store them in long-lived storage. But this would leak into the type signatures of e.g. `NeonResult` and APIs that use it, which would add a lot of complexity to Neon for little value. The panic is for a rare enough case that the extra type safety wouldn't be worth it IMO.

One more change I want to make in this PR is to add `!Send` phantom data to the `Throw` type so that it can't be shared outside of its thread.

### DIFF
--- a/src/result/mod.rs
+++ b/src/result/mod.rs
@@ -48,7 +48,7 @@ use std::marker::PhantomData;
 ///
 /// [unit]: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields
 #[derive(Debug)]
-pub struct Throw(PhantomData<*mut ()>); // *mut is !Send + !Sync, making it harder to forge
+pub struct Throw(PhantomData<*mut ()>); // *mut is !Send + !Sync, making it harder to accidentally store
 
 impl Throw {
     pub(crate) fn new() -> Self {
--- a/src/result/mod.rs
+++ b/src/result/mod.rs
@@ -37,6 +37,7 @@ use crate::context::Context;
 use crate::handle::Handle;
 use crate::types::Value;
 use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::marker::PhantomData;
 
 /// A [unit type][unit] indicating that the JavaScript thread is throwing an exception.
 ///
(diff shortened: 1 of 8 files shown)


src/result/mod.rs
```diff
@@ -47,7 +47,7 @@ use std::fmt::{Display, Formatter, Result as FmtResult};
 ///
 /// [unit]: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields
 #[derive(Debug)]
-pub struct Throw;
+pub struct Throw(pub(crate) ());
```
kjvalencik, line 50: What do you think about using a `!Send + !Sync` type here to make it a _little_ more difficult? Also, I would recommend a private constructor.

```rust
pub struct Throw(PhantomData<*mut c_void>);

impl Throw {
    pub(crate) fn new() -> Self {
        Self(PhantomData)
    }
}
```

test/dynamic/native/src/js/functions.rs
```diff
@@ -143,8 +145,43 @@ pub fn panic_and_catch(mut cx: FunctionContext) -> JsResult<JsValue> {
...
 }
 
+thread_local! {
+    static FORGED_THROW: RefCell<Option<Throw>> = RefCell::new(None);
+}
+
+fn forge_throw(mut cx: FunctionContext) -> JsResult<JsValue> {
+    // Force a random JS error to temporarily enter the throwing state.
+    let v: Handle<JsValue> = cx.null().upcast();
+    match v.downcast_or_throw::<JsNumber, _>(&mut cx) {
```
kjvalencik, line 155: `Context::throw_error` might be a little simpler:

```rust
let throw = cx.throw_error("Error").unwrap_err();
```

src/result/mod.rs
```diff
@@ -47,7 +48,13 @@ use std::fmt::{Display, Formatter, Result as FmtResult};
 ///
 /// [unit]: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields
 #[derive(Debug)]
-pub struct Throw;
+pub struct Throw(PhantomData<*mut ()>); // *mut is !Send + !Sync, making it harder to forge
```
kjvalencik, line 51: Nit, `*mut` doesn't make it harder to forge, but it does make it harder to store.  It limits the UB to storing it in TLS.

kjvalencik (approved):

PR #800, opened 2021-09-19
fix(neon): Make failing doctest work with both runtimes

dherman:
Fix failing doctest due to incompatible change between legacy and Node-API backends.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -44,7 +44,12 @@
 //!     cx: &mut impl Context<'a>,
 //!     object: Handle<'a, JsObject>
 //! ) -> JsResult<'a, JsArray> {
+//! #   #[cfg(feature = "legacy-runtime")]
+//! #   return object.downcast().or_throw(cx);
+//! #   #[cfg(feature = "napi-1")]
+//! #   return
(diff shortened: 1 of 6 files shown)


src/types/mod.rs
```diff
@@ -44,7 +44,12 @@
 //!     cx: &mut impl Context<'a>,
 //!     object: Handle<'a, JsObject>
 //! ) -> JsResult<'a, JsArray> {
+//! #   #[cfg(feature = "legacy-runtime")]
+//! #   return object.downcast().or_throw(cx);
+//! #   #[cfg(feature = "napi-1")]
+//! #   return
```
kjvalencik, line 50: Why is this return here? Also, in other places, we've disabled the entire test block.

https://github.com/neon-bindings/neon/blob/2dfe690d153c6a659a433fb12fa982d82d7d225c/src/context/mod.rs#L86

dherman: Ah that's a better way to do it. I did the `return` because the attribute required a statement rather than an expression. Your way is cleaner.

kjvalencik (approved):

PR #803, opened 2021-09-24
fix(neon-runtime): Catch panics in `Channel::send` callbacks and convert to `uncaughtException`

kjvalencik:
Panic unwinding across FFI is undefined behavior. Neon must catch panics in threadsafe function callbacks. A best effort is made to convert the panic message to an `uncaughtException`, but if unable, it will be printed to stderr.

`napi_fatal_exception` is only available in Node-API 3+, but since `napi_threadsafe_function` is only in Node-API 4+, we don't need any special handling.

### DIFF
--- a/crates/neon-runtime/src/napi/bindings/functions.rs
+++ b/crates/neon-runtime/src/napi/bindings/functions.rs
@@ -202,6 +202,17 @@ mod napi1 {
     );
 }
 
+#[cfg(feature = "napi-3")]
+mod napi3 {
+    use super::super::types::*;
+
+    generate!(
+        extern "C" {
+            fn fatal_exception(env: Env, err: Value) -> Status;
+        }
+    );
+}
+
 #[cfg(feature = "napi-4")]
 mod napi4 {
     use super::super::types::*;
@@ -285,6 +296,8 @@ mod napi6 {
 }
 
 pub(crate) use napi1::*;
+#[cfg(feature = "napi-3")]
+pub(crate) use napi3::*;
 #[cfg(feature = "napi-4")]
 pub(crate) use napi4::*;
 #[cfg(feature = "napi-5")]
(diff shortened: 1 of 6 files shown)


crates/neon-runtime/src/napi/tsfn.rs
```diff
@@ -186,10 +188,38 @@ impl<T: Send + 'static> ThreadsafeFunction<T> {
...
+
+        // Try to throw an uncaught exception
+        if let Some(env) = env {
+            if super::error::fatal_exception_from_utf8(env, msg).is_ok() {
+                return;
+            }
+        }
+
+        // Failed to trigger an `uncaughtException`; fallback to print
+        eprintln!("Failed to throw an exception for panic: {}", msg);
```
kjvalencik, line 222: Should we print or let this disappear? Should we consider using [`log`](https://crates.io/crates/log) so users can decide for themselves what should print?

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -202,6 +202,17 @@ mod napi1 {
     );
 }
 
+#[cfg(feature = "napi-3")]
+mod napi3 {
+    use super::super::types::*;
+
+    generate!(
+        extern "C" {
+            fn fatal_exception(env: Env, err: Value) -> Status;
```
kjvalencik, line 211: This is a _fantastic_ API for adding panic catches. It allows you to trigger `uncaughtException` even when you aren't on a JavaScript stack.

test/napi/lib/threads.js
```diff
@@ -71,4 +81,15 @@ const assert = require('chai').assert;
     // Asynchronously GC to give the task queue a chance to execute
     setTimeout(() => global.gc(), 10);
   });
+
+  it('should handle panic in channel callback', function (cb) {
+    process.removeAllListeners("uncaughtException");
```
kjvalencik, line 86: `mocha` adds an `uncaughtException` listener. It needs to be removed to catch it in the test.

kjvalencik:
This should likely go against `next/0.10` because it's technically a break change, even if it's for safety.

PR #804, opened 2021-09-25
Fix broken link in README

alaz:
The existing link for "platform resources" in README leads to `404`.


kjvalencik:
Thanks for the fix!

PR #806, opened 2021-09-29
feat(napi-8): Support Object.freeze & seal 

ovr:
Hello!

This PR introduces support for `Object.freeze` & `Object.seal` from N-API Version 8.

Thanks


crates/neon-runtime/src/napi/object.rs
```diff
@@ -8,6 +8,25 @@ pub unsafe fn new(out: &mut Local, env: Env) {
...
+pub unsafe fn freeze(env: Env, object: Local) -> bool {
+    let status = napi::object_freeze(env, object);
+
+    status == napi::Status::Ok
+}
+
+#[cfg(feature = "napi-8")]
+/// This method seals a given object. This prevents new properties from being added to it, as well as marking all existing properties as non-configurable.
+// This is described in Section 19.1.2.20 of the ECMA-262 specification.
+pub unsafe fn seal(env: Env, object: Local) -> bool {
```
kjvalencik, line 24: The `-> bool` pattern was only used for some legacy APIs and isn't great because it's unclear whether there was an unexpected error or an exception. It also doesn't force a clippy warning that a `Result` is unused.

Can you make this return a `Result(), napi::Status>` instead? It should only return an error on `PendingException`, anything else would be a Neon bug.

Maybe something like this:

```rust
pub unsafe fn seal(env: Env, object: Local) -> Result<(), napi::Status {
   let status = napi::object_seal(env, object);
   
   if status == napi::Status::PendingException {
       return Err(status);
   }
   
   assert_eq!(status, napi::Status::Ok);

    Ok(())
}
```

src/types/mod.rs
```diff
@@ -587,6 +587,16 @@ impl JsObject {
             Handle::new_internal(JsObject(local))
         }
     }
+
+    #[cfg(feature = "napi-8")]
+    pub fn freeze<'a, C: Context<'a>>(self, cx: &mut C) -> bool {
```
kjvalencik, line 592: These methods should return `NeonResult<()>` to reflect that an exception can be thrown.

```
> Object.freeze(Buffer.alloc(1))
Uncaught TypeError: Cannot freeze array buffer views with elements
    at Function.freeze (<anonymous>)
```

src/types/mod.rs
```diff
@@ -587,6 +587,16 @@ impl JsObject {
...
+        unsafe { neon_runtime::object::freeze(cx.env().to_raw(), self.to_raw()) }
+    }
+
+    #[cfg(feature = "napi-8")]
+    pub fn seal<'a, C: Context<'a>>(self, cx: &mut C) -> bool {
```
kjvalencik, line 597: I think it would be better to add `freeze` and `seal` to the [`Object` trait](https://github.com/neon-bindings/neon/blob/7c93fd8d4e68a277026707d734b42eca02c20e3d/src/object/mod.rs#L92) instead. That way they can be called from anything that is an `Object`, instead of needing to be upcast to a `JsObject` first.

kjvalencik (changes requested):
Thanks so much for your contribution! I've suggested a few changes. Let me know if you have any questions.

src/types/mod.rs
```diff
@@ -587,6 +587,16 @@ impl JsObject {
             Handle::new_internal(JsObject(local))
         }
     }
+
+    #[cfg(feature = "napi-8")]
```
kjvalencik, line 591: Documentation would be helpful. If that's not something you are comfortable writing, I can help.

kjvalencik:
Sorry that I forgot about this! This feature has been merged. I'll be ensure to give you some credit in the release notes for your excellent work.

PR #811, opened 2021-10-06
Create functions from closures

kjvalencik:
## What

Allow Neon users to create functions from closures.

```rust
    cx.export_function("count_called", {
        let n = std::cell::RefCell::new(0);

        move |mut cx| {
            *n.borrow_mut() += 1;

            Ok(cx.number(*n.borrow()))
        }
    })?;
```

## How

The closure is boxed and added as data. It works very similar to the current functions except the trampoline function is generic.

Dropping the Rust closure requires [`napi_add_finalizer`](https://nodejs.org/api/n-api.html#n_api_napi_add_finalizer) which was added in Node-API 5. For this reason, when using older Node-API versions, only `fn` instead of `Fn` are accepted.

## Future Expansion

It would be unsound to accept `FnMut` because a function may be reentrant. A function could trivially accept a reference to itself and then call that function. However, it is very cumbersome to use `RefCell` to get access to mutable state.

Since most functions will *not* have any possibility for reentrancy and since closure lifetime inference is very limited in Rust, it would be useful for Neon to provide a `JsFunction::new_mut` that accepts an `FnMut` and wraps it in a `RefCell`.

### DIFF
--- a/crates/neon-runtime/src/napi/fun.rs
+++ b/crates/neon-runtime/src/napi/fun.rs
@@ -51,12 +51,17 @@ where
             ptr::null_mut(),
         );
 
+        // If adding the finalizer fails the closure will leak, but it would
+        // be unsafe to drop it because there's no guarantee V8 won't use the
+        // pointer.
         assert_eq!(status, napi::Status::Ok);
     }
 
     Ok(out)
 }
 
+// C ABI compatible function for invoking a boxed closure from the data field
+// of a Node-API JavaScript function
 unsafe extern "C" fn call_boxed<F>(env: Env, info: napi::CallbackInfo) -> Local
 where
     F: Fn(Env, napi::CallbackInfo) -> Local + 'static,
--- a/crates/neon-runtime/src/napi/fun.rs
+++ b/crates/neon-runtime/src/napi/fun.rs
@@ -9,7 +9,7 @@ use crate::raw::{Env, Local};
 
 pub unsafe fn new<F>(env: Env, name: &str, callback: F) -> Result<Local, napi::Status>
 where
-    F: Fn(Env, napi::CallbackInfo) -> Local + Send + 'static,
+    F: Fn(Env, napi::CallbackInfo) -> Local + 'static,
 {
     let mut out = MaybeUninit::uninit();
     let data = Box::into_raw(Box::new(callback));
@@ -59,7 +59,7 @@ where
 
 unsafe extern "C" fn call_boxed<F>(env: Env, info: napi::CallbackInfo) -> Local
 where
-    F: Fn(Env, napi::CallbackInfo) -> Local + Send + 'static,
+    F: Fn(Env, napi::CallbackInfo) -> Local + 'static,
 {
     let mut data = MaybeUninit::uninit();
     let status = napi::get_cb_info(
--- a/crates/neon-runtime/src/napi/fun.rs
+++ b/crates/neon-runtime/src/napi/fun.rs
@@ -1,29 +1,81 @@
 //! Facilities for working with JS functions.
 
+use std::mem::MaybeUninit;
 use std::os::raw::c_void;
-use std::ptr::null;
+use std::ptr;
 
-use crate::call::CCallback;
 use crate::napi::bindings as napi;
 use crate::raw::{Env, Local};
 
-/// Mutates the `out` argument provided to refer to a newly created `v8::Function`. Returns
-/// `false` if the value couldn't be created.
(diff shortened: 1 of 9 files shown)


src/types/mod.rs
```diff
@@ -716,6 +719,41 @@ impl JsFunction {
...
+        V: Value,
+    {
+        use std::any;
+        use std::mem;
+        use std::panic::AssertUnwindSafe;
+        use std::ptr;
+
+        use crate::types::error::convert_panics;
+
+        let name = any::type_name::<F>();
```
kjvalencik, line 737: This seemed like a nice little addition without a breaking change. It sets the `Function.prototype.name` to the Rust type name of the function used to create it.

```
[Function: napi_tests::js::strings::return_js_string]
```

It's not the best on closures, but it still seems like an improvement over no name.

```
[Function: napi_tests::main::{{closure}}]
```

crates/neon-runtime/src/napi/fun.rs
```diff
@@ -22,6 +23,77 @@ pub unsafe fn new(out: &mut Local, env: Env, callback: CCallback) -> bool {
...
+        let status = napi::add_finalizer(
+            env,
+            out,
+            data.cast(),
+            Some(drop_function::<F>),
+            ptr::null_mut(),
+            ptr::null_mut(),
+        );
+
+        assert_eq!(status, napi::Status::Ok);
```
kjvalencik, line 70: It's not safe to drop the boxed callback on failure here because the finalizer has already been given to Node. We do not have a guarantee that it won't be called.

crates/neon-runtime/src/napi/fun.rs
```diff
@@ -22,6 +23,77 @@ pub unsafe fn new(out: &mut Local, env: Env, callback: CCallback) -> bool {
...
+        ) {
+            Box::from_raw(finalize_data.cast::<F>());
+        }
+
+        let status = napi::add_finalizer(
+            env,
+            out,
+            data.cast(),
+            Some(drop_function::<F>),
+            ptr::null_mut(),
```
kjvalencik, line 66: For consistency, we might want to put the data in the `finalize_hint` instead of `finalize_data` because that's where it goes on `JsBox`.

test/napi/lib/functions.js
```diff
@@ -114,4 +114,8 @@ describe('JsFunction', function() {
     assert.equal(addon.is_construct.call({}).wasConstructed, false);
     assert.equal((new addon.is_construct()).wasConstructed, true);
   });
+
+  it('should be able to call a function from an FnMut', function() {
```
kjvalencik, line 118: I tested manually that closures get dropped, but it would be nice to have an explicit test.

dherman:
As appealing as this is, unfortunately allowing `FnMut` is unsound. You can make a function call itself recursively, which Rust doesn't allow for closures (since there can be a live active borrow of the closure data).

```rust
let mut data = /* ... */;

let f = JsFunction::new(move |mut cx| {
    let this: Handle<JsObject> = cx.this()?;
    let method: Handle<JsFunction> = this.get(&mut cx, "recursive")?;
    let n: Handle<JsNumber> = cx.argument(0)?;
    let n = n.value(&mut cx) as u32;
    println!("obj.recursive({})", n);

    do_something(&data); // borrow the mutable data in the closure

    if (n > 0) {
        let n = cx.number(n - 1);
        let args = vec![n.upcast()];
        method.call(&mut cx, this, args)?;
    }
});

obj.set("recursive", f)?;
```

crates/neon-runtime/src/napi/fun.rs
```diff
@@ -1,29 +1,81 @@
...
+
+        assert_eq!(status, napi::Status::Ok);
+    }
+
+    Ok(out)
 }
 
-pub unsafe fn get_dynamic_callback(_env: Env, data: *mut c_void) -> *mut c_void {
-    data
+unsafe extern "C" fn call_boxed<F>(env: Env, info: napi::CallbackInfo) -> Local
```
dherman, line 60: A brief comment explaining the purpose of this function, i.e., what's boxed, would probably be good.

crates/neon-runtime/src/napi/fun.rs
dherman: Is this a potential leak then?

dherman (changes requested):
This looks great. I just have a few suggestions for clarity and one question about a possible leak.

dherman (approved):
Thanks for the clarifying comments. Looks great!

PR #815, opened 2021-10-28
perf: Performance improvements for Node-API backend

kjvalencik:
Switching to `raw::InheritedScope` is the more significant improvement. The `argv` change is more minor.

tl;dr -- We were *always* creating handle scopes on function calls, but Node-API has *already* done it for us. It's only necessary to create a scope for `execute_scoped` and `compute_scoped`.

### DIFF
--- a/crates/neon-runtime/src/napi/call.rs
+++ b/crates/neon-runtime/src/napi/call.rs
@@ -2,10 +2,45 @@ use std::mem::MaybeUninit;
 use std::os::raw::c_void;
 use std::ptr::null_mut;
 
-use smallvec::{smallvec, SmallVec};
+use smallvec::SmallVec;
 
 use crate::napi::bindings as napi;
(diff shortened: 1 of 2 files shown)


crates/neon-runtime/src/napi/call.rs
```diff
@@ -61,6 +96,7 @@ pub unsafe fn data(env: Env, info: FunctionCallbackInfo, out: &mut *mut c_void)
 }
 
 /// Gets the number of arguments passed to the function.
+// TODO: Remove this when `CallContext` is refactored to get call info upfront.
```
dherman, line 99: Idea: would it be useful to label TODO comments with labels for the future trigger that will allow the action to be taken, e.g. `TODO(remove-legacy)`?

dherman (approved):
Looks good! The `TODO` comment gave me one idea for a maintenance practice, see what you think.

PR #825, opened 2021-11-12
Simplify `call` and `construct` to only take a `AsRef<Handle<JsValue>>` that can be directly passed to args

kjvalencik:
A future PR will add a high level builder API for calling functions. In order to maintain a low level, high performance API, the existing APIs have been adapted to not require unnecessary allocation.

I went with `AsRef<[u8]>` so that `Vec` could be passed directly since this is an existing common pattern. The downside is that it could result in code bloat from monomorphization if many sized arrays are passed. We could minimize this later by moving the body of the function externally--this is a common pattern in `std`.


dherman (approved):
LGTM!

PR #829, opened 2021-11-24
Ergonomic function-calling API

dherman:
This PR implements an ergonomic function-calling API using an options object and method chaining. It offers two new methods, `JsFunction::call_with()` and `JsFunction::construct_with()`, along with an `Arguments` trait that enables lightweight syntax for heterogeneously-typed arguments expressed as Rust tuples. It also uses type inference to downcast the result of the function call to an expected type. All together, this makes calling JS functions from Rust much more ergonomic.

Before:

```rust
let this = cx.undefined();
let args = vec![
    cx.number(17).upcast(),
    cx.string("hello").upcast()
];
let v: Handle<JsArray> = f.call(&mut cx, this, args)?
    .downcast_or_throw()?;
```

After:

```rust
let v: Handle<JsArray> = f
    .call_with(&cx)
    .args((cx.number(17), cx.string("hello"))
    .apply(&mut cx)?;
```

Previous attempts:
- https://github.com/neon-bindings/neon/pull/796
- https://github.com/neon-bindings/neon/pull/816
- https://github.com/neon-bindings/neon/pull/817

Also related:
- https://github.com/neon-bindings/neon/pull/825


src/types/function/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Replaces the arguments list with the given arguments.
+    pub fn args<A: Arguments<'a>>(&mut self, args: A) -> &mut Self {
+        self.args = args.into_args_vec();
+        self
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is downcast to the type `V`, throwing a `TypeError` if the downcast fails.
+    pub fn apply<'b: 'a, V: Value, C: Context<'b>>(&self, cx: &mut C) -> JsResult<'b, V> {
+        let this = self.this.unwrap_or_else(|| cx.null().upcast());
```
kjvalencik, line 57: I think this should be `undefined` instead of `null`. We should also probably have a test that the correct `this` is being set; I don't see one.

src/types/function/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+        let v: Handle<JsValue> = self.callee.call(cx, this, &self.args)?;
+        v.downcast_or_throw(cx)
+    }
+
+    /// Make the function call for side effect, discarding the result value. This method is
+    /// preferable to [`apply()`](CallOptions::apply) when the result value isn't needed,
+    /// since it doesn't require specifying a result type.
+    pub fn exec<'b: 'a, C: Context<'b>>(&self, cx: &mut C) -> NeonResult<()> {
+        let this = self.this.unwrap_or_else(|| cx.null().upcast());
```
kjvalencik, line 66: Same as above.

src/types/function/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    /// Replaces the arguments list with the given arguments.
+    pub fn args<A: Arguments<'a>>(&mut self, args: A) -> &mut Self {
+        self.args = args.into_args_vec();
+        self
+    }
+
+    /// Make the constructor call. If the function returns without throwing, returns
+    /// the resulting object.
+    pub fn apply<'b: 'a, C: Context<'b>>(&self, cx: &mut C) -> JsResult<'b, JsObject> {
```
kjvalencik, line 109: How come we don't downcast on this one?

```rust
let global = cx.global();
let array_ctor: Handle<JsFunction> = global.get(&mut cx, "Array");
let arr: Handle<JsArray> = array_ctor.construct_with(&cx).arg(cx.number(42)).apply(&mut cx);
```

A more realistic example would probably be calling a method that returns a `JsBox`.

src/types/function/mod.rs
dherman: I agree, I wonder why I had the instinct to choose `null`. Here's I think a strong argument in favor of `undefined`:

```
% node
Welcome to Node.js v16.8.0.
Type ".help" for more information.
> function strict() { 'use strict'; return this }
undefined
> strict()
undefined
```

src/types/function/mod.rs
dherman: The only downside is that it's usually going to be `JsObject` but you'll sometimes have to manually type-annotate to assist inference:

```rust
let o: Handle<JsObject> = cx
    .argument::<JsFunction>(0)?
    .construct_with(&cx)
    .arg(cx.number(0.0))
    .apply(&mut cx)?;
```

But I think it's worth it so that this API consistently never requires you to use `downcast_or_throw()`.

kjvalencik (approved):

PR #835, opened 2021-12-05
Replace Electron test with Playwright

### LINKED ISSUE #834: Test needs update due to Spectron will be deprecated
[According to Electron team, Spectron will be deprecated on February 1st, 2022](https://www.electronjs.org/blog/spectron-deprecation-notice). Currently neon is using Spectron as a test framework for testing on Electron. It should be replaced to an alternative.

I will try to rewrite the test with [Playwright​](https://playwright.dev/) as possible.

MikuroXina:
Resolves #834.


test/electron/main.test.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+const assert = require("assert");
+const path = require("path");
+
+const { _electron: electron } = require("playwright");
+const { test } = require("@playwright/test");
+
+test("greeting", async () => {
+  const app = await electron.launch({
```
dherman, line 10: This is nice! I haven't used Playwright before, but this logic is very readable.

dherman (approved):
This looks good to me! Playwright seems like a good choice, and I have a good impression that it's well-maintained.

kjvalencik (approved):

Issue #706, opened 2021-03-24
bug(create-neon): Answering "no" leaves partial project on disk


kjvalencik:
The end of the `npm init` prompts ask if the `package.json` looks correct. Users may answer "no" and it terminates the child process with an exit code of `1`. `create-neon` properly aborts the remainder of the init; however, it has already created the directory and a partial `package.json`.

`create-neon` should delete the partial project directory on a non-zero exit code.

Caryyon:
Is this issue still valid? I just used `neon new foo` to test this, and it no longer gives me an option to give a 'yes' or 'no.' 
Happy to jump on this, but I believe it's been fixed.

kjvalencik:
@Caryyon Yes, this is still an issue. `neon new foo` and `npm init neon foo` are two different packages.

The `neon` CLI corresponds with `cli/` in the source repo and is _deprecated_. This issue is for `create-neon/` and is the recommended way to create a new Neon project.

To reproduce, run `npm init neon my-project` and then terminate in any way:
* Answering, "no"
* `Ctrl+C`
* kill the process

All of them will leave files on disk.

This should likely be implemented by initializing the project in a temp directory and once successfully bootstrapped, move the directory to the target location. We can try to delete the temp directory if the internal `npm init` exits with a non-zero code, but even if that fails or the process is killed, at worst we've left files in `/tmp`.

Caryyon:
I'll work on this!

kjvalencik:
Awesome! Thanks so much!

dherman:
@Caryyon Are you still interested in working on this? Let us know if you got stuck somewhere and we'd be happy to help.

kjvalencik:
This was fixed!

PR #840, opened 2022-01-06
solves issue #706

deepanshu44:
create-neon now deletes the partial project directory on a non-zero exit 
#706

### DIFF
--- a/create-neon/src/bin/create-neon.ts
+++ b/create-neon/src/bin/create-neon.ts
@@ -27,35 +27,50 @@ function inferVersions(): Versions {
     neon: versions.neon,
     napi: napi
   };
-}
+};
 
 async function main(name: string) {
+  let tmpFolderName:string = '';
+
   try {
+    // pretty lightweight way to check both that folder doesn't exist and
+    // that the user has write permissions.
     await fs.mkdir(name);
-  } catch (err) {
-    die(`Could not create \`${name}\`: ${err.message}`);
-  }
+    await fs.rmdir(name);
 
-  let pkg: Package;
+    tmpFolderName = await fs.mkdtemp(`${name}-`)
+
+  } catch (err) {
+     await die(`Could not create \`${name}\`: ${err.message}`, tmpFolderName);
+  };
+
+  let pkg: Package|undefined
 
   try {
-    pkg = await Package.create(name);
+      pkg = await Package.create(name,tmpFolderName);
+      await fs.mkdir(path.join(tmpFolderName, 'src'));
+
   } catch (err) {
-    die("Could not create `package.json`: " + err.message);
+      await die("Could not create `package.json`: " + err.message, tmpFolderName);
   }
-
-  await fs.mkdir(path.join(name, 'src'));
-
-  for (let source of Object.keys(TEMPLATES)) {
-    let target = path.join(name, TEMPLATES[source]);
+  if(pkg){
+    for (let source of Object.keys(TEMPLATES)) {
+    let target = path.join(tmpFolderName, TEMPLATES[source]);
     await expand(source, target, {
-      package: pkg,
-      versions: inferVersions()
-    });
-  }
+        package: pkg,
+        versions: inferVersions()
+      });
+    }
+   }
 
+  try{
+    await fs.rename(tmpFolderName, name)
+  }
+  catch(err){
+    await die(`Could not create \`${name}\`: ${err.message}`, tmpFolderName);
+  };
   console.log(`✨ Created Neon project \`${name}\`. Happy 🦀 hacking! ✨`);
-}
+};
 
 if (process.argv.length < 3) {
   console.error("✨ create-neon: Create a new Neon project with zero configuration. ✨");
@@ -65,6 +80,6 @@ if (process.argv.length < 3) {
   console.error("  name   The name of your Neon project, placed in a new directory of the same name.");
   console.error();
   process.exit(1);
-}
+};
 
 main(process.argv[2]);
--- a/create-neon/src/shell.ts
+++ b/create-neon/src/shell.ts
@@ -1,11 +1,12 @@
-import { spawn } from 'child_process';
-
+import { spawn } from "child_process";
+import {promises as fs} from "fs";
+import path from 'path';
 /**
  * Transparently shell out to an executable with a list of arguments.
  * All stdio is inherited directly from the current process.
  */
 export default function shell(cmd: string, args: string[], cwd: string): Promise<undefined> {
-  let child = spawn(cmd, args, { stdio: 'inherit', shell: true, cwd });
+  let child = spawn(cmd, args, { stdio: "inherit", shell: true, cwd });
 
   let resolve: (result: undefined) => void;
   let reject: (error: Error) => void;
@@ -15,19 +16,34 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
     reject = rej;
   });
 
-  child.on('exit', (code) => {
+  child.on("exit", async (code) => {
     if (code == null) {
-      process.exit(1);
-    }
+	reject(Error(`error code: ${code}`))
+    };
     if (code !== 0) {
-      process.exit(code);
-    }
+	//This will catch answering no and many other failures
+	reject(Error(`error code: ${code}`))
+    };
+
+    if (code === 0) {
+      try {
+          let data = await fs.readFile(path.join(cwd,'package.json'),'utf8')
+	  //Testing whether npm init was successful.
+	  //It will catch Ctrl+C and many other failures
+          let { description, author, license } = JSON.parse(data);
+          if ([description, author, license].includes(undefined)) {
+        	reject(Error(`error code: ${code}`))
+               }
+        } catch (e) {
+              reject(e)
+      };
+    };
+
     resolve(undefined);
   });
 
-  child.on('error', (error) => {
-    reject(error);
+  child.on("error", async (error) => {
+      reject(error);
   });
-
   return result;
-}
+};
--- a/create-neon/src/die.ts
+++ b/create-neon/src/die.ts
@@ -1,4 +1,13 @@
-export default function die(message: string): never {
+import { promises as fs } from 'fs';
+
+function deleteNeonDir(dir: string): Promise<void> {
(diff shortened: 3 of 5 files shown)


create-neon/src/shell.ts
```diff
@@ -1,11 +1,11 @@
-import { spawn } from 'child_process';
-
+import { spawn } from "child_process";
+import fs from "fs/promises";
```
kjvalencik, line 2: Unfortunately, the promises API is only exposed at `fs/promises` in newer node versions. In order to support older versions, this will need to be updated:

```suggestion
import { promises as fs } from "fs";
```

This can be seen in the failed CI unit tests.

create-neon/src/bin/create-neon.ts
```diff
@@ -30,7 +31,12 @@ function inferVersions(): Versions {
 }
 
 async function main(name: string) {
+  let workingDirPath: string = process.cwd()
+  let tmpDirPath: string = tmpdir()
+
   try {
+    //change PWD to temp
+    process.chdir(tmpDirPath)
     await fs.mkdir(name);
```
kjvalencik, line 40: In order to avoid potential collisions in the `tmpdir` can you use a random directory name?

https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

create-neon/src/bin/create-neon.ts
```diff
@@ -40,7 +46,9 @@ async function main(name: string) {
 
   try {
     pkg = await Package.create(name);
+
   } catch (err) {
+    await fs.rm(process.argv[2], { recursive: true, force: true })
```
kjvalencik, line 51: This should use `name` instead of the `argv` directly.

Also, I think it would be better to move the `fs.rm` into `die`.

create-neon/src/bin/create-neon.ts
```diff
@@ -53,7 +61,14 @@ async function main(name: string) {
       versions: inferVersions()
     });
   }
-
+  try{
+    // setting PWD back to working directory shouldn't be required
+    await fs.rename(tmpDirPath +'/'+name, workingDirPath+'/'+name)
```
kjvalencik, line 66: Please use `path.join` instead of concatenating strings. It would handle using the proper escaping and separators on all operating systems.

create-neon/src/shell.ts
```diff
@@ -15,19 +15,38 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
     reject = rej;
   });
 
-  child.on('exit', (code) => {
+  function deleteNeonDir(): Promise<void> {
+    return fs.rm(cwd, { force: true, recursive: true });
```
kjvalencik, line 19: Using `cwd` instead of a defined variable seems prone to error. What if we failed to changed directories?

create-neon/src/shell.ts
```diff
@@ -15,19 +15,38 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
...
+      await deleteNeonDir();
       process.exit(code);
     }
+
+    if (code === 0) {
+      try {
+          let data = await fs.readFile(`${cwd}/package.json`,'utf8')
+          let { description, author, license } = JSON.parse(data);
+          if ([description, author, license].includes(undefined)) {
+               throw new Error("Ctrl+C pressed");
```
kjvalencik, line 37: Is this error shown to the user? I think this could be multiple failures and not only `Ctrl+C`. Maybe something more generic?

```suggestion
               throw new Error("Failed to create node project with `npm init`");
```

create-neon/src/shell.ts
```diff
@@ -15,19 +15,38 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
...
     if (code !== 0) {
+      await deleteNeonDir();
       process.exit(code);
     }
+
+    if (code === 0) {
+      try {
+          let data = await fs.readFile(`${cwd}/package.json`,'utf8')
+          let { description, author, license } = JSON.parse(data);
+          if ([description, author, license].includes(undefined)) {
```
kjvalencik, line 36: Are these fields still defined if you hit `enter` without entering anything?

kjvalencik (changes requested):

create-neon/src/bin/create-neon.ts
```diff
@@ -30,7 +31,12 @@ function inferVersions(): Versions {
...
 
 async function main(name: string) {
+  let workingDirPath: string = process.cwd()
+  let tmpDirPath: string = tmpdir()
+
   try {
+    //change PWD to temp
+    process.chdir(tmpDirPath)
     await fs.mkdir(name);
   } catch (err) {
```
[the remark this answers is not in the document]
deepanshu44: A small question,
If file or folder exists, error message says- "file already exists"

Should we update the message to
"File or folder exists" 
Or
Should error message inform user that file, folder or symlink exists?

create-neon/src/shell.ts
```diff
@@ -15,19 +15,38 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
...
 
-  child.on('exit', (code) => {
+  function deleteNeonDir(): Promise<void> {
+    return fs.rm(cwd, { force: true, recursive: true });
+  }
+
+  child.on("exit", async (code) => {
     if (code == null) {
-      process.exit(1);
+      await deleteNeonDir();
```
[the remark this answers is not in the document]
deepanshu44: Alright, I will replace it with `die()`

create-neon/src/bin/create-neon.ts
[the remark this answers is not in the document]
kjvalencik: Current error looks like this, which is fine:

```
❯ npm init neon hello
❌ Could not create `hello`: EEXIST: file already exists, mkdir 'hello'
```

create-neon/src/shell.ts
deepanshu44: Yes, I tried doing the same with npm versions latest, LTS and a really old one 4.9.1 . Functionality is same and these fields do exist if we press `enter` without entering anything.

cli/src/async/child_process.ts
```diff
@@ -2,11 +2,11 @@ import { promisify } from 'util';
 import child from 'child_process';
 
 export function spawn(command: string, args?: string[], options?: child.SpawnOptions): Promise<number> {
-  return new Promise((resolve, reject) => {
-      let ps = child.spawn(command, args || [], options!);
-      ps.on('error', reject);
-      ps.on('close', resolve);
-  });
+    return new Promise((resolve, reject) => {
```
kjvalencik, line 5: Please revert these changes. This project uses 2  space indentation.

create-neon/src/bin/create-neon.ts
```diff
@@ -29,31 +30,60 @@ function inferVersions(): Versions {
   };
 }
 
+function deleteNeonDir(dir:string): Promise<void> {
```
kjvalencik, line 33: Nit, please leave a space after the `:` and before the type. This is repeated a few times in the PR.

Hopefully we can get `prettier` configured at some point!

create-neon/src/bin/create-neon.ts
```diff
@@ -29,31 +30,60 @@ function inferVersions(): Versions {
...
+
 async function main(name: string) {
+  let workingDirPath: string = process.cwd()
+  let tmpDirPath: string = os.tmpdir()
+  let tmpFolderName:string|undefined;
+
   try {
-    await fs.mkdir(name);
+    //check if folder or file with same name exists
+    await fs.mkdir(name)
```
kjvalencik, line 44: Semicolons are missing in many places. Can you please add them? Again, sorry for the lack of automatic formatting!

create-neon/src/bin/create-neon.ts
```diff
@@ -29,31 +30,60 @@ function inferVersions(): Versions {
...
+    await fs.rmdir(name)
```
kjvalencik, line 45: I like the `mkdir` and `rmdir` because it's a pretty lightweight way to check _both_ that it doesn't exist *and* that the user has write permissions.

Can you add a comment explaining this benefit?

create-neon/src/bin/create-neon.ts
```diff
@@ -29,31 +30,60 @@ function inferVersions(): Versions {
...
+
+    //change current working directory to /temp
+    process.chdir(path.join(tmpDirPath))
```
kjvalencik, line 48: `path.join` is unnecessary here. However, my preference is to use an absolute path for this directory instead of relying on the current working directory.

Can you do something like this instead?

```js
let tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), `${name}-`));
```

create-neon/src/bin/create-neon.ts
```diff
@@ -29,31 +30,60 @@ function inferVersions(): Versions {
...
+     await deleteNeonDir(path.join(tmpDirPath,tmpFolderName))
+    }
     die(`Could not create \`${name}\`: ${err.message}`);
   }
 
   let pkg: Package;
 
   try {
-    pkg = await Package.create(name);
+      pkg = await Package.create(name,path.join(tmpDirPath,tmpFolderName,'/'));
```
kjvalencik, line 63: Removing the trailing slash.

kjvalencik (changes requested):

create-neon/src/shell.ts
```diff
@@ -15,19 +16,31 @@ export default function shell(cmd: string, args: string[], cwd: string): Promise
...
-      process.exit(code);
-    }
+	reject(Error(`error code: ${code}`))
+    };
+
+    if (code === 0) {
+      try {
+          let data = await fs.readFile(path.join(cwd,'package.json'),'utf8')
+          let { description, author, license } = JSON.parse(data);
+          if ([description, author, license].includes(undefined)) {
```
kjvalencik, line 31: This could benefit from a comment explaining that this is testing to ensure `npm init` was successful. It will catch `Ctrl+C`, answering `no` and many other failures.

deepanshu44: See new [comments](https://github.com/neon-bindings/neon/commit/f7bb6e540301ce26638960dccf298e49650a9eb9)

kjvalencik:
Thanks for all your hard work on this!

Issue #843, opened 2022-01-15
Enforce isolation of instance data


dherman:
We don't have explicit logic for enforcing that instance data from one JS thread can't be shared with another JS thread. This [test case](https://github.com/dherman/instance-data-test) shows how a Neon module can store a rooted handle in a global cell from one JS thread and then attempt to access it in another thread. In my tests this seems to hang sometimes, and other times panics with a confusing error:

<details><summary>Stack trace</summary><pre>FATAL ERROR: v8::HandleScope::CreateHandle() Cannot create a handle without a HandleScope
 1: 0x1099e3d35 node::Abort() (.cold.1) [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 2: 0x1086d9239 node::Abort() [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 3: 0x1086d93af node::OnFatalError(char const*, char const*) [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 4: 0x108857b8e v8::Utils::ReportApiFailure(char const*, char const*) [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 5: 0x1089c9ce2 v8::internal::HandleScope::Extend(v8::internal::Isolate*) [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 6: 0x108859ca5 v8::HandleScope::CreateHandle(v8::internal::Isolate*, unsigned long) [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 7: 0x108694621 napi_get_reference_value [/Users/dherman/.volta/tools/image/node/16.8.0/bin/node]
 8: 0x10d317349 neon_runtime::napi::reference::get::h6858a65d09c3ab37 [/Users/dherman/Sources/instance-data-test/my-lib/index.node]
 9: 0x10d31398f my_lib::get::hef419b7ba393f378 [/Users/dherman/Sources/instance-data-test/my-lib/index.node]
10: 0x10d3120e4 neon::types::error::convert_panics::h4d928c6dbec5d1e5 [/Users/dherman/Sources/instance-data-test/my-lib/index.node]
11: 0x10d31294a neon::context::internal::Scope$LT$R$GT$::with::hb96cbf77f70198bc [/Users/dherman/Sources/instance-data-test/my-lib/index.node]
[...]

PR #847, opened 2022-01-26
fix(neon): Tag `Root` with an instance id and verify before using

kjvalencik:
## Overview

`Root` are Neon's mechanism for holding persistent references to values on the JavaScript heap. They are both `Send` and `Sync` to allow holding references across Rust threads. However, the current implementation has a soundness hole.

Resolves https://github.com/neon-bindings/neon/issues/843

## Problem

When using [Node worker threads](https://nodejs.org/api/worker_threads.html), there may be multiple JavaScript runtimes present in the process. As shown in https://github.com/neon-bindings/neon/issues/843, it is possible to use a `Root` with a `Context` from the wrong runtime and cause a crash/undefined behavior.

## Solution

This PR resolves the issue by panicking when attempting to dereference from the wrong module instance. In the future, we may want to add `try_` variants to the `Root` methods to allow error handling without a `panic`. However, in most cases, this will be a logic error that applications cannot recover.

## Approach

On Node-API 6+, we use [instance data](https://nodejs.org/api/n-api.html#environment-life-cycle-apis) to store a unique identifier initialized from a global counter. Each `Root` stores the identifier for comparing later when being used.

On previous versions, the [`ThreadId`](https://doc.rust-lang.org/std/thread/struct.ThreadId.html) is used. This may not be as robust and is sensitive to V8 implementation details.

### DIFF
--- a/crates/neon-runtime/src/napi/reference.rs
+++ b/crates/neon-runtime/src/napi/reference.rs
@@ -15,6 +15,8 @@ pub unsafe fn new(env: Env, value: Local) -> napi::Ref {
     result.assume_init()
 }
 
+/// # Safety
+/// Must only be used from the same module context that created the reference
 pub unsafe fn reference(env: Env, value: napi::Ref) -> usize {
     let mut result = MaybeUninit::uninit();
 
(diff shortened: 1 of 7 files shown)


dherman (approved):
This is so straightforward and non-clever (in a good way). LGTM!

PR #852, opened 2022-02-11
Add an NPM workspace and use prettier for non-rust source files

kjvalencik:
*Tip for reviewing*: Review the individual commits instead of the PR. Ignore the "Prettier!" commit where files are formatted in mass; there are no code changes in that commit.

This is a *very* large diff because of prettier and lock files, but the actual changes are all in config files.

* Make the root of the project an NPM workspace, creating a root `package-lock.json`
* Move all NPM packages in the project to the workspace; this deletes their individual lockfile
* Moved NPM tests from `src/lib.rs` executors into the workspace tests
* Updated CI to call `npm test` instead of `cargo test`
* Bumped package versions to unify the workspace and eliminate audit failures
* Disabled lib checks in `tsc` to fix compilation; we can revert this after `spectron` is removed
* Ran `prettier` on the entire project for all `js`, `ts`, `json`, and `yml` files (excluding generated)
* Added a prettier check to our CI lint job

@dherman I used the default prettier configuration. If there's something about the formatting you don't like, let me know and we can tweak it. I prefer tabs, but we were already using two-spaces.

As a nice benefit, it also fixes the legacy backend on newer NPM versions.


crates/neon-sys/native/binding.gyp
```diff
@@ -17,6 +17,7 @@
             'Debug': {
                 'msvs_settings': {
                     'VCCLCompilerTool': {
+                        'WholeProgramOptimization': 'false',
```
dherman, line 20: Given how painful it was to discover this, might be worth a comment explaining this is necessary for Node versions < x.y on Windows, maybe with a link to the issue.

src/lib.rs
```diff
@@ -485,27 +484,6 @@ mod tests {
             .success());
     }
 
-    fn cli_setup() {
```
dherman, line 488: GOOD RIDDANCE 🗑️

dherman (changes requested):
Lots of great cleanup! I'm excited to use workspaces and get rid of the recursive `cargo test`.

One suggestion in the comments, and then I'd suggest maybe just adding some brief docs to the main README on how to run the Neon tests (`npm test` in general, `npm test --workspace=test/foo` for individual subtests).

PR #854, opened 2022-02-14
fix(neon): Emit an unhandled rejection instead of an uncaught exception

kjvalencik:
## Why

Because of a Node bug, triggering a `fatal_exception` from a threadsafe function or task callback crashes the process with the default handler. This PR proposes that Neon trigger an `unhandledRejection` instead which avoids the bug.

## Impact

On recent versions of Node, this has the same intended consequence: the error is logged and the process terminates uncleanly. However, on older versions of Node, the rejection will only be logged.

## Why wasn't this caught?

While investigating the issue, I was curious why we didn't see it in unit tests. It appears that the internal code in the default exception handler is *necessary* for the bug. Since our tests replace this code, the bug never happens. The same would be true for any exception handler a user could write since it only uses internal/private APIs. This makes me wonder if the C++ crash is acceptable.

## Plan

It's strange to trigger an `unhandledRejection` for something that is unrelated to Promises. In the future, we may change this back to `uncaughtException`. However, in order to avoid a breaking change:

1. Add a feature flag that causes `uncaughtException` instead of `unhandledRejection`
2. Make the feature flag default in `create-neon` so that all new projects use it
3. Consider changing the default feature flag for everyone

We may choose to take a stance that it *shouldn't* impact many people and even though it's _technically_ a breaking change, the impact should be light and _could_ be considered a bug fix. In that case, we would make a feature flag to restore the original behavior and push out the change with only a minor revision.

### DIFF
--- a/crates/neon-runtime/src/napi/bindings/functions.rs
+++ b/crates/neon-runtime/src/napi/bindings/functions.rs
@@ -237,17 +237,6 @@ mod napi1 {
     );
 }
 
-#[cfg(feature = "napi-3")]
-mod napi3 {
-    use super::super::types::*;
-
-    generate!(
-        extern "C" {
-            fn fatal_exception(env: Env, err: Value) -> Status;
-        }
-    );
-}
-
 #[cfg(feature = "napi-4")]
 mod napi4 {
     use super::super::types::*;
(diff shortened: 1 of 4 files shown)


dherman (approved):
LGTM!

PR #856, opened 2022-02-19
Add FreeBSD

MikaelUrankar:


MikaelUrankar:
Your CI seems broken

kjvalencik:
This looks good, but FYI, this is updating the legacy backend which is deprecated and will be removed in 1.0. Is there a reason your project is unable to move to the Node-API backend?

Concerning CI, only Windows is broken. There is a fix in another PR that I haven't cherry picked. It broke when GitHub upgraded the Windows runners. Unfortunately, the legacy backend is very sensitive to versions.

MikaelUrankar:
> This looks good, but FYI, this is updating the legacy backend which is deprecated and will be removed in 1.0. Is there a reason your project is unable to move to the Node-API backend?

It's not my project, signal-desktop uses this crate: https://github.com/signalapp/ringrtc/blob/master/src/rust/Cargo.toml#L59

kjvalencik:
Thanks for the info!

PR #861, opened 2022-03-01
Support for `bigint`

zone117x:
Initial support for `JsBigInt` (neon) / `bigint` (js). So far implemented wrappers for creating and getting bigints using `i64` and `u64`.

TODO:
* [ ] Unit tests
* [ ] Doc updates
* [ ] Support arbitrary-length bigints via [`napi_create_bigint_words`](https://nodejs.org/api/n-api.html#napi_create_bigint_words)
  * [ ] Support rust `u128` and `i128` types by automatically using the above arbitrary-length API

Please let me know if this PR is headed in the right direction. I created it against the `next/0.10` branch as I had difficulties getting the `main` branch working correctly with the existing project I have to test this feature.

### DIFF
--- a/crates/neon-runtime/src/napi/bigint.rs
+++ b/crates/neon-runtime/src/napi/bigint.rs
@@ -35,7 +35,8 @@ pub unsafe fn new_bigint_from_u64(env: Env, value: u64) -> Local {
 pub unsafe fn value_i64(env: Env, p: Local) -> (i64, bool) {
     let mut value: i64 = 0;
     let mut lossless = false;
-    let status = napi::get_value_bigint_int64(env, p, &mut value as *mut _, &mut lossless as *mut _);
+    let status =
+        napi::get_value_bigint_int64(env, p, &mut value as *mut _, &mut lossless as *mut _);
     assert_eq!(status, napi::Status::Ok);
     (value, lossless)
 }
@@ -49,7 +50,8 @@ pub unsafe fn value_i64(env: Env, p: Local) -> (i64, bool) {
 pub unsafe fn value_u64(env: Env, p: Local) -> (u64, bool) {
     let mut value: u64 = 0;
     let mut lossless = false;
-    let status = napi::get_value_bigint_uint64(env, p, &mut value as *mut _, &mut lossless as *mut _);
+    let status =
+        napi::get_value_bigint_uint64(env, p, &mut value as *mut _, &mut lossless as *mut _);
     assert_eq!(status, napi::Status::Ok);
     (value, lossless)
 }
--- a/crates/neon-runtime/src/napi/bigint.rs
+++ b/crates/neon-runtime/src/napi/bigint.rs
@@ -1,0 +1,29 @@
+use crate::napi::bindings as napi;
+use crate::raw::{Env, Local};
+use std::mem::MaybeUninit;
+
+/// Create a new date object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+pub unsafe fn new_bigint(env: Env, value: i64) -> Local {
+    let mut local = MaybeUninit::zeroed();
+    let status = napi::create_bigint_int64(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::Status::Ok);
+    local.assume_init()
+}
+
+/// Get the value of a date object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+/// `Local` must be an NAPI value associated with the given `Env`
+pub unsafe fn value_i64(env: Env, p: Local) -> i64 {
+    let mut value: i64 = 0;
+    let mut lossless = false;
+    let status = napi::get_value_bigint_int64(env, p, &mut value as *mut _, &mut lossless as *mut _);
+    assert_eq!(status, napi::Status::Ok);
+    value
+}
--- a/crates/neon-runtime/src/napi/bigint.rs
+++ b/crates/neon-runtime/src/napi/bigint.rs
@@ -2,7 +2,7 @@ use crate::napi::bindings as napi;
 use crate::raw::{Env, Local};
 use std::mem::MaybeUninit;
 
-/// Create a new date object
+/// Create a new bigint object
 ///
 /// # Safety
 ///
@@ -14,16 +14,42 @@ pub unsafe fn new_bigint(env: Env, value: i64) -> Local {
     local.assume_init()
 }
 
-/// Get the value of a date object
+/// Create a new bigint object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+pub unsafe fn new_bigint_from_u64(env: Env, value: u64) -> Local {
+    let mut local = MaybeUninit::zeroed();
+    let status = napi::create_bigint_uint64(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::Status::Ok);
+    local.assume_init()
+}
+
+/// Get the value of a bigint object
 ///
 /// # Safety
 ///
 /// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
 /// `Local` must be an NAPI value associated with the given `Env`
-pub unsafe fn value_i64(env: Env, p: Local) -> i64 {
+pub unsafe fn value_i64(env: Env, p: Local) -> (i64, bool) {
     let mut value: i64 = 0;
     let mut lossless = false;
     let status = napi::get_value_bigint_int64(env, p, &mut value as *mut _, &mut lossless as *mut _);
     assert_eq!(status, napi::Status::Ok);
-    value
+    (value, lossless)
+}
+
+/// Get the value of a bigint object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+/// `Local` must be an NAPI value associated with the given `Env`
+pub unsafe fn value_u64(env: Env, p: Local) -> (u64, bool) {
+    let mut value: u64 = 0;
+    let mut lossless = false;
+    let status = napi::get_value_bigint_uint64(env, p, &mut value as *mut _, &mut lossless as *mut _);
+    assert_eq!(status, napi::Status::Ok);
+    (value, lossless)
 }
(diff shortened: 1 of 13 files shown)


crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -336,6 +336,38 @@ mod napi6 {
...
+
+            fn create_bigint_int64(env: Env, value: i64, result: *mut Value) -> Status;
+
+            fn create_bigint_uint64(env: Env, value: u64, result: *mut Value) -> Status;
+
+            fn create_bigint_words(
+                env: Env,
+                sign_bit: isize,
+                word_count: usize,
+                words: *mut u64,
```
kjvalencik, line 348:
```suggestion
                words: *const u64,
```

Also, `result` is missing. I recommend using the `bindings.rs` output from `nodejs-sys` as a reference.

```sh
cat ./target/debug/build/nodejs-sys-*/out/bindings.rs
```

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -336,6 +336,38 @@ mod napi6 {
...
+                env: Env,
+                value: Value,
+                result: *mut u64,
+                lossless: *mut bool,
+            ) -> Status;
+
+            fn get_value_bigint_words(
+                env: Env,
+                value: Value,
+                sign_bit: *mut isize,
```
kjvalencik, line 368: Same as previous comment, `c_int` or `i32`.

crates/neon-runtime/src/napi/bindings/functions.rs
```diff
@@ -336,6 +336,38 @@ mod napi6 {
...
+                words: *mut u64,
```
kjvalencik, line 369: Missing `word_count`.

crates/neon-runtime/src/napi/bigint.rs
```diff
@@ -0,0 +1,57 @@
+use crate::napi::bindings as napi;
+use crate::raw::{Env, Local};
+use std::mem::MaybeUninit;
+
+/// Create a new bigint object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+pub unsafe fn new_bigint(env: Env, value: i64) -> Local {
```
kjvalencik, line 10: Nit, name this `new` since it's namespaced by the `bigint` module.

```
use neon_runtime::bigint;

bigint::new(...)
```

crates/neon-runtime/src/napi/bigint.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut local = MaybeUninit::zeroed();
+    let status = napi::create_bigint_int64(env, value, local.as_mut_ptr());
+    assert_eq!(status, napi::Status::Ok);
```
kjvalencik, line 13: I haven't checked to see if this can ever throw an exception--which is necessary for this `assert` to be safe. Can you double check?

crates/neon-runtime/src/napi/bigint.rs
```diff
@@ -0,0 +1,57 @@
...
+    local.assume_init()
+}
+
+/// Get the value of a bigint object
+///
+/// # Safety
+///
+/// `env` is a raw pointer. Please ensure it points to a napi_env that is valid for the current context.
+/// `Local` must be an NAPI value associated with the given `Env`
+pub unsafe fn value_i64(env: Env, p: Local) -> (i64, bool) {
```
kjvalencik, line 35: Can you return a `Result` instead? It feels like more idiomatic Rust to me.

```suggestion
pub unsafe fn value_i64(env: Env, p: Local) -> Result<i64, i64> {
```

`Result::Err` indicates that loss occurred.

src/types/bigint.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.0
+    }
+
+    fn from_raw(_: Env, h: raw::Local) -> Self {
+        JsBigInt(h)
+    }
+}
+
+#[derive(Debug)]
+pub struct GetBigIntLossyValueResult<T> {
```
kjvalencik, line 36: It's fairly common in Rust to put the data in the `Err` variant when it is still usable after a partial failure. For example, see [`LockResult`](https://doc.rust-lang.org/std/sync/type.LockResult.html) on `Mutex`.

I think it would be more idiomatic to create an `Error` wrapper (mostly for clear printing and `?` error propagation) that indicates data loss. If the user doesn't mind the loss, they can add `.unwrap_or_else(Error::into_inner)`.

```rust
struct Error<T>(T);

impl<T> Error<T> {
    fn into_inner(self) -> T {
        self.0
    }
}

impl<T> std::error::Error for Error {}
```

src/types/bigint.rs
```diff
@@ -0,0 +1,107 @@
...
+impl<T> fmt::Display for GetBigIntLossyValueResult<T>
+where
+    T: fmt::Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.value)
+    }
+}
+
+impl Into<i64> for GetBigIntLossyValueResult<i64> {
```
kjvalencik, line 51: Typically `From` is implemented instead of `Into` since there is a reflexive implementation of `Into`.

```suggestion
impl From<GetBigIntLossyValueResult<i64>> for i64 {
    fn from(other: GetBigIntLossyValueResult<i64>) -> Self {
        self.value
    }
}
```

src/types/bigint.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl Into<u64> for GetBigIntLossyValueResult<u64> {
+    fn into(self) -> u64 {
+        self.value
+    }
+}
+
+impl JsBigInt {
+    pub fn new<'a, C: Context<'a>, T: Into<i64>>(cx: &mut C, x: T) -> Handle<'a, JsBigInt> {
```
kjvalencik, line 64: What do you think about making this generic over a trait? It could work well to allow providing an implementation for `num-bigint`.

Nit, `n` is preferable to `x` for consistency with other code in Neon.

src/types/bigint.rs
```diff
@@ -0,0 +1,107 @@
...
+        let env = cx.env().to_raw();
+        let value = x.into();
+        let local = unsafe { neon_runtime::bigint::new_bigint(env, value) };
+        let bigint = Handle::new_internal(JsBigInt(local));
+        bigint
```
kjvalencik, line 69: Nit let the type be returned directly instead of the unused variable binding.

```suggestion
        Handle::new_internal(JsBigInt(local))
```

src/types/bigint.rs
```diff
@@ -0,0 +1,107 @@
...
+        GetBigIntLossyValueResult {
+            value: result.0,
+            lossless: result.1,
+        }
+    }
+}
+
+impl ValueInternal for JsBigInt {
+    fn name() -> String {
+        "bigint".to_string()
```
kjvalencik, line 101: Match the JavaScript capitalization.

```suggestion
        "BigInt".to_string()
```

kjvalencik (changes requested):
@zone117x Thanks so much for this amazing contribution! I appreciate the thought that went into it! I have left some comments.

Additionally, you can remove `neon-sys` / `legacy` code. This backend is no longer supported and we can set some feature flags to avoid it being compiled.

I think my main question is what the API should look like. Is it better to have separate constructors for each supported `from_` type or is it better to have a `trait` and a single generic constructor? I think I favor being generic. What do you think?

crates/neon-runtime/src/napi/bigint.rs
zone117x: Great idea 👍 
I struggled with what to return here.

zone117x:
> I think my main question is what the API should look like. Is it better to have separate constructors for each supported `from_` type or is it better to have a `trait` and a single generic constructor? I think I favor being generic. What do you think?

Good question, I've spent some time thinking about that myself. In my opinion, the Neon code for bigint constructors should generally match the napi code -- there are three ways to create a bigint: [napi_create_bigint_int64](https://nodejs.org/api/n-api.html#napi_create_bigint_int64), [napi_create_bigint_uint64](https://nodejs.org/api/n-api.html#napi_create_bigint_uint64), and [napi_create_bigint_words](https://nodejs.org/api/n-api.html#napi_create_bigint_words).

The first two (construct bigint directly from a 64-bit integer) seem like a very common use-case. E.g. a lot of codebases use 64-bit ints where js `number` doesn't work, but arbitrary-length ints are overkill and can hurt performance. I think we could use single trait + generic to combine the u64 and i64 for these two cases, but I don't think we should lump in arbitrary-sized ints into it.

What are your thoughts?

crates/neon-runtime/src/napi/bigint.rs
zone117x: @kjvalencik do you have some guidance on how to determine this? The [napi docs for the function](https://nodejs.org/api/n-api.html#napi_create_bigint_int64) do not mention exceptions -- do I need to check the nodejs or v8 source code to determine this?

kjvalencik: Unfortunately, it's not super clear. It's a manual task that's a combination of reading the Node/Node-API source and sometimes the JS spec.

I expect that this *can't* ever fail except with bad data (which Rust type system protects against).

kjvalencik:
I agree with u64 and i64 being the common use cases and making that nice. We might be able to leave off arbitrary precision until a someone brings a compelling use case. Technically all of this could be accomplished with an `ArrayBuffer` and the `BigInt` constructor, so I think it's okay if the first revision is incomplete.

With that said, I don't think putting it behind a single trait and method is harmful. The implementation would be generic and specialized and shouldn't hurt performance. Did you have some example in mind where it would be harmful?

The disadvantage to a single generic method is that the API is less discoverable. It requires reading the trait definition and implementations to know that you can send a `u64` to the API.

dherman:
This is awesome!

I think I favor the trait approach for the constructor, too. It's mildly more future-proof for other argument types (like strings, or maybe even third-party math crates) without needing to keep growing the API surface, and it makes for a nicely clean and general single `cx.bigint()` entry point too.

kjvalencik:
@zone117x Sorry! It was closed unintentionally when the `next/0.10` branch was deleted after merging it into `main`.

I've re-opened this PR and changed the target branch to `main`.

kjvalencik:
@zone117x is this still something you are interested in working on? If not, I have some bandwidth to take it over--giving you credit, of course!

Cheers.

zone117x:
> @zone117x is this still something you are interested in working on? If not, I have some bandwidth to take it over--giving you credit, of course!
> 
> Cheers.

Sorry, I do not have bandwidth to continue working on this. Please feel free to take over!

kjvalencik:
@zone117x Thanks again for working on this! This feature was merged in https://github.com/neon-bindings/neon/pull/963 with you as co-author. Cheers!

PR #877, opened 2022-03-22
neon: Relax the lifetime constraints on `TypedArray` borrows

kjvalencik:
The current lifetimes are overly restrictive. The reference is valid for as long as the handle is valid. Since the `Context` is always the most narrow scope, it will be valid for at *least* as long as `Context`, even if the `Handle` type is dropped. This allows downcasting and returning a borrow within the same function.

See this thread for more details. https://rust-bindings.slack.com/archives/C0HE1SA65/p1647378845225649

### DIFF
--- a/src/types/buffer/mod.rs
+++ b/src/types/buffer/mod.rs
@@ -22,17 +22,17 @@ pub trait TypedArray: private::Sealed {
     ///
     /// This may not be used if a mutable borrow is in scope. For the dynamically
     /// checked variant see [`TypedArray::try_borrow`].
-    fn as_slice<'a: 'b, 'b, C>(&'b self, cx: &'b C) -> &'b [Self::Item]
+    fn as_slice<'cx, 'a, C>(&self, cx: &'a C) -> &'a [Self::Item]
     where
-        C: Context<'a>;
+        C: Context<'cx>;
 
     /// Statically checked mutable borrow of binary data.
     ///
     /// This may not be used if any other borrow is in scope. For the dynamically
     /// checked variant see [`TypedArray::try_borrow_mut`].
-    fn as_mut_slice<'a: 'b, 'b, C>(&'b mut self, cx: &'b mut C) -> &'b mut [Self::Item]
+    fn as_mut_slice<'cx, 'a, C>(&mut self, cx: &'a mut C) -> &'a mut [Self::Item]
     where
-        C: Context<'a>;
+        C: Context<'cx>;
 
     /// Dynamically checked immutable borrow of binary data, returning an error if the
     /// the borrow would overlap with a mutable borrow.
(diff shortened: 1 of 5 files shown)


src/types/buffer/mod.rs
```diff
@@ -22,43 +22,40 @@ pub trait TypedArray: private::Sealed {
     ///
     /// This may not be used if a mutable borrow is in scope. For the dynamically
     /// checked variant see [`TypedArray::try_borrow`].
-    fn as_slice<'a: 'b, 'b, C>(&'b self, cx: &'b C) -> &'b [Self::Item]
+    fn as_slice<'cx, 'a, C>(&self, cx: &'a C) -> &'a [Self::Item]
```
kjvalencik, line 25: `'a` is renamed to `'cx` across the board for clarity. `'b` becomes `'a` since that's the only lifetime we need.

src/types/buffer/mod.rs
```diff
@@ -22,43 +22,40 @@ pub trait TypedArray: private::Sealed {
...
     where
-        C: Context<'a>;
+        C: Context<'cx>;
 
     /// Statically checked mutable borrow of binary data.
     ///
     /// This may not be used if any other borrow is in scope. For the dynamically
     /// checked variant see [`TypedArray::try_borrow_mut`].
-    fn as_mut_slice<'a: 'b, 'b, C>(&'b mut self, cx: &'b mut C) -> &'b mut [Self::Item]
+    fn as_mut_slice<'cx, 'a, C>(&mut self, cx: &'a mut C) -> &'a mut [Self::Item]
```
kjvalencik, line 33: The `'a: 'b` bound was unnecessarily complicated. It's not possible to have references that don't meet this requirement.

src/types/buffer/mod.rs
```diff
@@ -22,43 +22,40 @@ pub trait TypedArray: private::Sealed {
...
     /// the borrow would overlap with a mutable borrow.
     ///
     /// The borrow lasts until [`Ref`] exits scope.
     ///
     /// This is the dynamically checked version of [`TypedArray::as_slice`].
-    fn try_borrow<'a: 'b, 'b, C>(
-        &self,
-        lock: &'b Lock<'b, C>,
-    ) -> Result<Ref<'b, Self::Item>, BorrowError>
+    fn try_borrow<'cx, 'a, C>(&self, lock: &'a Lock<C>) -> Result<Ref<'a, Self::Item>, BorrowError>
```
kjvalencik, line 43: Elided the inner lifetime of `Lock` to simplify.

test/napi/src/js/objects.rs
```diff
@@ -189,3 +191,31 @@ pub fn write_buffer_with_borrow_mut(mut cx: FunctionContext) -> JsResult<JsUndef
...
+fn get_bytes<'cx, 'a, C>(cx: &'a mut C, v: Handle<JsValue>) -> NeonResult<Cow<'a, [u8]>>
+where
+    C: Context<'cx>,
+{
+    if let Ok(v) = v.downcast::<JsString, _>(cx) {
+        return Ok(Cow::Owned(v.value(cx).into_bytes()));
+    }
+
+    if let Ok(v) = v.downcast::<JsBuffer, _>(cx) {
+        return Ok(Cow::Borrowed(v.as_slice(cx)));
```
kjvalencik, line 206: This is the primary thing we are trying to allow; returning a reference to a handle that was downcast.

dherman: This is really cool!

dherman (approved):
I feel like I should have some suggestion, but borrowck did all the heavy lifting for me :P 

LGTM!

PR #879, opened 2022-03-25
`Object::call_with()`

### LINKED ISSUE #873: [Enhancement] Object::call_method
libsignal-client has a pretty ubiquitous helper `call_method`:

```rust
/// A convenience for calling a method on an object.
///
/// Equivalent to calling `get`, downcasting to a function, and then using `call`.
pub fn call_method<'a>(
    cx: &mut impl Context<'a>,
    this: Handle<'a, impl Object>,
    method_name: &str,
    args: impl AsRef<[Handle<'a, JsValue>]>,
) -> JsResult<'a, JsValue> {
    let method: Handle<JsFunction> = this.get(cx, method_name)?;
    method.call(cx, this, args)
}
```

This would be a useful convenience to have on Object, though whether it should be built on the low-level `call` or the high-level `call_with` is an open question. (I suppose I could have put ours on an ObjectExt trait rather than making it a top-level function too.)

dherman:
Add `Object::call_with()` convenience method to call a method on an object.

Example:

```rust
let s: Handle<JsString> = date.call_with(&mut cx, "toLocaleString")?
    .arg(cx.string("en-GB"))
    .apply(&mut cx)?;
```

Closes #873.

## Design questions

- Are we OK with the inconsistencies with `JsFunction::call_with()`?
  - This is fallible.
  - This takes a mutable reference to the context.
- Alternative names: `date.call_with()` vs `date.call_method()` vs `date.call_method_with()`

### DIFF
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -314,10 +314,7 @@ mod traits {
             K: PropertyKey,
         {
             let mut options = self.get::<JsFunction, _, _>(cx, method)?.call_with(cx);
-            options.this(Handle::new_internal(Self::from_raw(
-                cx.env(),
-                self.to_raw(),
-            )));
+            options.this(JsValue::new_internal(self.to_raw()));
             Ok(options)
         }
     }
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -308,7 +308,7 @@ mod traits {
             Root::new(cx, self)
         }
 
-        fn call_with<'a, C, K>(&self, cx: &mut C, method: K) -> NeonResult<CallOptions<'a>>
+        fn call_method_with<'a, C, K>(&self, cx: &mut C, method: K) -> NeonResult<CallOptions<'a>>
         where
             C: Context<'a>,
             K: PropertyKey,
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -137,7 +137,7 @@ mod traits {
     use crate::result::{NeonResult, Throw};
     use crate::types::function::CallOptions;
     use crate::types::utf8::Utf8;
-    use crate::types::{build, JsUndefined, JsValue, Value, JsFunction};
+    use crate::types::{build, JsFunction, JsUndefined, JsValue, Value};
     use neon_runtime::raw;
 
     #[cfg(feature = "napi-6")]
@@ -314,7 +314,10 @@ mod traits {
             K: PropertyKey,
         {
             let mut options = self.get::<JsFunction, _, _>(cx, method)?.call_with(cx);
-            options.this(Handle::new_internal(Self::from_raw(cx.env(), self.to_raw())));
+            options.this(Handle::new_internal(Self::from_raw(
+                cx.env(),
+                self.to_raw(),
+            )));
             Ok(options)
         }
     }
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -135,8 +135,9 @@ mod traits {
     use crate::context::Context;
     use crate::handle::{Handle, Managed, Root};
     use crate::result::{NeonResult, Throw};
+    use crate::types::function::CallOptions;
     use crate::types::utf8::Utf8;
-    use crate::types::{build, JsUndefined, JsValue, Value};
+    use crate::types::{build, JsUndefined, JsValue, Value, JsFunction};
     use neon_runtime::raw;
 
     #[cfg(feature = "napi-6")]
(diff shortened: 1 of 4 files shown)


kjvalencik:
I like `call_with` because it's short. `call_method_with` is way too long.

`call_method` also seems reasonable.

src/object/mod.rs
```diff
@@ -306,6 +307,19 @@ mod traits {
...
+        fn call_with<'a, C, K>(&self, cx: &mut C, method: K) -> NeonResult<CallOptions<'a>>
+        where
+            C: Context<'a>,
+            K: PropertyKey,
+        {
+            let mut options = self.get::<JsFunction, _, _>(cx, method)?.call_with(cx);
+            options.this(Handle::new_internal(Self::from_raw(
+                cx.env(),
+                self.to_raw(),
+            )));
```
kjvalencik, line 320: `Self::from_raw` might perform extra steps--for example, unpacking a pointer on `JsBox`--that are unnecessary. Upcasting to a `JsValue` should be sufficient for binding `this`.

```suggestion
            options.this(JsValue::new_internal(self.to_raw()));
```

kjvalencik (changes requested):
@dherman This looks great! Other than the small suggestion to simplify, my only open question is on naming.

I'm torn between `call_with` and `call_method`.

mathieulj:
> Alternative names: date.call_with() vs date.call_method() vs date.call_method_with()

I personally prefer `call_method()` as it makes more of a distinction from `JsFunction::call_with()` and communicates the extra layer while scanning/reviewing code. I feel as though I would get caught by confusion a few times when reviewing if the two `.call_with` methods return different types. 

> call_method_with is way too long.

I do agree it's a little unwieldy, but maybe not a deal breaker.

mathieulj:
> Apologies if this seems like too much of a nit, but the two methods return the _same_ type, [`CallOptions`](https://docs.rs/neon/latest/neon/types/function/struct.CallOptions.html). However, I think your argument stands in respect to method arguments.

No worries, my thoughts are in flux, I value differences in opinions and I don't feel particularly strongly about it.

The two both return `CallOptions` but it is wrapped in a Result for this new one (`Object::call_with`) and returned bare for [`JsFunction::call_with`](https://docs.rs/neon/latest/neon/types/struct.JsFunction.html#method.call_with). In most code bases, this will translate to just an extra `?` and is probably not something to fret about on it's own. My point was about consistency and does also apply to arguments as you pointed out.

kjvalencik:
I'm sufficiently convinced that `call_with` is the least optimal choice. I slightly favor `call_method`, but can appreciate the symmetry in `call_method_with`. Maybe the extra 5 characters really don't make that much of a difference?

```diff
 // I don't mind it at all when chaining.
 let n: Handle<JsNumber> = obj
-    .call_method(&mut cx, "getN")?
+    .call_method_with(&mut cx, "getN")?
     .arg(cx.number(42))
     .apply(&mut cx)?;
    
// It's a little awkward on one line, but not terrible
- let n: Handle<JsNumber> = obj.call_method(&mut cx, "getN")?.applyt(&mut cx)?;
+ let n: Handle<JsNumber> = obj.call_method_with(&mut cx, "getN")?.applyt(&mut cx)?;
```

dherman:
Thanks for the example, @kjvalencik! I'm leaning towards `call_method_with()` for the consistency, but I feel like only real-world experience will tell. Should we go with `call_method_with()` for now, with an option to rename it via alias-and-deprecate later if usage experience shows it's too unwieldy?

kjvalencik:
@dherman I like that plan!

src/object/mod.rs
```diff
@@ -135,8 +135,9 @@ mod traits {
     use crate::context::Context;
     use crate::handle::{Handle, Managed, Root};
     use crate::result::{NeonResult, Throw};
+    use crate::types::function::CallOptions;
     use crate::types::utf8::Utf8;
-    use crate::types::{build, JsUndefined, JsValue, Value};
+    use crate::types::{build, JsFunction, JsUndefined, JsValue, Value};
```
kjvalencik, line 140: Nit, these additions are throwing unused warnings on the legacy backend. Probably not a big deal because that's getting removed very soon.

src/object/mod.rs
```diff
@@ -306,6 +307,16 @@ mod traits {
...
             Root::new(cx, self)
         }
+
+        fn call_method_with<'a, C, K>(&self, cx: &mut C, method: K) -> NeonResult<CallOptions<'a>>
+        where
+            C: Context<'a>,
+            K: PropertyKey,
+        {
+            let mut options = self.get::<JsFunction, _, _>(cx, method)?.call_with(cx);
+            options.this(JsValue::new_internal(self.to_raw()));
```
kjvalencik, line 317: Flagging this for follow-up. `JsValue::new_internal` should be marked `unsafe` since it allows arbitrarily extending a lifetime.

This usage of it is *safe* because it's bound by a new `Context` which will always be the shortest possible.

kjvalencik (approved):

src/object/mod.rs
dherman: death to legacy…

PR #884, opened 2022-04-13
Re-organize the Neon directory structure a bit

kjvalencik:
This is a little more consistent with workspaces. Currently `pkgs` only has a single module, but I will move `cargo-cp-artifact` in a separate PR.


dherman (approved):
Great improvements! I agree with removing `neon::meta::BUILD_PROFILE`.

PR #888, opened 2022-04-15
Remove scope abstraction that was only necessary for the legacy backend

kjvalencik:
* Active check is no longer necessary since changing `execute_scoped` and `compute_scoped` to take `&mut self`
* Remove scope metadata since `compute_scoped` and `execute_scoped` are the *only* places scope needs to be tracked
* Apply simplifications everywhere from the first two changes

In order to use `Drop` for scopes--which is a little cleaner and protects us from `panic` leaks--`env` is stored both in the scope and on the `Context`. This additional pointer is a small cost and overall much smaller than the removed `ScopeMetadata`.

### DIFF
--- a/crates/neon/src/context/internal.rs
+++ b/crates/neon/src/context/internal.rs
@@ -1,15 +1,11 @@
-use std::{
-    cell::{Cell, RefCell},
-    ffi::c_void,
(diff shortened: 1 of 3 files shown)


dherman (approved):
🚢

Issue #887, opened 2022-04-15
Update `argument`, `argument_opt` and `len` to use usize


kjvalencik:
For compatibility with the legacy backend, they cast `usize` to `i32`. While we have the opportunity for breaking changes, they should be changed to `usize`.

PR #889, opened 2022-04-15
argument, argument_ops, and len i32 -> usize

lazops:
Resolves: #887

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -653,7 +653,7 @@ impl<'a, T: This> CallContext<'a, T> {
 
     /// Indicates the number of arguments that were passed to the function.
     pub fn len(&self) -> usize {
-        self.info.len(self) as usize
+        self.info.len(self)
     }
 
     /// Indicates if no arguments were passed to the function.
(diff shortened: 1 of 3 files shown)


src/context/mod.rs
```diff
@@ -652,8 +652,8 @@ impl<'a, T: This> CallContext<'a, T> {
     }
 
     /// Indicates the number of arguments that were passed to the function.
-    pub fn len(&self) -> i32 {
-        self.info.len(self)
+    pub fn len(&self) -> usize {
+        self.info.len(self) as usize
```
kjvalencik, line 656: Can you change this on the `CallbackInfo::len` and `sys::call::len` as well instead of casting?

test/napi/src/js/functions.rs
```diff
@@ -144,7 +144,7 @@ trait CheckArgument<'a> {
 
 impl<'a, T: This> CheckArgument<'a> for CallContext<'a, T> {
     fn check_argument<V: Value>(&mut self, i: i32) -> JsResult<'a, V> {
-        self.argument::<V>(i)
+        self.argument::<V>(i as usize)
```
kjvalencik, line 147: Can you update the `CheckArgument` trait instead of casting?

src/context/mod.rs
kjvalencik, line 656: This cast shouldn't be necessary.

kjvalencik (approved):

PR #891, opened 2022-04-15
Add Object.freeze and Object.seal

lazops:
This change adds the bindings for Object.freeze and Object.seal as well as their respective integration tests.

### DIFF
--- a/src/object/mod.rs
+++ b/src/object/mod.rs
@@ -36,7 +36,7 @@ use crate::{
     context::{internal::Env, Context},
     handle::{Handle, Managed, Root},
     result::{NeonResult, Throw},
-    sys::{self, raw},
+    sys::{self, raw, Status},
     types::{build, function::CallOptions, utf8::Utf8, JsFunction, JsUndefined, JsValue, Value},
 };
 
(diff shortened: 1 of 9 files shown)


src/object/mod.rs
```diff
@@ -188,6 +188,20 @@ pub trait Object: Value {
         })
     }
 
+    #[cfg(feature = "napi-8")]
+    fn freeze<'a, C: Context<'a>>(&self, cx: &mut C) -> bool {
```
kjvalencik, line 192: I think we can have a better return type on failure (e.g., when trying to freeze a Uint8Array). What do you think about following the [ECMAScript spec](https://262.ecma-international.org/12.0/#sec-object.freeze) and throwing a TypeError?

What do you think of `JsResult<&self>`? Returning `self` would also allow chaining:

```rust
let o = cx.empty_object()
    .freeze(&mut cx)?
    .root(&mut cx);
```

src/sys/object.rs
```diff
@@ -10,6 +10,18 @@ pub unsafe fn new(out: &mut Local, env: Env) {
     napi::create_object(env, out as *mut _);
 }
 
+#[cfg(feature = "napi-8")]
+pub unsafe fn freeze(env: Env, obj: Local) -> bool {
+    let status = napi::object_freeze(env, obj);
+    status == napi::Status::Ok
```
kjvalencik, line 16: I think `GenericFailure` and `ExceptionPending` are the only valid responses. Everything else would be an error on Neon's part.

Maybe we could return `Status` and let the higher level handle that match?

test/napi/lib/objects.js
```diff
@@ -28,6 +28,33 @@ describe("JsObject", function () {
     );
   });
 
+  it("freeze a JsObject", function () {
```
kjvalencik, line 31: Can you add a test for freezing something that can't be frozen, like an Array view?

kjvalencik (changes requested):

src/object/mod.rs
lazops: Would you be okay with returning a `NeonResult<&Self>`? `Object` does not implement `Managed`, which is required for `JsResult` to be implemented properly. Also, I think there could be some advantage to returning an owned `Self` instead of a reference to it, since the freeze would consume it and then output it again.

src/sys/object.rs
lazops: I agree this would be a good change.

test/napi/lib/objects.js
lazops: I can't seem to find any documentation on an `ArrayView`, and freezing arrays as well as array buffers works fine. Do you mind pointing me in the right direction here?

kjvalencik: [`ArrayBufferView`](https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView) is the name for typed arrays like `Uint32Array`.

```
$ node -e 'Object.freeze(new Uint32Array(8))'
[eval]:1
Object.freeze(new Uint32Array(8))
       ^

TypeError: Cannot freeze array buffer views with elements
    at Function.freeze (<anonymous>)
    at [eval]:1:8
    at Script.runInThisContext (node:vm:129:12)
    at Object.runInThisContext (node:vm:305:38)
    at node:internal/process/execution:75:19
    at [eval]-wrapper:6:22
    at evalScript (node:internal/process/execution:74:60)
    at node:internal/main/eval_string:27:3
```

src/object/mod.rs
kjvalencik: `NeonResult` should be fine. I don't have a strong sense of whether the chaining is valuable.

I do think it's a little more ergonomic to return a `TypeError` on failure instead of a `bool`--which matches the ECMAScript spec.

kjvalencik:
This isn't necessary in this PR, but it would be nice to bump the [Node-API version in create-neon](https://github.com/neon-bindings/neon/blob/main/create-neon/data/versions.json#L3) so that new projects with `npm init neon ...` have these features enabled.

lazops:
bumped now. should be good to go!

dherman:
I love the chaining API! Really nice work.

kjvalencik (approved):

PR #893, opened 2022-04-26
Simplify CI

kjvalencik:
* Remove clang dependency on Windows
* Drop Node 12, add Node 18
* Use latest npm 8
* Drop Rust beta channel from the build matrix

I looked through a lot of history and the best I could tell, testing on `beta` has never caught a bug. Nightly is nice for new linting, but I chose to drop beta since it's redundant with `nightly` and will shrink the build matrix some.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
         rust-toolchain: [nightly]
 
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
```
kjvalencik, line 20: Current LTS.

dherman (approved):
:shipit:

Issue #896, opened 2022-05-22
`JsArrayBuffer::external` soundness hole


Cassy343:
Undefined behavior can be triggered with a misuse of [`JsArrayBuffer::external`](https://docs.rs/neon/latest/neon/types/struct.JsBuffer.html#method.external) using safe code.

Rust:
```rs
use neon::prelude::*;
use std::{
    mem,
    sync::{Arc, Mutex},
    thread,
    time::Duration,
};

#[neon::main]
fn main(mut cx: ModuleContext) -> NeonResult<()> {
    let buffer = Arc::new(Mutex::new(vec![0u8, 1, 2, 3]));

    cx.export_function("get_buf", {
        let buffer = Arc::clone(&buffer);
        move |mut cx| {
            Ok(JsArrayBuffer::external(&mut cx, &mut **buffer.lock().unwrap()))
        }
    })?;

    thread::spawn(move || {
        thread::sleep(Duration::from_secs(3));
        drop(mem::take(&mut *buffer.lock().unwrap()));
        println!("Buffer dropped");
        // `buffer` is also implicitly dropped here, so the explicit drop isn't strictly necessary
    });

    Ok(())
}
```

JavaScript:
```js
const rust = require('.');
let buf = rust.get_buf();
setTimeout(() => console.log(buf), 5000);
```

Result:
```
$ node index.js 
Buffer dropped
ArrayBuffer { [Uint8Contents]: <d0 b3 53 06>, byteLength: 4 }
```

Cassy343:
There's another more subtle error (bug?) that shows up when you only add `'static`. From my testing, the v8 engine will crash if you create two array buffers with overlapping regions in memory, but the `'static` bound allows you to violate that like so:
```rs
use neon::prelude::*;

#[neon::main]
fn main(mut cx: ModuleContext) -> NeonResult<()> {
    cx.export_function("empty_buf", move |mut cx| {
        let buffer = JsArrayBuffer::external(
            &mut cx,
            &mut []
        );
        Ok(buffer)
    })?;

    Ok(())
}
```

Result:
```
$ node
Welcome to Node.js v16.13.0.
Type ".help" for more information.
> const rust = require('.');
undefined
> rust.empty_buf()
ArrayBuffer { [Uint8Contents]: <>, byteLength: 0 }
> rust.empty_buf()


#
# Fatal error in , line 0
# Check failed: result.second.
#
#
#
#FailureMessage Object: 0x7ffc8cf945f0
[...]

PR #897, opened 2022-05-23
soundness(neon): Require `T: 'static` on `JsArrayBuffer::external` and `JsBuffer::external`

kjvalencik:
Currently, types used as the backing storage of external `ArrayBuffer` do not need to be `'static`! This makes it trivial to create a soundness hole (use after free), with only safe Rust.

```rust
pub fn soundness_hole(mut cx: FunctionContext) -> JsResult<JsArrayBuffer> {
    let mut data = vec![0u8, 1, 2, 3];
    
    // Creating an external from `&mut [u8]` instead of `Vec<u8>` since there is a blanket impl
    // of `AsMut<T> for &mut T`
    let buf = JsArrayBuffer::external(&mut cx, data.as_mut_slice());

    // `buf` is still holding a reference to `data`!
    drop(data);

    Ok(buf)
}
```

Fixes https://github.com/neon-bindings/neon/issues/896

### DIFF
--- a/crates/neon/src/types/buffer/types.rs
+++ b/crates/neon/src/types/buffer/types.rs
@@ -44,7 +44,7 @@ impl JsBuffer {
     pub fn external<'a, C, T>(cx: &mut C, data: T) -> Handle<'a, Self>
     where
         C: Context<'a>,
-        T: AsMut<[u8]> + Send,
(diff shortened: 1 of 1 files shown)


kjvalencik:
Closing this PR. I'm going to reverse the cherry pick order so that `main` gets the `CHANGELOG`.

Issue #728, opened 2021-05-07
Feature request: expose per-instance data


jrose-signal:
N-API provides the rather limited `napi_set_instance_data` to allow associating a single word of data with an instance. Neon already uses this API to store its own per-instance data, but it would be great if that were exposed to users in a convenient way. A possible shape for the API:

```rust
pub struct PerInstanceKey<T: 'static>(PhantomData<&'static mut T>);
impl<T: 'static> PerInstanceKey<T> {
  pub const fn new() {
    Self(PhantomData)
  }
}

impl Context {
  pub fn set_instance_data<T: 'static>(&mut self, key: &'static PerInstanceKey<T>, value: T) -> Option<T> {
    let mut user_data: HashMap = &mut InstanceData::get(self).user_data;
    let previous_value = user_data.insert(key as *const (), Box::new(value as dyn Any));
    previous_value.map(|old_value| *old_value.downcast().expect("type-safe API"))
  }

  pub fn get_instance_data_mut<T: 'static>(&mut self, key: &'static PerInstanceKey<T>) -> Option<&mut T> {
    unimplemented!() // same as set_* simplified
  }
}
```

```rust
static MY_DATA: PerInstanceKey<String>;

cx.set_instance_data(&MY_DATA, "hello".to_string());
```

An alternate version of the API would include a function to construct the initial value if not present, more like `thread_local!`, but this version is the simplest I could think of.

kjvalencik:
@jrose-signal This looks like a good minimal API. I think it might need to return some type of borrow guard to ensure borrowed instance data lives long enough, unless `set_instance_data` is made to error if already set.

Also, I could see benefits of _only_ providing an immutable reference, otherwise, we would need to bind to the lifetime of `&mut Context` to prevent aliasing.

It would be really neat to make this typesafe without `Any` by making `Contex` generic over the type. Unfortunately, I think that would be incredibly disruptive to the Neon API.

jrose-signal:
> I think it might need to return some type of borrow guard to ensure borrowed instance data lives long enough, unless `set_instance_data` is made to error if already set.
> 
> Also, I could see benefits of _only_ providing an immutable reference, otherwise, we would need to bind to the lifetime of `&mut Context` to prevent aliasing.

In this most limited form it's safe *because* it's using the lifetime of the `&mut Context`, but that does rule out a lot of patterns (such as accessing *two* instance data values without copying). Something more flexible would definitely require more design.

[...]

PR #902, opened 2022-05-24
feature(neon): API for instance-global data

dherman:
This PR adds a `neon::instance::Global` API for storing instance-global data:

```rust
static THREAD_ID: Global<u32> = Global::new();

THREAD_ID.set(&mut cx, x);

let thread_id: u32 = THREAD_ID.get(&mut cx).unwrap();
```


crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use once_cell::sync::OnceCell;
+
+use crate::context::Context;
+use crate::lifecycle::InstanceData;
+
+static COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+fn next_id() -> usize {
+    COUNTER.fetch_add(1, Ordering::SeqCst)
```
kjvalencik, line 13: `Ordering:: Relaxed` should be sufficient for unique IDs. This code doesn't rely on ordering for any fencing/protections. I.e., if concurrent bits of code attempt to `fetch_add` at the same time, the first one to start might be the 2nd to finish, but that doesn't matter. The only thing that matters is that the fetch and add are atomic.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn borrow<'a, 'b, C>(&self, cx: &'b mut C) -> Option<&'b T>
+    where
+        C: Context<'a>,
+    {
+        InstanceData::globals(cx)[self.id()]
+            .as_ref()
+            .map(|boxed| boxed.downcast_ref().unwrap())
+    }
+
+    pub fn borrow_mut<'a, 'b, C>(&self, cx: &'b mut C) -> Option<&'b mut T>
```
kjvalencik, line 46: I prefer `get` and `get_mut` instead of `borrow` since there is [precedence](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.get) for `get` returning an `Option`.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    fn id(&self) -> usize {
+        *self.id.get_or_init(next_id)
+    }
+}
+
+impl<T: Any + Send> Global<T> {
+    pub fn borrow<'a, 'b, C>(&self, cx: &'b mut C) -> Option<&'b T>
```
kjvalencik, line 37: Seeing this written out, I think there's value to add a `get_or_init`--even prematurely. Specifically, I'm thinking of the use case where a caller only needs `&T`, but doesn't know if it's initialized yet. This would require `&mut C`. Best case, this requires manually dropping the created item and then re-calling, worst case it requires `unsafe`.

Instead, we could handle that invariant for the caller. I'm okay leaving this for a follow-up PR.

kjvalencik, line 37: Nit, even though we use it in other places, I get confused which is `'a` and which is `'b`. What do you think of trying to call the inner lifetime `'cx` more frequently?

```suggestion
    pub fn borrow<'a, 'cx, C>(&self, cx: &'a mut C) -> Option<&'a T>
    where
        C: Context<'cx>,
```

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn set<'a, 'b, C>(&self, cx: &'b mut C, v: T)
+    where
+        C: Context<'a>,
+    {
+        InstanceData::globals(cx)[self.id()] = Some(Box::new(v));
+    }
+}
+
+impl<T: Any + Send + Clone> Global<T> {
+    pub fn get<'a, C>(&self, cx: &mut C) -> Option<T>
```
kjvalencik, line 64: A `get` that clones is a little strange. I think this should be removed since it could be replaced without much boilerplate:

```rust
// With `get`
let data = DATA.get(&cx);

// Without `get`
let data = DATA.borrow(&cx).cloned();
```

If we do keep this, I think it shoudl take `&C` and not `&mut C`.

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +58,36 @@ pub(crate) struct InstanceData {
 
     /// Shared `Channel` that is cloned to be returned by the `cx.channel()` method
     shared_channel: Channel,
+
+    /// Table of user-defined global cells.
+    globals: GlobalTable,
+}
+
+pub(crate) struct GlobalTable {
+    cells: Vec<Option<Box<dyn Any + Send>>>,
```
kjvalencik, line 67: Does this _need_ to be `Send`? I'm not actually clear if it does since theoretically it is only available on the same thread. If we don't want to think about it too hard, we could relax this bound in the future.

I do think it needs to be `'static`.

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +58,36 @@ pub(crate) struct InstanceData {
...
+}
+
+impl GlobalTable {
+    fn new() -> Self {
```
kjvalencik, line 71: Could be replaced with:

```rust
#[derive(Default)
pub(crate) struct GlobalTable {
    /* ... */
}

GlobalTable::default()
```

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +58,36 @@ pub(crate) struct InstanceData {
...
+        Self { cells: Vec::new() }
+    }
+}
+
+impl Index<usize> for GlobalTable {
+    type Output = Option<Box<dyn Any + Send>>;
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.cells[index]
```
kjvalencik, line 80: Since this doesn't resize, it will panic if it hasn't been used yet. I don't think `.borrow(..)` panicking, but `borrow_mut(..)` not panicking is great UX.

crates/neon/src/instance/mod.rs
dherman: Ooh, I didn't know about `Option::cloned()`, that does help. I ran into annoying paper cuts with `get` because it keeps the borrow of the context live, so you can easily end up with borrowck errors like this:

image

But that's pretty easily written like this:

```rust
pub fn get_thread_id(mut cx: FunctionContext) -> JsResult<JsValue> {
    let id = THREAD_ID.get(&mut cx).cloned();
    Ok(match id {
        Some(id) => cx.number(id).upcast(),
        None => cx.undefined().upcast(),
    })
}
```

I'll drop the method for now and maybe recommend `Option::cloned()` in the API docs for `get()` as a useful idiom.

crates/neon/src/instance/mod.rs
dherman: Good precedent, let's go with `get` and `get_mut`.

crates/neon/src/instance/mod.rs
dherman: I agree, it'll probably be a common case. I think I'll add it. I'm also considering adding `take()`—WDYT?

crates/neon/src/lifecycle.rs
dherman: Yeah. It's just a `pub(crate)` internal data structure and this is an unused method (required because `IndexMut` extends `Index`) so I didn't worry too much about this.

I liked being able to use the `[]` syntax because I found it more readable, but it's only used once or twice so maybe it just doesn't pay for itself and I should just use an inherent method.

crates/neon/src/instance/mod.rs
kjvalencik: How would that differ from `Option::take`? I.e., `.get_mut(&mut cx).take()`.

crates/neon/src/instance/mod.rs
kjvalencik: I think this might have been the reason we originally considered a `OnceCell` model that *doesn't* give mutability.

If we never vend a `&mut T`, we could give `&'cx T` instead of `&'a T`. I'm still not sure which is better; requiring a reference counting/cloning for *this* use case or requiring interior mutability for changing.

crates/neon/src/instance/mod.rs
kjvalencik: The difference is that we are giving out `Option<&mut T>` instead of `&mut Option<T>`. This is a valuable API.

dherman: I think we could minimize the size of our API and give people the full expressiveness of the `Option` API if we boxed the cells of the vec and gave an `&mut Option<T>` out instead. That comes at the cost of extra heap allocation but I think the intended use of globals is to be used sparingly anyway, so that might be a worthwhile tradeoff.

That said, we've been talking about trying an immutable version of this API instead anyway, which would make the question moot.

crates/neon/src/instance/mod.rs
dherman: TBH I don't really understand this because I don't understand atomics well enough. Given that this is not at all in a hot path, should we just go with `SeqCst` so maintainers don't need to think too hard about it?

crates/neon/src/lifecycle.rs
dherman: We discussed `Send` on a call and it's hard to predict. You could imagine a future Node runtime that runs workers on green threads and might execute the same worker on different OS threads at different times. But it's hard to know if even that is unsafe since it'd still be sequential (maybe if it depends on TLS?) and it might not even be a likely possibility. All the same, it just feels like we should do the safer thing for now and we could consider relaxing it in the future.

crates/neon/src/instance/mod.rs
dherman: I just pushed a commit that eliminates all the `&mut T` and uses `&'cx T` instead of `&'a T`. Let's take a look and see what we think!

crates/neon/src/instance/mod.rs
```diff
@@ -35,30 +35,33 @@ impl<T> Global<T> {
...
+            .downcast_ref()
+            .unwrap();
 
-    pub fn set<'cx, 'a, C>(&self, cx: &'a mut C, v: T)
-    where
-        C: Context<'cx>,
-    {
-        *InstanceData::globals(cx).get(self.id()) = Some(Box::new(v));
+        unsafe {
+            std::mem::transmute(r)
```
kjvalencik, line 64: Given the inherent danger of `transmute`, I have a slight preference to name the types directly even though they can be inferred. This will make it less likely that a future change will accidentally do something unexpected.

```suggestion
            std::mem::transmute::<&'a T, &'cx T>(r)
```

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+            .get(self.id())
+            .as_ref()
+            .map(|boxed| boxed.downcast_ref().unwrap());
+
+        unsafe {
+            std::mem::transmute::<Option<&'a T>, Option<&'cx T>>(r)
+        }
+    }
+
+    pub fn get_or_init<'cx, 'a, C, F>(&self, cx: &'a mut C, f: F) -> &'cx T
```
kjvalencik, line 52: `get_or_try_init` is also useful.

If we wanted to be lazy and directly expose the full expressive power of `OnceCell` we could make `Global::<T>::get` return [`&'cx once_cell::unsync::OnceCell<T>`](https://docs.rs/once_cell/latest/once_cell/unsync/struct.OnceCell.html) instead of `Option<&'cx T>`.

Instead of `Vec<Option<Box<Any>>>` we would store `Vec<Box<Any>>` where the `Any` are `OnceCell<T>` instead of `T`.

This would be flexible, but might be too prescriptive since a user could always store `OnceCell` themselves and just have an extra `.get_or_init(Default::default)` call in the chain.

```rust
static GLOBAL: Global<OnceCell<u32>> = Global::new();

let answer = GLOBAL.get_or_init(Default::default).get_or_init(|| 42);
```

dherman:
@jrose-signal We're playing around with two different versions of this API: an immutable one and a mutable one. (Either variation of the API should be _more or less_ equally expressive, but they each make certain use cases more or less ergonomic.) I'll describe the pros and cons below but since you filed #728 I'm curious if you have a sense of whether your use cases would lean more on the mutable or immutable side?

----

## Immutable

### Pros

- We can give out an `&'cx T` to the contents of the cell, which is live for the duration of the context (usually the duration of a Neon function)
- Encourages the use of stable types such as constants or extend-only data structures, which is generally safer for global data

### Cons

- Any mutability needs to be wrapped in interior mutability e.g. `RefCell`

## Mutable

### Pros

- More convenient for arbitrary mutable datatypes

### Cons

- We can't safely hand out a long-lived reference, i.e. `Global::get(&'a mut cx)` can only produce an `Option<&'a T>`
- Most methods end up locking the `&mut cx` context, requiring users to work around borrowck more (e.g. by cloning the contents)
- Requires either extra helper methods (e.g. `Global::take()`) for common cases, or an extra layer of internal boxing in order to vend pointers to the cell's `Option<T>` so users can take advantage of the full `std::option::Option` APIs

jrose-signal:
I'm afraid I've paged out the scheme I was going to use with this feature, especially since @indutny-signal successfully got Node to speed up "running this microtask has put another task on the microtask queue". I *think* I was just going to stick a unique ID in each instance and then have an API that did something like the following:
```rust
fn run_or_send(self, cx: &mut impl Context<'_>) {
  if UNIQUE_ID.get(cx) == self.target_context_unique_id {
    self.run(cx);
  } else {
    let channel = self.channel.clone();
    channel.send(move |cx| self.run(cx));
  }
}
```
This is not the most interesting use of instance-global data, but it is perfectly satisfied by the immutable API.

* * *

I haven't looked at your implementation, but it seems *possible* for you to provide *both* APIs, `fn get<'cx>(&self, cx: &Context<'cx>) -> &'cx Self::Target` and `fn get_mut<'a>(&self, &'a mut Context<'_>) -> &'a mut Self::Target`. But there is value in opinionated APIs as well, and I did completely gloss over the initialization part. So I guess it comes down to what someone would actually *do* with the mutable API where RefCell would be painful.

jrose-signal:
Oh yeah, I also had this use case:

> But in general anything that can be computed once and cached benefits from being in PerInstanceData, such as [recording a set of JavaScript-defined Error types to use for Rust errors](https://github.com/signalapp/libsignal-client/blob/master/rust/bridge/shared/src/node/error.rs#L14-L20).

This is a bit trickier, because the initialization logic wants to use the Context, and that could end up calling some other Neon function. I think that means I wouldn't be able to use `get_or_init`; I'd have to `get`, check for None, do some other stuff, and *then* `get_or_init`.

jrose-signal:
To be clear, even if `get_or_init` took the Context, you could still end up in this scenario:
```
outer_function
\ get_or_init
  \ some_js_function
    \ inner_function
      \ get_or_init
```
and you should not be able to init *twice*. So I *don't* think `get_or_init` should take the Context, but at the same time you can't *just* rely on `get_or_init` if you need the Context to do the init.

dherman:
I have seen this pattern a million times (an API using a closure to initialize a data structure, which makes the API re-entrant) and I still missed it! Thanks for the eagle eye, @jrose-signal.

I'm wondering if we can handle the double-initialization scenario automatically and panic. I think we can plausibly make the case that re-entrant re-initialization is a rare corner case and a bug, so it wouldn't need pollute the signature.

There's definitely precedent for having multiple variants to `get_or_init` such as `Option::get_or_insert`, `Option::get_or_insert_default`, `Options::get_or_insert_with`. The context version would pollute the API with `NeonResult`s but having variants with simpler signatures maybe makes it feel more ergonomic.

Some possible sketches:

```rust
impl<T> Global<T> {
    fn get_or_init<'cx, 'a, C>(&self, cx: &'a mut C, value: T) -> &'cx T
    where
        C: Context<'cx>,
    { ... }

    fn get_or_init_with<'cx, 'a, C, F>(&self, cx: &'a mut C) -> NeonResult<&'cx T>
    where
        C: Context<'cx>,
        F: FnOnce(&mut C) -> NeonResult<T>,
    { ... }
}

impl<T: Default> Global<T> {
    fn get_or_init_default<'cx, 'a, C>(&self, cx: &'a mut C) -> &'cx T
    where
        C: Context<'cx>,
    {
        self.get_or_init_with(cx, Default::default)
    }
}
```

Thoughts?

dherman:
I pushed an implementation of my ideas above. See what you think? @jrose-signal I think this might work for your use case without you having to do a complicated `get` + `get_or_init` dance, because it simply does the double-initialization checking for you.

kjvalencik:
I could be re-entrant, but I think that's okay as long as we define the semantics. It doesn't *need* to panic, it _could_ overwrite (since we know all other references of dropped). Overwrite semantics would be identical to if you hand wrote something like:
```
let data = if let Some(data) = DATA.get(&mut cx) {
    data
} else {
    DATA.insert(&mut cx, Data::new(cx.channel()));
}
```

dherman:
I think re-entrant initialization is going to be so rare that we should treat it as an error, but instead of waiting for the outer initialization to fail, we should fail on the inner initialization. I've pushed an implementation that does this checking by setting the state to a third "dirty" state during `get_or_try_init()` and then panics if any initialization occurs during that state.

crates/neon/src/instance/mod.rs
kjvalencik: Good point. I was thinking we would checking `COUNTER` on each access, but it's only when the `OnceCell` is uninitialized.

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +59,176 @@ pub(crate) struct InstanceData {
...
+        match self {
+            GlobalCell::Uninit => {
+                *self = f();
+            }
+            GlobalCell::Trying => panic!("attempt to reinitialize Global during initialization"),
+            GlobalCell::Init(_) => {}
+        }
+    }
+
+    pub(crate) fn get<'cx, 'a, C>(cx: &'a mut C, id: usize) -> Option<&mut GlobalCellValue>
```
kjvalencik, line 100: I haven't tested removing it, but it looks like this doesn't need to be `pub(crate)`.

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +59,176 @@ pub(crate) struct InstanceData {
...
+/// the state of a cell is properly managed through all possible control paths.
+/// As soon as the transaction begins, the cell is labelled as being in a dirty
+/// state (`GlobalCell::Trying`), so that any additional re-entrant attempts to
+/// initialize the cell will fail fast. The `Drop` implementation ensures that
+/// after the transaction, the cell goes back to a clean state of either
+/// `GlobalCell::Uninit` if it fails or `GlobalCell::Init` if it succeeds.
+struct TryInitTransaction<'cx, 'a, C: Context<'cx>> {
+    cx: &'a mut C,
+    id: usize,
+    _lifetime: PhantomData<&'cx ()>,
```
kjvalencik, line 187: I don't think this is needed because `'cx` is already a part of `C`.

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +59,176 @@ pub(crate) struct InstanceData {
...
+        tx
+    }
+
+    /// _Post-condition:_ If this method returns an `Ok` result, the cell is in the
+    /// `GlobalCell::Init` state.
+    fn run<F>(&mut self, f: F) -> NeonResult<()>
+    where
+        F: FnOnce(&mut C) -> NeonResult<GlobalCellValue>,
+    {
+        if self.is_trying() {
```
kjvalencik, line 207: Shouldn't this panic instead of succeeding?

crates/neon/src/lifecycle.rs
```diff
@@ -54,6 +59,176 @@ pub(crate) struct InstanceData {
...
+
+    fn is_trying(&mut self) -> bool {
+        match self.cell() {
+            GlobalCell::Trying => true,
+            _ => false,
+        }
+    }
+}
+
+impl<'cx, 'a, C: Context<'cx>> Drop for TryInitTransaction<'cx, 'a, C> {
```
kjvalencik, line 226: I love `TryInitTransaction` so much! This is a brilliant way to protect state! 😍

crates/neon/src/lifecycle.rs
```diff
@@ -83,7 +258,7 @@ impl InstanceData {
     /// # Safety
     /// No additional locking (e.g., `Mutex`) is necessary because holding a
     /// `Context` reference ensures serialized access.
-    pub(crate) fn get<'a, C: Context<'a>>(cx: &mut C) -> &'a mut InstanceData {
+    pub(crate) fn get<'cx, C: Context<'cx>>(cx: &mut C) -> &mut InstanceData {
```
kjvalencik, line 261: Thanks for fixing this soundness hole!

test/napi/lib/workers.js
```diff
@@ -106,3 +123,59 @@ describe("Worker / Root Tagging Tests", () => {
...
+    // 1. Global should start out uninitialized
+    assert.strictEqual(null, addon.get_reentrant_value());
+
+    // 2. Re-entrancy should panic
+    try {
+      let result = addon.reentrant_try_init(() => {
+        addon.reentrant_try_init(() => {});
+      });
+      assert.fail("should have panicked on re-entrancy");
+    } catch (expected) {}
```
kjvalencik, line 149: We should also assert that the inner closure was never executed. I haven't tested, but I think it current *is* being executed.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+        unsafe { std::mem::transmute::<&'a T, &'cx T>(r) }
+    }
+
+    /// Gets the current value of the global, initializing it with the result of
+    /// calling `f` if it has not yet been initialized. Returns `Err` if the
+    /// callback triggers a JavaScript exception.
+    ///
+    /// During the execution of `f`, calling any methods on this `Global` that
+    /// attempt to initialize it will panic.
+    pub fn get_or_try_init<'cx, 'a, C, F>(&self, cx: &'a mut C, f: F) -> NeonResult<&'cx T>
```
kjvalencik, line 98: Does this need to be constrained to a `NeonResult`? Since we are panicking instead of throwing an exception, I think this can be more generally `Result<&'cx T, E>`.

This would make it a bit more ergonomic for cases that are fallible for things that *aren't* Neon related.

```rust
let runtime = RUNTIME.get_or_try_init(&mut cx, |_| Runtime::new())?;
```

kjvalencik (changes requested):
I'm really excited about this! The API looks great and the transaction code is robust. ❤️

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl<T: Any + Send + 'static> Global<T> {
+    /// Gets the current value of the global. Returns `None` if the global has not
+    /// yet been initialized.
+    pub fn get<'cx, 'a, C>(&self, cx: &'a mut C) -> Option<&'cx T>
+    where
+        C: Context<'cx>,
+    {
+        // Safety: The type bound Global<T> and the fact that every Global has a unique
```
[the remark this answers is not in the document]
dherman: Yeah, exactly. Would it be better to write "Panic safety" or something like that?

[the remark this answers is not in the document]
dherman: Or "Unwrap safety"?

crates/neon/src/lifecycle.rs
dherman: It's used by `crate::instance::Global::get()`.

crates/neon/src/lifecycle.rs
dherman: I thought so too, but unfortunately rustc doesn't notice that and gives a type error.

crates/neon/src/instance/mod.rs
dherman: That's a really cool idea!

test/napi/lib/workers.js
dherman: Great point. I added the test logic. It's actually not being executed, because the inner `get_or_try_init` calls `tx.cell().pre_init()` during `TryInitTransaction::new()`, which panics before we ever get to calling `run()`.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl<T> Global<T> {
+    /// Creates a new global value. This method is `const`, so it can be assigned to
+    /// static variables.
+    pub const fn new() -> Self {
+        Self {
+            _type: PhantomData,
+            id: OnceCell::new(),
+        }
```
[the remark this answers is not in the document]
dherman: Oops, this failed to type check because `T` might not implement `Default` (annoying!). I reverted it.

crates/neon/src/lifecycle.rs
dherman: No, the `run` method is the actual body of the transaction. We panic in `new` if we were already `Trying`. Otherwise, `new()` _sets_ the state from `Uninit` to `Trying`. But if we are already initialized there's nothing to do. So this conditional is our way of checking to see if we have work to do or not.

crates/neon/src/instance/mod.rs
dherman: Implemented and pushed

crates/neon/src/instance/mod.rs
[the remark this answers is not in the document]
dherman: I went with "Unwrap safety" but I'm open to suggestions.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    fn id(&self) -> usize {
+        *self.id.get_or_init(next_id)
+    }
+}
+
+impl<T: Any + Send + 'static> Global<T> {
```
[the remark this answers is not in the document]
dherman: Nice idea. Let's save it for the future but keep this in mind.

dherman:
This test has me second-guessing the API name `Global`:

image

You can see a somewhat confusing ambiguity between "a JS global object" and "instance global data".

Some other possibilities:

### `neon::instance::Private` ("instance-private data")

```rust
use neon::instance::Private;

static GLOBAL_OBJECT: Private<Root<JsObject>> = Private::new();
```

### `neon::instance::Local` ("instance-local data")

```rust
use neon::instance::Local;

static GLOBAL_OBJECT: Local<Root<JsObject>> = Local::new();
```

### `neon::instance::Cell` ("instance cell")

```rust
use neon::instance::Cell;

static GLOBAL_OBJECT: Cell<Root<JsObject>> = Cell::new();
```

### `neon::instance::InstanceCell` ("instance cell")

```rust
use neon::instance::InstanceCell;

static GLOBAL_OBJECT: InstanceCell<Root<JsObject>> = InstanceCell::new();
```

crates/neon/src/lifecycle.rs
kjvalencik: Ah! Thanks for explaining.

kjvalencik:
I agree that `Global` could be a little confusing. I think `Private` could also be confusing if it's declared `pub`.

I like `neon::instance::Local` or `std::instance::LocalKey` since it aligns with the naming of [`std::thread::LocalKey`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html),

crates/neon/src/instance/mod.rs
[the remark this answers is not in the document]
kjvalencik: Oh, yea. Also, `Default::default` isn't a `const fn`. It makes more sense to go the other way and implement `Default` with `Global::new`.

```rust
pub struct Global<T> {
    _type: PhantomData<T>,
    id: OnceCell<usize>,
}

impl<T> Default for Global<T> {
    fn default() -> Self {
        Self::new()
    }
}

impl<T> Global<T> {
    pub const fn new() -> Self {
        Self {
            _type: PhantomData,
            id: OnceCell::new(),
        }
    }
}
```

The derived `Default` will require `T: Default`, but if you manually implement, that bound isn't necessary.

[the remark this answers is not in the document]
kjvalencik: Oh, you're right. This would require removing `'cx` and replacing it with a `for<'cx>` *everywhere*. While that works, it's much simpler to have a `PhantomData`.

kjvalencik (approved):

dherman:
I’m open to `Local`. I’ll try a version of the PR with that and see how it feels.

I think my instinct behind `Global` was that it didn’t ask the user to think too hard about “local to what scope?” since the “instance” concept is kind of nonobvious since it’s connected to (a) JS realms, which are not widely understood, and (b) the lifecycle of a Node addon, which is also not widely understood.

So maybe the way to counteract that is for the module docs for `neon::instance` to explain in approachable terms what instance scope means, and how it ties to JS realms, and how a Neon module is instantiated once per realm/worker/thread.

crates/neon/src/instance/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        // Safety: Since the Box is immutable and heap-allocated, it's guaranteed not to
+        // move or change for the duration of the context.
+        unsafe { std::mem::transmute::<&'a T, &'cx T>(r) }
+    }
+
+    /// Gets the current value of the cell, initializing it with the result of
+    /// calling `f` if it has not yet been initialized. Returns `Err` if the
+    /// callback triggers a JavaScript exception.
+    ///
+    /// During the execution of `f`, calling any methods on this `Local` that
```
kjvalencik, line 123: It would be nice to add this in a `# Panics` section to make it more noticable.

```
    /// # Panics
    ///
    /// During the execution of `f`, calling any methods on this `Local` that
    /// attempt to initialize it will panic.
```

[Example](https://docs.rs/tokio/1.18.2/tokio/sync/oneshot/struct.Receiver.html#panics) of how `tokio` does it.

jrose-signal:
I'll bikeshed that I'd prefer "LocalKey" to "Local". Especially once someone does `use neon::instance::Local;`, there's no indication at the use site what it's local *to,* at which point it's weird that it's being declared as a global. At least with LocalKey you know it's local relative to something.

kjvalencik:
@dherman it looks like you came to the same conclusions as `once_cell`:
1. Preventing reentrancy in the API is too limiting
2. Reentrancy is incorrect usage and should panic

https://docs.rs/once_cell/1.12.0/once_cell/unsync/struct.OnceCell.html#panics

dherman:
> I'll bikeshed that I'd prefer "LocalKey" to "Local". Especially once someone does `use neon::instance::Local;`, there's no indication at the use site what it's local _to,_ at which point it's weird that it's being declared as a global. At least with LocalKey you know it's local relative to something.

I think I agree that a global called local is weird, and that not knowing what it's local to is weird. I don't love "key" because from the user's perspective it really behaves more like a storage cell and not a key into another table (the table is a private implementation detail).

I wonder if we should play with the notion that this is instance-private data (per-instance data can simply be stashed on the module object itself). What do you think of `neon::instance::Private` or `neon::instance::PrivateCell`?

jrose-signal:
Hm…it behaves like a storage cell when you have exactly one instance, which I expect is *by far* the most common case for Neon users today. But if you could prove statically that you had exactly one instance, you wouldn't use this; you'd use a global. As soon as you have multiple instances, it no longer behaves like a cell, because two temporally-subsequent calls `foo.get(&mut cx)` might return different results even without intervening calls to `foo.set(&mut cx, bar)` or whatever.

dherman:
That’s a really great point.

I’ve always found `LocalKey` a confusing name. (Key to what? What data structure do I used it to index info?) But maybe the reality is that this kind of implicitly fanning-out data structure doesn’t really have an obvious analog, and I should try to get comfortable with the idiomatic Rust vocabulary.

kjvalencik:
@dherman I agree that it's not the perfect name, but my preference is to err towards aligning with `std`. In both cases the name and module path align:
```
// Key to data that is thread local
use std::thread::LocalKey;

// Key to data that is instance local
use neon::instance::LocalKey;
```

dherman:
Yeah I'm on board, and in fact, this made me realize that even instance is not a particularly helpful term, and we can fully align with std by calling this `neon::thread::LocalKey` and explaining this as "thread-local storage for JavaScript threads." Since JavaScript now has threads, and Node refers to them as such (e.g. `require('node:worker_threads')`), this fully aligns this API with the intuitions of TLS.

jrose-signal:
I was worried that someone would expect data set in one addon to be available in another one (per your diagram in https://github.com/neon-bindings/neon/issues/728#issuecomment-858972587), but since LocalKeys aren't created with any sort of user-visible representation, they'd never be the "same" key across addons anyway, any more than two copies of the same global would be the same variable at runtime in two dynamic libraries. (C header semantics notwithstanding…) So that lets Neon ignore the distinction between "worker/thread-local" and "addon/instance-local".

dherman:
@jrose-signal Just to make sure I follow, does that mean you do like this idea of calling it thread-local? (I pushed the change quickly just to see what it looks like, but I'm totally open to feedback.)

jrose-signal:
I think it's not bad. I don't love it because JS might some day use "thread" to mean something else; calling it "worker-local" would probably be the equivalent using today's terminology. "instance-local" isn't wrong but pushes people to learn about instances when they may not need to, so I understand why you're trying to move away from it.

dherman:
That makes sense. I think the reason I favor thread over worker is that (a) the main thread isn't a worker, and (b) Node refers to workers as threads already anyway.

dherman:
> [MDN refers to "threads" too](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) so yeah, should be clear enough. (And of course there can be a "technically it's per-instance" section in the docs.)

I'll add something to the docs…

dherman:
I added some text to the docs based on these suggestions. Reviews/feedback welcome!

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
+//! Thread-local storage for JavaScript threads.
+//!
+//! At runtime, an instance of a Node.js addon can contain its own local storage,
+//! which can then be shared and accessed as needed between Rust modules. This can
```
kjvalencik, line 4: I don't understand this bit, "as needed between Rust modules".

```suggestion
//! which can then be shared and accessed as needed from Rust in a Neon module. This can
```

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+//!
+//! pub fn thread_id<'cx, C: Context<'cx>>(cx: &mut C) -> NeonResult<u32> {
+//!     THREAD_ID.get_or_try_init(cx, |cx| {
+//!         let global = cx.global();
+//!         let require: Handle<JsFunction> = global.get(cx, "require")?;
+//!         let worker: Handle<JsObject> = require.call_with(cx)
+//!             .arg(cx.string("node:worker_threads"))
+//!             .apply(cx)?;
+//!         let threadId: Handle<JsNumber> = worker.get(cx, "threadId")?;
+//!         Ok(threadId.value(cx) as u32)
```
kjvalencik, line 24: Nit, not idiomatic Rust naming.

```suggestion
//!         let thread_id: Handle<JsNumber> = worker.get(cx, "threadId")?;
//!         Ok(thread_id.value(cx) as u32)
```

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+//!
+//! For some use cases, a single shared global constant stored in a `static` variable
+//! might be sufficient:
+//!
+//! ```
+//! static MY_CONSTANT: &'static str = "hello Neon";
+//! ```
+//!
+//! This variable will be allocated when the addon is first loaded into the Node.js
+//! process. This works fine for single-threaded applications, or global immutable
```
kjvalencik, line 39: I don't think mutability is important. Maybe say, "or data that can be safely shared across instances of a module."

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+//! data.
+//!
+//! However, since the addition of [worker threads][workers] in Node v10,
+//! modules can be instantiated multiple times in a single Node process. This means
+//! that while the dynamically-loaded binary library (i.e., the Rust implementation of
+//! the addon) is only loaded once in the running process, but its `main()` function
```
kjvalencik, line 45:
```suggestion
//! the addon) is only loaded once in the running process, but its [`#[main]`](neon::main) function
```

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+//!
+//! ![The Node.js addon lifecycle, described in detail below.][lifecycle]
+//!
+//! This means that any thread-local data needs to be initialized separately for each
+//! instance of the addon. This module provides a simple container type, [`LocalKey`](LocalKey),
+//! for allocating and initializing thread-local data. (Technically, this data is stored in the
+//! addon's [module instance][environment], which is equivalent to being thread-local.)
+//!
+//! A common example is when an addon defines a custom datatype, which cannot be used across
+//! separate threads. The constructor for the datatype can be [rooted](crate::handle::Root) and
```
kjvalencik, line 56: This might be a bit confusing. It _seems_ to imply it's the *custom* part that makes it unable to be shared. In the example, this can't be shared because it's a reference to a JavaScript value and this is generally true about all `JsValue`.

```suggestion
//! A common example is when an addon needs to maintain a reference to a JavaScript value. References cannot
//! be used across separate threads. The constructor for the datatype can be [rooted](crate::handle::Root) and
```

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+//! pub fn my_constructor<'cx, C: Context<'cx>>(cx: &mut C) -> JsResult<'cx, JsFunction> {
+//!     let constructor = MY_CONSTRUCTOR.get_or_try_init(cx, |cx| {
+//!         let constructor: Handle<JsFunction> = initialize_my_datatype(cx)?;
+//!         Ok(constructor.root(cx))
+//!     })?;
+//!     Ok(constructor.to_inner(cx))
+//! }
+//! ```
+//!
+//! Notice that if this code were implemented without a `LocalKey`, it would fail whenever
```
kjvalencik, line 74:
```suggestion
//! Notice that if this code were implemented without a `LocalKey`, it would panic whenever
```

As an aside, it looks like I didn't update the `Root` documentation when this behavior was added.

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+            LocalCell::get(cx, self.id()).map(|value| value.downcast_ref().unwrap());
+
+        // Safety: Since the Box is immutable and heap-allocated, it's guaranteed not to
+        // move or change for the duration of the context.
+        unsafe { std::mem::transmute::<Option<&'a T>, Option<&'cx T>>(r) }
+    }
+
+    /// Gets the current value of the cell, initializing it with `value` if it has
+    /// not yet been initialized.
+    pub fn get_or_init<'cx, 'a, C>(&self, cx: &'a mut C, value: T) -> &'cx T
```
kjvalencik, line 167: Bike shedding a bit, but I would remove this method since it saves at most 8 characters. This frees up the name to rename `get_or_init_with` the shorter `get_or_init`.

```diff
- let global = GLOBAL.get_or_init(&mut cx, value);
+ let global = GLOBAL.get_or_init(&mut cx, move || value);
```

crates/neon/src/thread/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        // Safety: Since the Box is immutable and heap-allocated, it's guaranteed not to
+        // move or change for the duration of the context.
+        unsafe { std::mem::transmute::<&'a T, &'cx T>(r) }
+    }
+
+    /// Gets the current value of the cell, initializing it with the result of
+    /// calling `f` if it has not yet been initialized. Returns `Err` if the
+    /// callback triggers a JavaScript exception.
+    ///
+    /// During the execution of `f`, calling any methods on this `LocalKey` that
```
kjvalencik, line 204: Can we put this under a `# Panics` headline?

```suggestion
    /// # Panics
    ///
    /// During the execution of `f`, calling any methods on this `LocalKey` that
```

crates/neon/src/thread/mod.rs
dherman: Good idea, and this makes a better symmetry in the `get_or_init` / `get_or_try_init` naming. `once_cell` does the same thing.

crates/neon/src/thread/mod.rs
dherman: Yes, good call and sorry I missed this change when you suggested it earlier.

crates/neon/src/thread/mod.rs
dherman: I went with "thread-safe"

crates/neon/src/thread/mod.rs
dherman: I went ahead and added panic notes to `Root::{into_inner, to_inner}`

kjvalencik (approved):

kjvalencik:
Looks great! Love the docs. :shipit:

dherman:
My sincere thanks to both @jrose-signal and @kjvalencik for all the excellent feedback, ideas, and reviews. I'm happy with how this API came together.

PR #904, opened 2022-06-08
neon: Remove JsResultExt

kjvalencik:
`JsResultExt` is redundant with `ResultExt`. They are identical traits except for an additional `V: Value` bound in `JsResultExt`.

Having two traits presents the opportunity for ambiguous method resolution. This PR proposes that we take the opportunity before 1.0 to make a breaking change and remove `JsResultExt`.

As demonstrated in the diff, replacing a `JsResultExt` implementation with `ResultExt` is as trivial as replacing `JsResultExt<..>` with `ResultExt<Handle<..>>`.


dherman (approved):
I agree, there's no real point to `JsResultExt` anymore.

PR #907, opened 2022-06-10
safety(neon): Improve safety of JsBox on Node-API >= 8

kjvalencik:
Generate a unique type id for the crate and tag externals. This is done at runtime instead of compile time to help with reproducible builds.

This requires that the module be built for Node-API 8+. In the future, we may want to do optional symbol loading for internal code that makes runtime decisions. That way we could provide this additional safety guarantee on newer versions of Node even if it was compiled for an older version.

### DIFF
--- a/crates/neon/src/lib.rs
+++ b/crates/neon/src/lib.rs
@@ -98,3 +98,26 @@ pub use neon_macros::*;
 
 #[cfg(feature = "napi-6")]
 mod lifecycle;
+
+#[cfg(feature = "napi-8")]
+static MODULE_TAG: once_cell::sync::Lazy<crate::sys::TypeTag> = once_cell::sync::Lazy::new(|| {
+    let mut lower = [0; std::mem::size_of::<u64>()];
+
+    // Generating a random module tag at runtime allows Neon builds to be reproducible. A few
+    //  alternativeswere considered:
+    // * Generating a random value at build time; this reduces runtime dependencies but, breaks
+    //   reproducible builds
+    // * A static random value; this solves the previous issues, but does not protect against ABI
+    //   differences across Neon and Rust versions
+    // * Calculating a variable from the environment (e.g. Rust version); this theoretically works
+    //   but, is complicated and error prone. This could be a future optimization.
+    getrandom::getrandom(&mut lower).expect("Failed to generate a Neon module type tag");
+
+    // We only use 64-bits of the available 128-bits. The rest is reserved for future versioning and
+    // expansion of implementation.
+    let lower = u64::from_ne_bytes(lower);
+
+    // Note: `upper` must be non-zero or `napi_check_object_type_tag` will always return false
+    // https://github.com/nodejs/node/blob/5fad0b93667ffc6e4def52996b9529ac99b26319/src/js_native_api_v8.cc#L2455
+    crate::sys::TypeTag { lower, upper: 1 }
+});
--- a/crates/neon/Cargo.toml
+++ b/crates/neon/Cargo.toml
@@ -20,6 +20,7 @@ anyhow = "1.0.58" # used for a doc example
 nodejs-sys = "0.13.0"
 
 [dependencies]
+getrandom = { version = "0.2.7", optional = true }
 libloading = "0.7.3"
 semver = "1"
 smallvec = "1.4.2"
@@ -55,7 +56,7 @@ napi-4 = ["napi-3"]
 napi-5 = ["napi-4"]
 napi-6 = ["napi-5"]
 napi-7 = ["napi-6"]
-napi-8 = ["napi-7"]
+napi-8 = ["napi-7", "getrandom"]
 napi-latest = ["napi-8"]
 napi-experimental = ["napi-8"]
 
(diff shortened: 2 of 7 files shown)


crates/neon/src/lib.rs
```diff
@@ -98,3 +98,10 @@ pub use neon_macros::*;
 
 #[cfg(feature = "napi-6")]
 mod lifecycle;
+
+#[cfg(feature = "napi-8")]
+static MODULE_TAG: once_cell::sync::Lazy<crate::sys::TypeTag> = once_cell::sync::Lazy::new(|| {
+    let (lower, upper) = rand::random::<(u64, u64)>();
```
kjvalencik, line 104: Is there another way to get a unique identifier for a shared library? We don't really need this to be dynamic, we just need it to be unique to this shared library.

It could be generated at build time, but a build script seemed a little heavy.

crates/neon/build.rs
```diff
@@ -0,0 +1,13 @@
+use std::{env, fs, path::PathBuf};
+
+fn main() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out_file = out_path.join("type_tag.rs");
+    let (lower, upper) = rand::random::<(u64, u64)>();
```
kjvalencik, line 6: We should only generate 64-bits of randomness and use `0` for `upper`. This gives us some reserved space to change the scheme at a later date.

crates/neon/src/lib.rs
```diff
@@ -98,3 +98,6 @@ pub use neon_macros::*;
 
 #[cfg(feature = "napi-6")]
 mod lifecycle;
+
+#[cfg(feature = "napi-8")]
+static MODULE_TAG: crate::sys::TypeTag = include!(concat!(env!("OUT_DIR"), "/type_tag.rs"));
```
kjvalencik, line 103: We only need a single tag because Neon only stores a single type `Box<dyn Any>`. However, as a future optimization, without any breaking changes, we could use [`TypeId`](https://doc.rust-lang.org/stable/std/any/struct.TypeId.html) as part of the definition and avoid an extra layer of indirection--implementing our *own* `Any`. However this would rely on the internal representation of `TypeId` which is not part of Rust's stable API.

crates/neon/src/lib.rs
complexspaces: Wouldn't a random, static identifier interfere with reproducible builds?

kjvalencik: That's a good reason to generate at runtime instead of build time.

Do you know if there's some cargo variable that could be used here? What we're actually asserting is the ABI stability of std::any::Any.

complexspaces: I believe that it could still be generated at buildtime, FWIW. What if it was implemented like this?

```rust
use std::{path::PathBuf, process::Command};

fn env_var(key: &'static str) -> String {
    std::env::var(key).unwrap()
}

fn main() {
    let out_path = PathBuf::from(env_var("OUT_DIR"));
    let out_file = out_path.join("type_tag.rs");

    let rustc_version = Command::new(env_var("RUSTC")).arg("--version").map(|out| String::from_utf8(out.stdout).unwrap()).expect("failed to run rustc");
    
    let type_tag = format!(
        "crate::sys::TypeTag {{ rustc: {} }}",
        rustc_version
    );

    fs::write(out_file, type_tag).unwrap();

}
```

This should be enough to guarantee that an `Any` definition will be stable across all crates. I don't believe the toolchain is needed because Node modules can only be loaded on a single architecture, so the OS loader prevents any issues there 🤔. If I'm wrong, the contents of `TARGET` could be included as well.

kjvalencik: There are multiple toolchains that can target the same system. For example, msvc and mingw.

This is a pretty good idea, but it would need to be hashed. Node-API type tags are a u128.

complexspaces: What if we hashed the rustc version into one `u64` and then the toolchain/target information into another using [DefaultHasher](https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html)?

kjvalencik: Ideally we wouldn't use the full 128-bits. I would like to reserve at least a byte for some static version identifier.

Hashing Rust metadata feels a little brittle to me. Do you happen to know if there's something that can be used for a unique identifier of a shared library that's been open with `dlopen`? Memory address or something?

crates/neon/src/lib.rs
```diff
@@ -98,3 +98,16 @@ pub use neon_macros::*;
 
 #[cfg(feature = "napi-6")]
 mod lifecycle;
+
+#[cfg(feature = "napi-8")]
+static MODULE_TAG: once_cell::sync::Lazy<crate::sys::TypeTag> = once_cell::sync::Lazy::new(|| {
+    let mut lower = [0; std::mem::size_of::<u64>()];
```
dherman, line 104: Maybe add a comment to say we aren't using this 64 bits so for now it's just reserved for future use.

dherman (approved):
Looks good, and I agree with the design decision you made.

It would be great to capture the rationale you explained to me somewhere (a short RFC?). And we should add to the `JsBox` docs information about (lack of) interoperability between module instances. But that could be done as followup.

crates/neon/src/lib.rs
kjvalencik: I'll add this. FYI, we're using 64-bits and not the full 128-bits.

PR #909, opened 2022-06-17
Typed Array discoverability

dherman:
This PR adds API docs for typed arrays, as well as several conveniences in the API:

- `JsTypedArray::new()` and `JsTypedArray::from_buffer` for constructing typed array objects
- Type aliases for typed array types (e.g. `JsUint32Array`)

Collectively, these changes should **make the Neon typed arrays API more discoverable** and easier to get started using.

RFC: neon-bindings/rfcs#47


crates/neon/src/lib.rs
```diff
@@ -89,7 +89,13 @@ pub mod result;
 mod sys;
 #[cfg(feature = "napi-6")]
 pub mod thread;
-pub mod types;
+// To use the #[aquamarine] attribute on the top-level neon::types module docs, we have to
+// use this hack so we can keep the module docs in a separate file.
+// See: https://github.com/mersinvald/aquamarine/issues/5#issuecomment-1168816499
+mod types_docs;
```
dherman, line 95: I ended up putting this crate at the top level as a peer of `types_impl` instead of nesting it inside of `types_impl`, because otherwise the glob export ends up polluting `neon::types` with a useless `neon::types::exports` module.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,760 @@
...
+///     let mut array: Handle<JsUint8Array> = cx.argument(0)?;
+///
+///     for elem in array.as_mut_slice(&mut cx).iter_mut() {
+///         *elem *= 2;
+///     }
+///
+///     Ok(cx.undefined())
+/// }
+/// ```
+pub type JsUint8Array = JsTypedArray<u8>;
```
kjvalencik, line 600: It would be nice to generate the aliases and examples in the `impl_typed_array` macro.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,760 @@
+use std::{marker::PhantomData, slice};
```
dherman, line 1: A better diff for this file, via command line:

```
git diff f747e670d63b30d434e3b303b76056f7fe6759cb:crates/neon/src/types/buffer/types.rs HEAD:crates/neon/src/types_impl/buffer/types.rs
```

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,760 @@
...
+            // The borrowed data must be guarded by `Ledger` before returning
+            Ledger::try_borrow_mut(
+                &lock.ledger,
+                slice::from_raw_parts_mut(info.data.cast(), info.length),
+            )
+        }
+    }
+}
+
+impl<T: Binary> JsTypedArray<T> {
```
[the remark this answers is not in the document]
dherman: Yeah, it's an interesting idea but I think I agree the duplication would on balance be a worse docs experience.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,760 @@
...
+        Ledger::try_borrow_mut(&lock.ledger, unsafe {
+            sys::arraybuffer::as_mut_slice(lock.cx.env().to_raw(), self.to_raw())
+        })
+    }
+}
+
+/// A marker trait for all possible element types of binary buffers.
+///
+/// This trait can only be implemented within the Neon library.
+pub trait Binary: private::Sealed + Copy {
```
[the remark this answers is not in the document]
dherman: I just pushed a change to make it pub-in-private. I think I like the docs better this way. It does still mention the `T: Binary` constraint on `JsTypedArray<T>` but with the big table in the `JsTypedArray` docs showing the full set of supported type arguments, I think it's fine.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,760 @@
...
+    /// The internal Node-API enum value for this binary type.
+    const RAW: TypedArrayType;
```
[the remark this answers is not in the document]
dherman: I went with `TYPE_TAG`.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,758 @@
...
+            Ok(Handle::new_internal(Self {
+                local: arr,
+                _type: PhantomData,
+            }))
+        } else {
+            Err(Throw::new())
+        }
+    }
+
+    pub fn new<'cx, 'a, C>(cx: &'a mut C, len: usize) -> JsResult<'cx, Self>
```
kjvalencik, line 502: Nit, `'a` can be elided.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,758 @@
...
+            Ledger::try_borrow_mut(
+                &lock.ledger,
+                slice::from_raw_parts_mut(info.data.cast(), info.length),
+            )
+        }
+    }
+}
+
+impl<T: private::Binary> JsTypedArray<T> {
+    pub fn from_array_buffer<'cx, 'a, C>(
```
kjvalencik, line 473: Nit, `'a` can be elided.

kjvalencik: It might be nice to have one more helper that assumes the entire buffer should be in the typed array. It might be good for that version to have the name `from_array_buffer` and maybe use some other name for this, e.g., `from_array_buffer_slice`.

```rust
impl<T: private::Binary> JsTypedArray<T> {
    pub fn from_array_buffer<'cx, C>(cx: &mut C, buffer: Handle<JsArrayBuffer>) -> JsResult<'cx, Self>
    where
        C: Context<'cx>,
    {
        let len = buffer.as_slice(cx).len();

        Self::from_array_buffer_slice(cx, buffer, 0, len)
    }
    
    /* ... */
}
```

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,758 @@
...
+/// [BigInt64Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array
+/// [Float32Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array
+/// [Float64Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float64Array
+/// [typed-arrays]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays
+/// [ArrayBuffer]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer
+/// [Buffer]: https://nodejs.org/api/buffer.html
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct JsTypedArray<T: private::Binary> {
+    local: raw::Local,
```
kjvalencik, line 375: I noticed we're missing one more typed array feature. Neon doesn't expose a way to get the backing `ArrayBuffer`.

https://nodejs.org/api/n-api.html#napi_get_typedarray_info

crates/neon/src/types_impl/buffer/types.rs
dherman: Would it make sense to use Rust's range syntax for this, and have it be a single method?

```rust
JsUint32Array::from_array_buffer(cx, buffer, ..) // full buffer
JsUint32Array::from_array_buffer(cx, buffer, 0 .. len) // buffer from half-open range [0, len)
JsUint32Array::from_array_buffer(cx, buffer, i ..) // buffer from index i to end
```

dherman: Also I wonder if the `array_` is just unnecessary in the name:

```rust
JsUint32Array::from_buffer(cx, buffer, ..)       // full buffer
JsUint32Array::from_buffer(cx, buffer, 0 .. len) // buffer from half-open range [0, len)
JsUint32Array::from_buffer(cx, buffer, i ..)     // buffer from index i to end
```

On the one hand, Node does have a legacy `Buffer` type, but:
- it's used less often than it used to be since JS typed arrays are standard and more fully-featured
- the standard JS API already has the `.buffer` property which returns an `ArrayBuffer`
- `.from_buffer` reads nicer IMO

kjvalencik: I like dropping `array`. I'm not sure about range syntax. It looks awkward, but it does make expressing the two optional arguments nicely.

Something to keep in mind is that a range makes it a little complicated because there are a bunch of different range types (Range, FullRange, etc.).

It would be nice to put the range on the buffer as an index, but since that returns a reference, I'm not sure how that would work.

kjvalencik: Actually, looks like there's a trait that covers all of them.

https://doc.rust-lang.org/std/ops/trait.RangeBounds.html

dherman: Right, I was thinking maybe we make `from_buffer` take a generic `impl RangeBounds` and `.buffer()` could return `(Handle<ArrayBuffer>, Range)`.

dherman: Alternatively we could have `from_buffer()` and `buffer()` just take and return the ArrayBuffer, and do what I said above but call it `from_buffer_range()` and `buffer_range()`.

dherman: There’s some precedent in std like `String::replace_range()`: https://doc.rust-lang.org/alloc/string/struct.String.html#method.replace_range

dherman: Plot twist! Both the standard JS APIs and Node-API use (_byte offset_ x _number of typed elements_), so a single range has an impedance mismatch. Hmmmm…

dherman: @kjvalencik I think I'm coming back around to your original suggestion. I'm not sure about the word "slice" since it sort of suggests the input should be a Rust slice datatype. I thought of "range" (which led me to this exploration of range syntax), but that sort of suggests the input should be a Rust range datatype.

`from_buffer_region`?

Or do you think the benefit of the "slice" analogy outweighs the inconsistency/confusion from not actually taking a slice?

dherman: I went with `from_buffer_region`. I think it reads pretty nicely.

crates/neon/src/types_impl/buffer/types.rs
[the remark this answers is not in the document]
dherman: I had to revert this change because I realized that you can't write helper functions that operate generically on any `JsTypedArray<T>` unless you can refer to the `Binary` trait to add the proper constraint.

test/napi/src/js/objects.rs
```diff
@@ -167,6 +167,105 @@ pub fn return_external_array_buffer(mut cx: FunctionContext) -> JsResult<JsArray
...
+
+    obj.set(cx, "byteOffset", byte_offset)?;
+    obj.set(cx, "length", len)?;
+    obj.set(cx, "byteLength", byte_length)?;
+    obj.set(cx, "buffer", buffer)?;
+
+    Ok(obj)
+}
+
+pub fn get_typed_array_info(mut cx: FunctionContext) -> JsResult<JsObject> {
```
dherman, line 241: This implementation is unfortunate, but it's I think a pretty hard-to-avoid consequence of the dynamic/static divide between JavaScript and Rust.

One option would be to have some kind of a more dynamic super type of `JsTypedArray<T>` like `JsAnyTypedArray` with some kind of dynamic API for introspecting on the element type. But if we created an enum for the element type, it wouldn't be very future-proof in case JavaScript adds new array types (like 128 bit integers or 16 bit floats, for example). That would add new variants to the enum, which wouldn't be backwards-compatible.

Maybe it could be made to work with some kind of even more clever use of traits or something, not sure.

But this is a pretty peculiar function needed for the sake of testing Neon, and my gut tells me it'll be fairly rare that people need to write (a) generic code that operates on all `JsTypedArray<T>` and (b) expose that code directly to JavaScript as a JS API that provides generic typed array functionality.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,721 @@
...
+    }
+
+    /// Returns the offset (in bytes) of the typed array from the start of its
+    /// [`JsArrayBuffer`](JsArrayBuffer).
+    pub fn byte_offset<'cx, C>(&self, cx: &mut C) -> usize
+    where
+        C: Context<'cx>,
+    {
+        let env = cx.env();
+        let info = unsafe { sys::typedarray::info(env.to_raw(), self.to_raw()) };
```
kjvalencik, line 590: It's not very efficient to repeatedly call `typedarray::info` for this data. Can we memoize it on an `Option` field in `JsTypedArray`?

It does make the struct a little larger to copy, but I think that's okay since it's still not *that* large.

dherman: I like this idea but didn't we run into trouble before with making some handles into fat pointers?

kjvalencik: No, we have a couple. Most notably is `JsBox`.

dherman: OK, I pushed an implementation of this optimization! See what you think.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,765 @@
...
+        T: AsMut<[u8]> + Send + 'static,
+    {
+        let env = cx.env().to_raw();
+        let value = unsafe { sys::arraybuffer::new_external(env, data) };
+
+        Handle::new_internal(Self(value))
+    }
+}
+
+impl<'cx> Handle<'cx, JsArrayBuffer> {
```
kjvalencik, line 213: I suggest a few changes to this:
* Renaming to `to_region` for consistency
* Taking `&self` instead of `self`
* Implementing on `JsArrayBuffer` instead of `Handle` for better docs

```rust
impl JsArrayBuffer {
    pub fn to_region<T: Binary>(&self, byte_offset: usize, len: usize) -> Region<T> {
        Region {
            buffer: Handle::new_internal(JsArrayBuffer(self.0)),
            byte_offset,
            len,
            phantom: PhantomData,
        }
    }
}
```

The downside of this approach is that the lifetime is now tied to the borrow instead `'cx`. This means that a `Region` can no longer escape a helper function.

The better documentation *might* be worth the limitation, but we could also fix this up if [`arbitrary_self_types`](https://github.com/rust-lang/rust/issues/44874) stabilizes.

dherman: Update: We discussed all these in the RFC, and I've updated this PR to match the RFC.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,783 @@
...
+            len: info.length,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Constructs a new typed array of length `len`.
+    ///
+    /// The resulting typed array has a newly allocated storage buffer of
+    /// size `(len * size_of::<T>())` bytes.
+    pub fn new<'cx, C>(cx: &mut C, len: usize) -> JsResult<'cx, Self>
```
kjvalencik, line 624: I had one additional idea for a helper. `JsTypedArray::from_slice`.

```rust
impl<T: Binary> JsTypedArray<T> {
    pub fn from_slice<'cx, C>(cx: &mut C, data: &[T]) -> JsResult<'cx, Self>
    where
        C: Context<'cx>,
    {
        let buffer = cx.array_buffer(data.len() * std::mem::size_of::<T>())?;
        let mut array = Self::from_region(cx, buffer.region(0, data.len()))?;

        array.as_mut_slice(cx).copy_from_slice(data);

        Ok(array)
    }
}
```

```rust
let nums = [1, 2, 3, 4, 5];
let u32array = JsTypedArray::from_slice(&mut cx, nums);
```

kjvalencik (changes requested):

crates/neon/src/types_impl/buffer/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+/// //              +-------+-------+
+/// //         arr: |       |       |
+/// //              +-------+-------+
+/// //               0       1       2
+/// let buf = cx.array_buffer(16)?;
+/// let arr = JsUint32Array::from_region(&mut cx, buf.region(4, 2))?;
+/// # Ok(arr)
+/// # }
+/// ```
+#[derive(Clone, Copy)]
```
kjvalencik, line 222: I'm not sure `Region` should be `Copy`. It's a little large, but on the line.

crates/neon/src/types_impl/buffer/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+pub struct Region<'cx, T: Binary> {
+    pub(super) buffer: Handle<'cx, JsArrayBuffer>,
+    pub(super) offset: usize,
+    pub(super) len: usize,
+    pub(super) phantom: PhantomData<T>,
+}
+
+impl<'cx, T: Binary> Region<'cx, T> {
+    /// Returns the handle to the region's buffer.
+    pub fn buffer(self) -> Handle<'cx, JsArrayBuffer> {
```
kjvalencik, line 232: I think each of these methods should take `&self` instead of relying on `Copy`.

crates/neon/src/types_impl/buffer/types.rs
dherman: Agreed! I'll tackle that in a separate PR, as you suggest.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,783 @@
...
+
+use crate::{
+    context::{internal::Env, Context},
+    handle::{internal::TransparentNoCopyWrapper, Handle, Managed},
+    object::Object,
+    result::{JsResult, Throw},
+    sys::{self, raw, TypedArrayType},
+    types::buffer::{
+        lock::{Ledger, Lock},
+        private::{self, JsTypedArrayInner},
```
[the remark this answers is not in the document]
dherman: This is confusing because of all the funky re-exporting at the top level `lib.rs` -- `neon::types` actually just _is_ `neon::types_impl`. But I'll change these `use` declarations to import more directly from `types_impl` so it's more self-explanatory.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,783 @@
...
+    ) -> JsResult<'cx, Self>
+    where
+        C: Context<'cx>,
+    {
+        let size = buffer.size(cx);
+        let elt_size = std::mem::size_of::<T>();
+        let len = size / elt_size;
+
+        if (len * elt_size) != size {
+            panic!(
```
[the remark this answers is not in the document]
dherman: I'm open to anything, and a `RangeError` is certainly consistent with what the JS constructors do (e.g. `new Uint32Array(new ArrayBuffer(7))` throws a `RangeError`). I think an error other than `JsResult` would make the API too painful, so that leaves either panicking or throwing a `RangeError`.

I guess my only question is whether it's weird to throw a JavaScript exception from something that seems more like a pure Rust operation. But I think since it's trying to allocate a new JS object, it's reasonable for that to throw a JS exception. So why don't we go with throwing a `RangeError`.

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,783 @@
...
+
+    fn size<'cx, C: Context<'cx>>(&self, cx: &mut C) -> usize {
+        unsafe { sys::arraybuffer::size(cx.env().to_raw(), self.to_raw()) }
+    }
+}
+
+/// A marker trait for all possible element types of binary buffers.
+///
+/// This trait can only be implemented within the Neon library.
+pub trait Binary: private::Sealed + Copy + std::fmt::Debug {
```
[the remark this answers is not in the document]
dherman: Since these bounds are only required for the `PhantomData<T>` field of `JsTypedArrayInner`, I can reduce the requirements from `Debug + Copy` to just `Clone` by manually implementing `Debug`.

I'm a little reluctant to remove the `Clone` super type constraint and propagate it into the declarations, since it pollutes all the impl declarations everywhere. This means a bunch of methods just disappear when your `Binary` type happens not to be `Clone`, which just seems like `JsTypedArray` just doesn't really work as expected unless your `Binary` type is also `Clone`. I feel like I'd rather just express that it's a requirement of the trait.

It is an annoying requirement of the Rust compiler, though. It's frustrating that you can't declare `impl<T: Binary> Copy for JsTypedArrayInner<T> {}` if `T` isn't `Clone`, even though it's only used as a phantom type. (I wonder if that's what [this issue](https://github.com/rust-lang/rust/issues/50980) is about?)

But in the end it seems pretty reasonable to require a `Binary` type to be `Clone`, since binary data is just sequences of bits. What do you think of leaving it at `trait Binary: Sealed + Clone`?

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -0,0 +1,783 @@
...
+impl_typed_array!(Uint16Array, u16, TypedArrayType::U16, U16, JsUint16Array, 2);
+impl_typed_array!(Int32Array, i32, TypedArrayType::I32, I32, JsInt32Array, 2);
+impl_typed_array!(Uint32Array, u32, TypedArrayType::U32, U32, JsUint32Array, 2);
+impl_typed_array!(
+    Float32Array,
+    f32,
+    TypedArrayType::F32,
+    F32,
+    JsFloat32Array,
+    2.0,
```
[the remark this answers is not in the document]
dherman: You can't use an integer literal on the RHS of a `*=` operator for some reason ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9f15f98c91972f6d3dd7b61e8f47db66)). It's just a little extra overhead in the source in order to make the doc example work, so I figured it's tolerable.

test/napi/lib/typedarrays.js
```diff
@@ -0,0 +1,550 @@
...
+  let promise = new Promise((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  DETACH_WORKER.once("message", (message) => {
+    resolve(message);
+  });
+
+  return promise;
```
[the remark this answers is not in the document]
dherman: Did you mean:
```js
return new Promise(resolve => DETACH_WORKER.once("message", resolve));
```
?

crates/neon/src/types_impl/buffer/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+/// let buf = cx.array_buffer(16)?;
+/// let arr = JsUint32Array::from_region(&mut cx, buf.region(4, 2))?;
+/// # Ok(arr)
+/// # }
+/// ```
+#[derive(Clone, Copy)]
+pub struct Region<'cx, T: Binary> {
+    pub(super) buffer: Handle<'cx, JsArrayBuffer>,
+    pub(super) offset: usize,
+    pub(super) len: usize,
```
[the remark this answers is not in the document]
dherman: Good catch, thanks.

crates/neon/src/types_impl/buffer/mod.rs
dherman: I've left it as `Copy` (which you mentioned on Slack is your preference) but changed all the methods to take `&Region`. I'm persuadable either way on `Copy`.

crates/neon/src/types_impl/buffer/types.rs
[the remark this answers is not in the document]
kjvalencik: That makes sense to me. Thanks for the explanation!

test/napi/lib/typedarrays.js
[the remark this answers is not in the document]
kjvalencik: Yes, I did.

crates/neon/src/types_docs.rs
```diff
@@ -0,0 +1,142 @@
+#[cfg_attr(aquamarine, aquamarine::aquamarine)]
```
kjvalencik, line 1: I think this should be `feature = "aquamarine"`.  This isn't currently being rendered in the docs.

https://deploy-preview-909--neon-bindings-docs.netlify.app/neon/types/index.html#the-javascript-type-hierarchy

.cargo/config.toml
```diff
@@ -3,5 +3,5 @@
 # The following aliases simplify linting the entire workspace
 neon-check = " check  --all --all-targets --features napi-experimental,futures"
 neon-clippy = "clippy --all --all-targets --features napi-experimental,futures -- -A clippy::missing_safety_doc"
-neon-test = "  test   --all               --features=napi-experimental,futures"
-neon-doc = "   rustdoc -p neon            --features=napi-experimental,futures -- --cfg docsrs"
+neon-test = "  test   --all               --features=doc-comment,napi-experimental,futures"
```
kjvalencik, line 6: I don't think `doc-comment` should be here.

dherman: The reason it's in there is so the doc tests run for the API docs generated by the doc-comment. I verified that this works.

crates/neon/src/types_docs.rs
dherman: Great catch, thanks. Pushed a fix.

PR #910, opened 2022-06-23
doc(neon): Use anyhow instead of failure in a doc example

MikuroXina:
[`failure` crate](https://github.com/rust-lang-deprecated/failure) is deprecated and unmaintained now, so I replace the code using `failure::Error` with `anyhow::{Context, Result}`.


kjvalencik:
Great clean-up! Thanks!

PR #912, opened 2022-07-01
Bump dependency versions

kjvalencik:


crates/neon/Cargo.toml
```diff
@@ -21,7 +21,7 @@ nodejs-sys = "0.13.0"
 
 [dependencies]
 libloading = "0.7.3"
-semver = "0.9.0"
+semver = "1"
```
kjvalencik, line 24: This is a breaking change since `semver::Version` is exposed in Neon's public API, but we already have some breaking changes staged and this seemed pretty minor.

We may want to consider completely removing the `meta` feature prior to 1.0 since users could leverage the [auditable](https://crates.io/crates/auditable) crate for this.

pkgs/create-neon/test/create-neon.ts
```diff
@@ -1,17 +1,21 @@
 import { assert } from "chai";
 import { spawn } from "child_process";
-import execa from "execa";
 import * as path from "path";
 import { promises as fs } from "fs";
 import * as TOML from "toml";
 import expect from "../dev/expect";
 import rimraf from "../dev/rimraf";
 
+async function execa(command: string, args: readonly string[]): Promise<void> {
```
kjvalencik, line 9: `execa` is an ESM. Rather than adding complexity to building--because Node 14 doesn't provide a built-in loader--I chose to add a simple wrapper.

crates/neon/Cargo.toml
dherman: That's a good point, and just the fact that we're exporting public API that uses types from another crate (`semver`) creates some risk.

pkgs/create-neon/test/create-neon.ts
dherman: Maybe just add a comment that we can remove this shim once we drop support for enough old versions of Node.

dherman (approved):
Just one minor suggestion, otherwise looks good.

PR #913, opened 2022-07-06
fix: Fix race condition on shutdown

kjvalencik:


crates/neon/src/sys/no_panic.rs
```diff
@@ -121,6 +121,12 @@ impl FailureBoundary {
     }
 }
 
+// HACK: Force `NAPI_PREAMBLE` to run without executing any JavaScript to tell if it's
```
dherman, line 124: I'd add one more line to the comment that explains that `NAPI_PREAMBLE` is a macro in the implementation of Node-API that checks if it's legal to call into JS. Or at least link to something so someone can read more about what it is.

test/napi/lib/workers.js
```diff
@@ -10,6 +10,9 @@ const addon = require("..");
 
 // Receive a message, try that method and return the error message
 if (!isMainThread) {
+  // Attempt to reproduce shutdown race condition bug
+  addon.reject_after(new Error("Oh, no!"), 200).catch(() => {});
```
dherman, line 14: This is technically a nondeterministic test, so it's worth at least leaving some kind of hint in the comments that it could become flaky. Maybe just a tag like `// RACE: Attempt to reproduce shutdown race condition bug`?

dherman (approved):
Looks great, just a couple small suggestions for the comments.

PR #914, opened 2022-07-07
fix: Make native modules built with Neon work in Bun

kjvalencik:
[bun](https://bun.sh/) is a new JavaScript runtime designed to be Node.js compatible--including Node-API. However, it currently ships with an incomplete set of Node-APIs.

The following changes symbol loading to print a warning to `stderr` if a symbol is missing instead of panicking. This should be no change on Node because symbols are guaranteed to exist if the Node-API version number matches. On `bun`, this converts a panic on load to a panic when attempting to use the API.

With these changes, simple Neon modules work with `bun`. However, because of Neon's heavily reliance on `napi_get_instance_data`, many more powerful features will not work until this API is implemented in bun.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ Neon projects to 0.10!
 Support for [LTS versions of Node](https://github.com/nodejs/LTS#release-schedule) and current are expected. If you're
 using a different version of Node and believe it should be supported, let us know.
 
+### Bun (experimental)
+
+[Bun](https://bun.sh/) is an alternate JavaScript runtime that targets Node compatibility. In many cases Neon modules will work in bun; however, at the time of this writing, some Node-API functions are [not implemented](https://github.com/Jarred-Sumner/bun/issues/158).
+
 ### Rust
 
 Neon supports Rust stable version 1.18 and higher. We test on the latest stable, beta, and nightly versions of Rust.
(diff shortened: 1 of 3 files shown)


crates/neon/src/sys/bindings/mod.rs
```diff
@@ -139,21 +139,31 @@ macro_rules! generate {
...
+        ) {
             assert!(
                 actual_napi_version >= expected_napi_version,
-                "Minimum required N-API version {}, found {}.",
+                "Minimum required Node-API version {}, found {}.",
                 expected_napi_version,
                 actual_napi_version,
             );
 
+            let print_warn = |err| eprintln!("WARN: {}", err);
```
kjvalencik, line 150: Sample output:

```
WARN: bun: undefined symbol: napi_get_value_external
WARN: bun: undefined symbol: napi_close_escapable_handle_scope
WARN: bun: undefined symbol: napi_open_escapable_handle_scope
WARN: bun: undefined symbol: napi_create_external
WARN: bun: undefined symbol: napi_escape_handle
WARN: bun: undefined symbol: napi_set_instance_data
WARN: bun: undefined symbol: napi_get_instance_data
```

README.md
```diff
@@ -49,6 +49,10 @@ Neon projects to 0.10!
 Support for [LTS versions of Node](https://github.com/nodejs/LTS#release-schedule) and current are expected. If you're
 using a different version of Node and believe it should be supported, let us know.
 
+### Bun
```
dherman, line 52:
```suggestion
### Bun (experimental)
```

dherman (approved):
Looks good! I made one README suggestion.

PR #919, opened 2022-07-22
fix: More accurate lifetime bounds on scoped methods to allow valid code to compile

kjvalencik:
Prior to this change, it was impossible to escape a value in `execute_scoped` that was created in an outer scope. This is because the
function definition allowed for the outer and inner lifetimes to be entirely unrelated.

This is technically a breaking change, but in many cases the more strict bound will cause code to be more likely to compile.

The following code compiles *after* this change, but not before:

```rust
fn example(mut cx: FunctionContext) -> JsResult<JsValue> {
    let value = cx.argument::<JsValue>(0)?;

    cx.compute_scoped(move |_| Ok(value))
}
```

### DIFF
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -242,9 +242,10 @@ pub trait Context<'a>: ContextInternal<'a> {
     /// Handles created in the new scope are kept alive only for the duration of the computation and cannot escape.
     ///
     /// This method can be useful for limiting the life of temporary values created during long-running computations, to prevent leaks.
-    fn execute_scoped<T, F>(&mut self, f: F) -> T
+    fn execute_scoped<'b, T, F>(&mut self, f: F) -> T
     where
-        F: for<'b> FnOnce(ExecuteContext<'b>) -> T,
+        'a: 'b,
+        F: FnOnce(ExecuteContext<'b>) -> T,
     {
         let env = self.env();
         let scope = unsafe { HandleScope::new(env.to_raw()) };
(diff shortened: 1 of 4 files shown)


crates/neon/src/context/mod.rs
```diff
@@ -585,19 +586,18 @@ impl<'a> ContextInternal<'a> for ExecuteContext<'a> {
 impl<'a> Context<'a> for ExecuteContext<'a> {}
 
 /// An execution context of a scope created by [`Context::compute_scoped()`](Context::compute_scoped).
-pub struct ComputeContext<'a, 'outer> {
+pub struct ComputeContext<'a> {
     env: Env,
     phantom_inner: PhantomData<&'a ()>,
-    phantom_outer: PhantomData<&'outer ()>,
```
kjvalencik, line 591: This was required before to ensure that `ComputeContext` couldn't accidentally outlive the outer context. However, `compute_scoped` now includes a `'a: 'b` bound that provides the same guarantee.

crates/neon/src/context/mod.rs
```diff
@@ -242,9 +242,10 @@ pub trait Context<'a>: ContextInternal<'a> {
     /// Handles created in the new scope are kept alive only for the duration of the computation and cannot escape.
     ///
     /// This method can be useful for limiting the life of temporary values created during long-running computations, to prevent leaks.
-    fn execute_scoped<T, F>(&mut self, f: F) -> T
+    fn execute_scoped<'b, T, F>(&mut self, f: F) -> T
```
kjvalencik, line 245: This change doesn't provide as much tangible value, but I thought it was good to keep them aligned.

dherman (approved):
I know it's not a huge deal but the fact that you were able to eliminate the extra phantom lifetime feels like an affirmation that it's a cleanup of the previous signature.

PR #922, opened 2022-09-14
Speed-Up CI

kjvalencik:
This makes a few changes to our CI process to make it run faster.

* Adds caching of artifacts (Rust, NPM, Electron)
* Collapses to a single CI file for tests to minimize duplication
* Only run on Node LTS and Rust Stable in PRs
  - Runs the full matrix if the "full matrix" label is provided
  - Runs the full matrix on Linux
* Runs the full matrix when *merged*

Try changing the labels on the PR to test it out!

**Note**: Matrix includes do not include arrays for a 2nd pass of expansion; instead they become literal arrays. As a workaround, I included a JavaScript snippet in the setup step that performs a manual expansion of the matrix.


dherman (approved):
This is awesome! Only downside I see is that the config files are getting harder to follow. The Actions YAML syntax doesn't give us much room for abstractions, though. Not sure if you want to add any clarifying comments. Anyway, that's your call, and otherwise 🚢

kjvalencik:
@dherman You can write a custom action and include it with your project, giving a path for the action. The action can then invoke other actions. AFAIK, this is the only way to share a snippet which I would consider too complicated for the amount of duplication here.

If these eventually get much more complicated, we could consider that.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+    steps:
+      - name: Set Matrix
+        id: set_matrix
+        run: |
+          node << 'EOF'
+            const os = ["ubuntu-latest", "windows-latest", "macos-latest"];
+            const rustToolchains = ["stable", "nightly"];
+            const nodeVersions = ["14.x", "16.x", "18.x"];
+            const partialRustToolchains = ["stable"];
+            const partialNodeVersions = ["16.x"];
```
kjvalencik, line 32: @dherman Do you think these constants are clear enough embedded here? I originally had them in environment variables, but the logic to read and parse them added a bunch more code without doing anything but, moving them from YAML to environment variables.

Since Github Actions doesn't support array literals, they were a JSON string even when in YAML.

```yaml
env:
    RUST_TOOLCHAINS: '["stable", "nightly"]'
```

dherman (approved):
:shipit: !

PR #925, opened 2022-09-16
`TypedArray::from_slice()`

dherman:
This PR follows on #909 with an additional `TypedArray::from_slice()` method, making it convenient and easy to copy from a Rust slice into a new typed array.

This is not only convenient, but also mitigates the ergonomic cost of deprecating `JsArrayBuffer::external()`, which seems like it'll be necessary due to recent changes in V8 to deprecate external typed arrays.

### DIFF
--- a/crates/neon/src/types_impl/buffer/mod.rs
+++ b/crates/neon/src/types_impl/buffer/mod.rs
@@ -14,7 +14,7 @@ use crate::{
     result::{JsResult, NeonResult, ResultExt},
     types::{
         buffer::lock::{Ledger, Lock},
-        JsArrayBuffer, JsTypedArray,
+        JsArrayBuffer, JsTypedArray, Value,
     },
 };
 
@@ -47,7 +47,7 @@ pub use types::Binary;
 /// ```
 ///
 /// [typed-arrays]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays
-pub trait TypedArray: private::Sealed {
+pub trait TypedArray: private::Sealed + Value {
     type Item: Binary;
 
     /// Statically checked immutable borrow of binary data.
@@ -91,6 +91,11 @@ pub trait TypedArray: private::Sealed {
 
     /// Returns the size, in bytes, of the allocated binary data.
     fn size<'cx, C>(&self, cx: &mut C) -> usize
+    where
+        C: Context<'cx>;
+
+    /// Constructs an instance from a slice by copying its contents.
+    fn from_slice<'cx, C>(cx: &mut C, slice: &[Self::Item]) -> JsResult<'cx, Self>
     where
         C: Context<'cx>;
 }
--- a/crates/neon/src/types_impl/buffer/mod.rs
+++ b/crates/neon/src/types_impl/buffer/mod.rs
@@ -47,7 +47,7 @@ pub use types::Binary;
 /// ```
 ///
 /// [typed-arrays]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays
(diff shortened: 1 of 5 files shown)


dherman:
I tried adding `from_slice()` to the TypedArray trait, but that trait doesn't extend `Value` and `JsTypedArray<T>` isn't declared to universally implement `Value`, which makes satisfying the type checker a challenge when trying to add a method that returns `JsResult<Self>`.

dherman:
Update: I've moved the `from_slice()` method into `TypedArray` (thanks to @kjvalencik's help leveling up my understanding of the power of `where` clauses).

test/napi/src/js/typedarrays.rs
```diff
@@ -8,6 +8,18 @@ pub fn return_array_buffer(mut cx: FunctionContext) -> JsResult<JsArrayBuffer> {
...
+    let len: Handle<JsNumber> = cx.argument(0)?;
+    let len: f64 = len.value(&mut cx);
+    let len: u32 = len as u32;
+    let mut v: Vec<u8> = Vec::new();
+    for i in 0..len {
+        v.push(i as u8);
+    }
+    let b: Handle<JsArrayBuffer> = JsArrayBuffer::from_slice(&mut cx, &v)?;
+    Ok(b)
+}
```
kjvalencik, line 21: FYI, this can be simplified a bit with an iterator collect.

```suggestion
pub fn return_array_buffer_from_slice(mut cx: FunctionContext) -> JsResult<JsArrayBuffer> {
    let len = cx.argument::<JsNumber>(0)?.value(&mut cx) as usize;
    let v = (0..len).map(|i| i as u8).collect::<Vec<_>>();

    JsArrayBuffer::from_slice(&mut cx, &v)
}
```

crates/neon/src/types_impl/buffer/types.rs
```diff
@@ -475,7 +496,11 @@ impl<T: Binary> Managed for JsTypedArray<T> {
     }
 }
 
-impl<T: Binary> TypedArray for JsTypedArray<T> {
+impl<T> TypedArray for JsTypedArray<T>
+where
+    T: Binary + Copy,
```
kjvalencik, line 501: Do you think it makes sense to make `Copy` a supertype of `Binary` instead of adding the bound here?

crates/neon/src/types_impl/buffer/mod.rs
```diff
@@ -47,7 +47,7 @@ pub use types::Binary;
 /// ```
 ///
 /// [typed-arrays]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays
-pub trait TypedArray: private::Sealed {
+pub trait TypedArray: private::Sealed + Value {
```
kjvalencik, line 50: We could drop `Sealed` here since `Value` is already sealed with `ValueInternal`.

kjvalencik (commented):
It would be nice to have some docs added to `JsTypedArray` (and possibly a mirrored `JsTypedArray::from_slice` method) to help users find this constructor since it's not _typical_ in Rust for a types constructor to be on a trait.

crates/neon/src/types_impl/buffer/types.rs
dherman: I originally did… you asked me to remove it 😝

kjvalencik: 😆 It made sense in the last PR, but then once we introduced a `Value` constraint, the separation was less valuable.

dherman: I dig it 👍

kjvalencik (approved):
Awesome! Love these improvements! :shipit:

PR #928, opened 2022-09-30
cargo clippy & fix rust warning: unused return value of `Box::<T>::from_raw` that must be used

usrtax:
image

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,10 @@
+edition = "2018"
+tab_spaces = 4
+unstable_features = true
+condense_wildcard_suffixes = true
+newline_style = "Unix"
+use_field_init_shorthand = true
+use_try_shorthand = true
+imports_granularity = "Crate"
+group_imports = "StdExternalCrate"
+reorder_imports = true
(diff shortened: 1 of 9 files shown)


.rustfmt.toml
```diff
@@ -0,0 +1,10 @@
+edition = "2021"
```
kjvalencik, line 1: Can you drop this file? We would prefer to stay on the defaults.

crates/neon-macros/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Dave Herman <david.herman@gmail.com>"]
 description = "Procedural macros supporting Neon"
 repository = "https://github.com/neon-bindings/neon"
 license = "MIT/Apache-2.0"
-edition = "2018"
+edition = "2021"
```
kjvalencik, line 8: Please revert this change until we can discuss bumping the minimum supported rust version.

crates/neon/Cargo.toml
```diff
@@ -79,7 +79,7 @@ doc-dependencies = ["doc-comment", "aquamarine"]
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 features = [
-    "futures",
-    "napi-experimental",
-    "doc-dependencies",
+  "futures",
```
kjvalencik, line 82: Please revert this unrelated change.

kjvalencik (changes requested):
Thanks for cleaning up these clippy warnings! 🎉 

I asked for a couple minor changes to back out unrelated diffs.

PR #934, opened 2022-10-31
Fix two cases that can panic during shutdown

rf-figma:
This PR reverts the revert of #932, adding a second commit that should fix a related race condition that was causing a build failure.

### DIFF
--- a/test/napi/lib/workers.js
+++ b/test/napi/lib/workers.js
@@ -4,6 +4,7 @@ const {
   isMainThread,
   parentPort,
   threadId,
+  workerData,
 } = require("worker_threads");
 
 const addon = require("..");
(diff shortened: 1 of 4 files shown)


test/napi/lib/workers.js
```diff
@@ -194,9 +203,13 @@ describe("Instance-local storage", () => {
...
-    const worker = new Worker(__filename);
+    const worker = new Worker(__filename, {
+      workerData: "notify_when_startup_complete",
+    });
 
-    setTimeout(() => worker.terminate(), 50);
-    setTimeout(cb, 100);
+    worker.once("message", async () => {
+      await worker.terminate();
+      cb();
```
kjvalencik, line 212: I believe this still needs to be delayed a bit for the race.

kjvalencik:
Merged via https://github.com/neon-bindings/neon/commit/000b6c532452c46a22ac1c131952f84159a59001

PR #938, opened 2022-11-07
Remove unused `cx.null()` from channel docs

dnaka91:
First of all, thank you for this awesome crate, I love it. Currently use it to build an extension for VS Code, and this crate allows me to write most of it in Rust. Awesome work!

While starting to use the event APIs to be able to do background tasks with callback, I noticed that the docs call `cx.null()`, but never actually use the created instance. Seems it's not needed, so probably a leftover or so.

### DIFF
--- a/crates/neon/src/event/channel.rs
+++ b/crates/neon/src/event/channel.rs
@@ -76,7 +76,6 @@ type Callback = Box<dyn FnOnce(Env) + Send + 'static>;
 ///         channel.send(move |mut cx| {
 ///             let callback = callback.into_inner(&mut cx);
 ///             let this = cx.undefined();
-///             let null = cx.null();
 ///             let args = vec![
 ///                 cx.null().upcast::<JsValue>(),
 ///                 cx.number(result).upcast(),
--- a/crates/neon/src/event/mod.rs
+++ b/crates/neon/src/event/mod.rs
@@ -75,7 +75,6 @@
 //!     channel.send(move |mut cx| {
 //!         let callback = callback.into_inner(&mut cx);
 //!         let this = cx.undefined();
-//!         let null = cx.null();
 //!         let args = match result {
 //!             Ok(psd) => {
 //!                 // Extract data from the parsed file.


kjvalencik:
Good find! Yes, these are definitely left over from a previous refactor. Additionally, they are using the older low-level function calls instead of the newer more ergonomic ones. For example, I might instead do something like:

```rust
///         channel.send(move |mut cx| {
///             callback.into_inner(&mut cx)
///                 .call_with(&cx)
///                 .args((cx.null(), cx.number(result)))
///                 .exec(&mut cx)
///         });
```

```rust
//!     channel.send(move |mut cx| {
//!         let mut callback = callback
//!             .into_inner(&mut cx)
//!             .call_with(&cx);
//!
//!         let psd = match result {
//!             Ok(psd) => psd,
//!             Err(err) => return callback.arg(cx.string(err.to_string())).exec(&mut cx),
//!         };
//!
//!         // Extract data from the parsed file.
//!         let width = cx.number(psd.width());
//!         let height = cx.number(psd.height());
//!
//!         // Save the data in a result object.
//!         let obj = cx.empty_object();
//!         obj.set(&mut cx, "width", width)?;
//!         obj.set(&mut cx, "height", height)?;
//!
//!         callback.args((cx.null(), obj)).exec(&mut cx)
//!     });
```

However, it would be even better to do these with the promise API and task APIs now that has been built. The first example becomes very simple:

```rust
/// fn async_fibonacci(mut cx: FunctionContext) -> JsResult<JsPromise> {
///     let n = cx.argument::<JsNumber>(0)?.value(&mut cx);
///     let promise = cx
///         .task(move || fibonacci(n))
///         .promise(|mut cx| Ok(cx.number(result)));
/// 
///     Ok(promise)
/// }
```

Even keeping the Rust thread--instead of a `libuv` thread--it's still simpler:

```rust
/// fn async_fibonacci(mut cx: FunctionContext) -> JsResult<JsPromise> {
///     // These types (`f64`, `Root<JsFunction>`, `Channel`) may all be sent
///     // across threads.
///     let n = cx.argument::<JsNumber>(0)?.value(&mut cx);
///     let (deferred, promise) = cx.promise();
///     let channel = cx.channel();
///
///     // Spawn a thread to complete the execution. This will _not_ block the
///     // JavaScript event loop.
///     std::thread::spawn(move || {
///         let result = fibonacci(n);
///
///         deferred.settle_with(&channel, |mut cx| Ok(cx.number(result)));
///     });
///
///     Ok(promise)
/// }
```

PR #940, opened 2022-11-09
Release 1.0.0-alpha.2

kjvalencik:


RELEASES.md
```diff
@@ -1,3 +1,47 @@
...
+
+## Breaking Changes
+
+### `neon::object::This`
+
+https://github.com/neon-bindings/neon/pull/918
+
+Trait [`neon::object::This`](https://docs.rs/neon/latest/neon/object/trait.This.html) has been removed. `This` was primarily added for use with the `declare_types!` macro to generate classes. The macro was removed and `This` is no longer needed. Additionally, the `This` argument on `JsFunction` was found to be _invalid_ because it asserted at compile time a type for `this` that could change at runtime.
+
+**Note**: This is _not_ unsound because the type would be checked by Node-API and result in a `panic`.
```
dherman, line 11: Since this is functionality we've removed, it's probably not worth a whole separate line and bold comment. I'd just put it in a parenthetical at the end of the previous paragraph and use the past tense:

```
...a type for `this` that could change at runtime. (Note that this
was _not_ unsound because the type would be checked by Node-API
and result in a `panic`.)
```

RELEASES.md
```diff
@@ -1,3 +1,47 @@
...
+## Improvements
+
+* Lifetimes were relaxed on `execute_scoped` to allow valid code to compile. (https://github.com/neon-bindings/neon/pull/919)
+* Added a `from_slice` helper on `TypedArray` (https://github.com/neon-bindings/neon/pull/925)
+* `JsTypedArray` construction and type aliases (https://github.com/neon-bindings/neon/pull/909)
+
+## Bug Fixes
+
+* Fixed a panic on VM shutdown when using `Channel` (https://github.com/neon-bindings/neon/pull/934)
+* Type tags were added to `JsBox` to reduce the likelihood of undefined behavior when multiple native add-ons are used (https://github.com/neon-bindings/neon/pull/907)
```
dherman, line 34: s/reduce the likelihood of/prevent/

Yes, it's still technically possible but it's vanishingly unlikely. I think it's fair to call this prevention.

RELEASES.md
```diff
@@ -1,3 +1,47 @@
...
+
+## Docs
+
+* Significantly improved documentation of `TypedArray` (https://github.com/neon-bindings/neon/pull/909)
+* Removed unused values in `Channel` docs (https://github.com/neon-bindings/neon/pull/925)
+
+### `cargo-cp-artifact`
+
+`0.1.7` includes a fix to unlink `.node` files before copying to reduce the chances of a code signing error on macOS (https://github.com/neon-bindings/neon/pull/921).
```
dherman, line 43: Similarly, I think we could s/reduce the chances of a code signing error/address code signing errors/

dherman (changes requested):
Exciting release!

RELEASES.md
kjvalencik: > address _common_ code signing errors

RELEASES.md
kjvalencik: Neon uses feature detection and type tags are backported all the way to 12. I think this wording is good.

dherman (approved):
:shipit: 🚀

PR #942, opened 2022-11-14
API docs improvements

dherman:
Improvements to API docs:

- [x] Align wording for all `neon::types` doc summaries.
- [x] Examples for all `neon::types` types.
  - [x] `JsArray`
  - [x] `JsBoolean`
  - [x] `JsDate`
  - [x] `JsError`
  - [x] `JsFunction`
  - [x] `JsFuture`
  - [x] `JsNull`
  - [x] `JsNumber`
  - [x] `JsObject`
  - [x] `JsPromise`
  - [x] `JsString`
  - [x] `JsUndefined`
  - [x] `JsValue`

----

<details><summary>Main <code>neon::types</code> page before:</summary>
API docs page before
</details>

<details><summary>Main <code>neon::types</code> page after:</summary>
API docs page after
</details>

PR #944, opened 2022-11-15
Add functions to get JsString as UTF-16

### LINKED ISSUE #943: Support for UTF-16 strings
I'm using Neon to build a VS Code extension, and I am listening for editing events on open text editors. While doing so, I realized that JavaScript strings are represented as UTF-16.

Usually that's not an issue, because Neon uses the NAPI APIs that auto-convert to UTF-8 and the string can be used as usual in Rust.

Now these editing events include line and character positions, and sadly those character positions are not actual char positions but UTF-16 byte positions. So I'll have to adjust them to UTF-8 lengths to work with them in Rust, which I currently do in JavaScript/TypeScript.

I saw that the NAPI has `napi_get_value_string_utf16`, so it would be possible to get the UTF-16 variant of the string, and I could do all these conversions directly in Rust.

The standard library doesn't have any specific UTF-16 string type, but the [`widestring`](https://docs.rs/widestring/latest/widestring) crate seems very popular and has what would be needed for this. Probably preferred to hide this behind a feature flag (although, the crate has zero deps).

---

It would be great if we can add APIs for the UTF-16 string variants of this. Maybe not even necessary to add `widestring`, and could possibly just expose a `to_string_utf16()` on the `JsString` type that returns a `Vec<u16>` 🤔. The `widestring` variant would then return a [`Utf16String`](https://docs.rs/widestring/latest/widestring/utfstring/struct.Utf16String.html).

What do you think about this idea? I could prepare a PR for this, if there is interest to extend the APIs a bit.

dnaka91:
Extending the API of `JsString` to retrieve the value as UTF-16 encoded `Vec<u16>`. Also, added a `size_utf16` function for completeness.

I added a few code samples to the API docs for `JsString` to show at least some basic uses, as the docs felt a bit empty for this rather core component.

Fixes #943

### DIFF
--- a/crates/neon/src/sys/bindings/functions.rs
+++ b/crates/neon/src/sys/bindings/functions.rs
@@ -65,6 +65,9 @@ mod napi1 {
                 result: *mut usize,
             ) -> Status;
 
+            // The `buf` argument is defined as a `char16_t` which _should_ be a `u16` on most
+            // platforms. When generating bindings with `rust-bindgen` it unconditionally defines
+            // it as `u16` as well.
             fn get_value_string_utf16(
                 env: Env,
                 value: Value,
--- a/crates/neon/src/sys/bindings/functions.rs
+++ b/crates/neon/src/sys/bindings/functions.rs
@@ -65,6 +65,14 @@ mod napi1 {
                 result: *mut usize,
             ) -> Status;
 
+            fn get_value_string_utf16(
+                env: Env,
+                value: Value,
+                buf: *mut u16,
+                bufsize: usize,
+                result: *mut usize,
+            ) -> Status;
+
             fn create_type_error(env: Env, code: Value, msg: Value, result: *mut Value) -> Status;
 
             fn create_range_error(env: Env, code: Value, msg: Value, result: *mut Value) -> Status;
--- a/crates/neon/src/sys/string.rs
+++ b/crates/neon/src/sys/string.rs
@@ -11,41 +11,41 @@ pub unsafe fn new(out: &mut Local, env: Env, data: *const u8, len: i32) -> bool 
     status == napi::Status::Ok
 }
 
-pub unsafe fn utf8_len(env: Env, value: Local) -> isize {
+pub unsafe fn utf8_len(env: Env, value: Local) -> usize {
     let mut len = MaybeUninit::uninit();
     let status = napi::get_value_string_utf8(env, value, ptr::null_mut(), 0, len.as_mut_ptr());
 
     assert_eq!(status, napi::Status::Ok);
 
-    len.assume_init() as isize
+    len.assume_init()
 }
 
-pub unsafe fn data(env: Env, out: *mut u8, len: isize, value: Local) -> isize {
+pub unsafe fn data(env: Env, out: *mut u8, len: usize, value: Local) -> usize {
     let mut read = MaybeUninit::uninit();
     let status =
-        napi::get_value_string_utf8(env, value, out as *mut _, len as usize, read.as_mut_ptr());
+        napi::get_value_string_utf8(env, value, out as *mut _, len, read.as_mut_ptr());
 
     assert_eq!(status, napi::Status::Ok);
 
-    read.assume_init() as isize
+    read.assume_init()
 }
 
-pub unsafe fn utf16_len(env: Env, value: Local) -> isize {
+pub unsafe fn utf16_len(env: Env, value: Local) -> usize {
     let mut len = MaybeUninit::uninit();
     let status = napi::get_value_string_utf16(env, value, ptr::null_mut(), 0, len.as_mut_ptr());
 
     assert_eq!(status, napi::Status::Ok);
 
-    len.assume_init() as isize
+    len.assume_init()
 }
 
-pub unsafe fn data_utf16(env: Env, out: *mut u16, len: isize, value: Local) -> isize {
+pub unsafe fn data_utf16(env: Env, out: *mut u16, len: usize, value: Local) -> usize {
     let mut read = MaybeUninit::uninit();
-    let status = napi::get_value_string_utf16(env, value, out, len as usize, read.as_mut_ptr());
+    let status = napi::get_value_string_utf16(env, value, out, len, read.as_mut_ptr());
 
     assert_eq!(status, napi::Status::Ok);
 
-    read.assume_init() as isize
+    read.assume_init()
 }
 
 pub unsafe fn run_script(out: &mut Local, env: Env, value: Local) -> bool {
--- a/crates/neon/src/sys/string.rs
+++ b/crates/neon/src/sys/string.rs
@@ -30,6 +30,24 @@ pub unsafe fn data(env: Env, out: *mut u8, len: isize, value: Local) -> isize {
     read.assume_init() as isize
 }
 
+pub unsafe fn utf16_len(env: Env, value: Local) -> isize {
+    let mut len = MaybeUninit::uninit();
+    let status = napi::get_value_string_utf16(env, value, ptr::null_mut(), 0, len.as_mut_ptr());
+
+    assert_eq!(status, napi::Status::Ok);
+
+    len.assume_init() as isize
+}
+
+pub unsafe fn data_utf16(env: Env, out: *mut u16, len: isize, value: Local) -> isize {
+    let mut read = MaybeUninit::uninit();
+    let status = napi::get_value_string_utf16(env, value, out, len as usize, read.as_mut_ptr());
+
+    assert_eq!(status, napi::Status::Ok);
+
+    read.assume_init() as isize
+}
+
 pub unsafe fn run_script(out: &mut Local, env: Env, value: Local) -> bool {
     let status = napi::run_script(env, value, out as *mut _);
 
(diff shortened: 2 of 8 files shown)


kjvalencik:
@dherman can you take a look at the docs in this since you are actively working on them in https://github.com/neon-bindings/neon/pull/942?

crates/neon/src/sys/bindings/functions.rs
```diff
@@ -65,6 +65,14 @@ mod napi1 {
                 result: *mut usize,
             ) -> Status;
 
+            fn get_value_string_utf16(
+                env: Env,
+                value: Value,
+                buf: *mut u16,
```
kjvalencik, line 71: I'm not super strong in C++. Is `char16_t` practically always `u16`? The spec seems to define it as "at _least_ a u16." It looks like `rust-bindgen` unconditionally makes it a `u16`.

crates/neon/src/sys/string.rs
```diff
@@ -30,6 +30,24 @@ pub unsafe fn data(env: Env, out: *mut u8, len: isize, value: Local) -> isize {
     read.assume_init() as isize
 }
 
+pub unsafe fn utf16_len(env: Env, value: Local) -> isize {
```
kjvalencik, line 33: It looks like a mistake that we are casting to `isize` in the UTF-8 functions. I would prefer to fix them and have the breaking change prior to 1.0.

crates/neon/src/types_impl/mod.rs
```diff
@@ -362,6 +424,54 @@ impl JsString {
...
+    ///     println!("JavaScript string as UTF-16: {}", s);
+    ///
+    ///     Ok(cx.undefined())
+    /// }
+    /// ```
+    pub fn to_utf16<'a, C: Context<'a>>(&self, cx: &mut C) -> Vec<u16> {
+        let env = cx.env().to_raw();
+
+        unsafe {
+            let capacity = sys::string::utf16_len(env, self.to_raw()) + 1;
```
kjvalencik, line 466: I don't remember exactly how the Node-API string functions work. Do they unconditionally write a null byte at the end?

crates/neon/src/types_impl/mod.rs
```diff
@@ -362,6 +424,54 @@ impl JsString {
...
+            let mut buffer: Vec<u16> = Vec::with_capacity(capacity as usize);
+            let p = buffer.as_mut_ptr();
+            std::mem::forget(buffer);
```
kjvalencik, line 469: I see this is how the existing function is written and appreciate keeping things consistent. However, when reviewing this, I noticed that this could result in a memory leak if `data_utf16` panics due a non-`Ok` response.

The nice side is this protects against UB if Node-API continues to write to the data after the panic, but that seems a bit extreme. This _seems_ like it could be simplified to using `unsafe fn set_len`.

@dherman what do you think? Is this safety a bit too extreme?

crates/neon/src/types_impl/mod.rs
```diff
@@ -362,6 +424,54 @@ impl JsString {
...
+    ///
+    /// Again a function that expects a single JavaScript string as argument, but utilizes the
+    /// [`widestring`](https://crates.io/crates/widestring) crate to handle the raw [`Vec<u16>`] as
+    /// a typical string.
+    ///
+    /// ```rust
+    /// # use neon::prelude::*;
+    /// fn print_with_widestring(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    ///     let s = cx.argument::<JsString>(0)?.to_utf16(&mut cx);
+    ///     let s = unsafe { widestring::Utf16String::from_vec_unchecked(s) };
```
kjvalencik, line 456: It's mentioned above, but can you also add a comment here explaining the safety of this block?

test/napi/lib/strings.js
```diff
@@ -5,6 +5,14 @@ describe("JsString", function () {
   it("should return a JsString built in Rust", function () {
     assert.equal(addon.return_js_string(), "hello node");
   });
+  describe("encoding", function () {
+    it("should return the UTF-8 string length", function () {
+      assert.equal(addon.return_length_utf8("a🥹"), 5);
+    });
+    it("should return the UTF-16 string length", function () {
```
kjvalencik, line 12: I appreciate all these great tests and this isn't necessary, but I think it would be valuable to have a test as an example that uses `TextDecoder` (or `Buffer`) to decode UTF-16 to a string.

crates/neon/src/sys/bindings/functions.rs
dnaka91: I was worried about that one too. Thought maybe even the `c_short` type is a better fit. As you said, [rust-bindgen](https://github.com/rust-lang/rust-bindgen/blob/678daf56a2384252304fc27ed2b46e769da1890f/bindgen/ir/int.rs#L57) defines it as  `u16` as well, so I'd say it's correct 🤔

crates/neon/src/sys/string.rs
dnaka91: I saw that and wondered why `isize` over `usize` here, but simply followed the existing code to keep it consistent. I'll change both to `usize` then.

crates/neon/src/types_impl/mod.rs
dnaka91: The docs were not 100% clear about it, only talking about it in terms of truncation of insufficient buffers. So I had a peek at the source, and they indeed always add a null byte at the end.

Getting the length only doesn't include the extra byte for the null byte at the end, so we have to add `1` to the length here, or we'd lose the last character.

crates/neon/src/types_impl/mod.rs
dnaka91: Sure thing :+1:

crates/neon/src/types_impl/mod.rs
dnaka91: Is the `std::mem::forget` actually needed at all?

There is probably some specific case that I'm missing here. Maybe related to panics as well?

I personally would have just passed the pointer `p` to `sys::string::data`/`sys::string::data_utf16`, without calling anything in between, then calling `set_len` as you mentioned, and last depending on the function either:
- `value`: use `String::from_utf8_unchecked` to turn the `Vec<u8>` into a `String` afterwards.
- `to_utf16`: Nothing, because it's already the expected `Vec<u16>`.

```rust
// in `value`
unsafe {
    let capacity = sys::string::utf8_len(env, self.to_raw()) + 1;
    let mut buffer: Vec<u8> = Vec::with_capacity(capacity as usize);
    let len = sys::string::data(env, buffer.as_mut_ptr(), capacity, self.to_raw());
    buffer.set_len(len as usize);
    String::from_utf8_unchecked(buffer) // turn into a String
}

// in `to_utf16`
unsafe {
    let capacity = sys::string::utf8_len(env, self.to_raw()) + 1;
    let mut buffer: Vec<u8> = Vec::with_capacity(capacity as usize);
    let len = sys::string::data(env, buffer.as_mut_ptr(), capacity, self.to_raw());
    buffer.set_len(len as usize);
    buffer // Just return the Vec
}
```

crates/neon/src/sys/string.rs
kjvalencik: It was likely a carry over from the original C++/NAN backend and matched the types defined there. We were trying to allow swapping the backend without changing the Neon API. However, that backend has been removed leading up to 1.0 and it would be preferred to fix this.

Thanks for taking that on! If you prefer to hold off, we can file a follow-up.

crates/neon/src/sys/bindings/functions.rs
kjvalencik: Looking around, it seems like this is always a `char16_t` and Node itself doesn't support a lot of esoteric CPU architectures, so I think this is fine.

Do you mind adding a comment here as a hint in case it *does* cause a problem at some point?

crates/neon/src/types_impl/mod.rs
kjvalencik: Awesome, thanks! Does this mean these functions will fail if the string includes a `NULL` or do you just end up with a null terminated string that isn't actually compatible a with a C-style string (not a problem for us)?

kjvalencik: Reading the code it will *not* error if the string contains a `null`. The string is written without it and then a `null` is written at the end.

https://github.com/nodejs/node/blob/09ae62b9a869ff19d634b0fd1f5a17f198cd1ae7/src/js_native_api_v8.cc#L2310

dherman:
> @dherman can you take a look at the docs in this since you are actively working on them in #942?

These look nice! I'll just stay out of the `JsString` docs in that PR for now, and then afterwards I can do a followup PR to touch up any stylistic inconsistencies.

dnaka91:
Sorry for the delay. Updated the PR with the requested changes, and I resolved the discussions accordingly. Feel free to re-open them if I missed anything.

dnaka91:
Is there anything to be done from my side, or is this just missing a final review?

Would be great to have these extra APIs available soon-ish, but I might just point cargo to my PR branch in the meantime then 👍

kjvalencik:
Should be just the final review. I should have a chance this week. I haven't had a chance with Thanksgiving last week.

kjvalencik (approved):
This looks really great! Thanks for this addition and the fantastic docs, examples and tests!

PR #952, opened 2023-01-03
fix(neon): Close scopes in release builds

kjvalencik:


dherman (approved):
That is so subtle! What a great catch.

:shipit:

PR #963, opened 2023-02-06
JsBigInt

kjvalencik:


crates/neon/src/types_impl/bigint.rs
```diff
@@ -0,0 +1,377 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+/// Indicates a lossless conversion from a [`JsBigInt`] to a Rust integer
+/// could not be performed.
+///
+/// Failures include:
+/// * Negative sign on an unsigned int
+/// * Overflow of an int
+/// * Underflow of a signed int
+pub struct Error<T>(T);
```
dherman, line 29: Why don't we call this `RangeError` to be more precise. For better or worse (both, really) we've avoided the Rust convention of overloading the name `Error` in Neon.

crates/neon/src/types_impl/bigint.rs
```diff
@@ -0,0 +1,377 @@
...
+                sys::Status::Ok,
+            );
+
+            v.assume_init()
+        };
+
+        Handle::new_internal(Self(v))
+    }
+
+    /// Creates a `BigInt` from a [`u64`].
```
dherman, line 91: I'd probably mildly prefer we have examples for every entry point, but if you think it's just noise for the reader of the docs I'm persuadable.

crates/neon/src/types_impl/bigint.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    /// Creates a `BigInt` from a [`u128`].
+    pub fn from_u128<'cx, C>(cx: &mut C, n: u128) -> Handle<'cx, Self>
+    where
+        C: Context<'cx>,
+    {
+        Self::from_u128_sign(cx, Self::POSITIVE, n)
+    }
+
+    /// Creates a `BigInt` from a signed magnitude (i.e., _not_ two's complement).
```
dherman, line 143: This needs some more explanation of what endianness means precisely here, especially what happens if you're building on a big-endian device.

test/napi/src/js/bigint.rs
```diff
@@ -0,0 +1,514 @@
+/// Tests for [`JsBigInt`]. All unit tests are prefixed with `test_` and exported by
```
dherman, line 1: omg so much tests 🤩

dherman (changes requested):
I love the API and the code is super clean and straightforward. And my god, it's full of tests…

I left one suggestion for a name change (`Error` -> `RangeError`) and some docs suggestions.

dherman (approved):
Looks great! I'd just add the endianness formula to `read_digits_le` as well.

PR #970, opened 2023-02-27
feat(neon): sys feature to give access to Node-API

kjvalencik:
There have been [multiple](https://github.com/neon-bindings/neon/pull/801) [examples](https://github.com/neon-bindings/neon/pull/969) where users would like to do something that extends Neon, but requires direct access to Node-API. Sometimes these features to not align with Neon's philosophy.

As an escape hatch, we provide an unsafe `sys` feature that exposes Node-API and conversions to and from Neon types.

### DIFF
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -147,7 +147,7 @@ pub use crate::types::buffer::lock::Lock;
 
 use crate::{
     event::TaskBuilder,
-    handle::{Handle, Managed},
+    handle::Handle,
     object::Object,
     result::{JsResult, NeonResult, Throw},
     sys::{
@@ -157,6 +157,7 @@ use crate::{
     types::{
         boxed::{Finalize, JsBox},
         error::JsError,
+        private::ValueInternal,
         Deferred, JsArray, JsArrayBuffer, JsBoolean, JsBuffer, JsFunction, JsNull, JsNumber,
         JsObject, JsPromise, JsString, JsUndefined, JsValue, StringResult, Value,
     },
@@ -276,9 +277,11 @@ pub trait Context<'a>: ContextInternal<'a> {
             phantom_inner: PhantomData,
         };
 
-        let escapee = unsafe { scope.escape(f(cx)?.to_raw()) };
+        let escapee = unsafe { scope.escape(f(cx)?.to_local()) };
 
-        Ok(Handle::new_internal(V::from_raw(self.env(), escapee)))
+        Ok(Handle::new_internal(unsafe {
+            V::from_local(self.env(), escapee)
+        }))
     }
 
     #[cfg_attr(
@@ -366,9 +369,9 @@ pub trait Context<'a>: ContextInternal<'a> {
     /// Throws a JS value.
     fn throw<T: Value, U>(&mut self, v: Handle<T>) -> NeonResult<U> {
         unsafe {
-            sys::error::throw(self.env().to_raw(), v.to_raw());
+            sys::error::throw(self.env().to_raw(), v.to_local());
+            Err(Throw::new())
         }
-        Err(Throw::new())
     }
 
     /// Creates a direct instance of the [`Error`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Error) class.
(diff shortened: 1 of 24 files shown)


crates/neon/src/context/mod.rs
```diff
@@ -726,3 +736,40 @@ impl<'a> ContextInternal<'a> for FinalizeContext<'a> {
...
 impl<'a> Context<'a> for FinalizeContext<'a> {}
+
+#[cfg(feature = "sys")]
+#[cfg_attr(docsrs, doc(cfg(feature = "sys")))]
+pub struct SysContext<'cx> {
+    env: Env,
+    _phantom_inner: PhantomData<&'cx ()>,
+}
+
+#[cfg(feature = "sys")]
```
dherman, line 747:
```suggestion
#[cfg(feature = "sys")]
/// An execution context constructed from a raw [`Env`](crate::sys::bindings::Env).
```

crates/neon/src/result/mod.rs
```diff
@@ -52,7 +52,21 @@ use crate::{context::Context, handle::Handle, types::Value};
...
 
 impl Throw {
-    pub(crate) fn new() -> Self {
+    #[cfg(feature = "sys")]
+    /// Creates a `Throw` struct representing a JavaScript exception
+    /// state.
+    ///
+    /// # Safety
+    ///
+    /// `Throw` should *only* be constructed when the JavaScript VM is in a
```
dherman, line 61: I really like how the safety requirements can be more concretely explained now in terms of the `sys` crate.

crates/neon/src/sys/bindings/functions.rs
```diff
@@ -209,7 +210,7 @@ mod napi1 {
 
             fn strict_equals(env: Env, lhs: Value, rhs: Value, result: *mut bool) -> Status;
 
-            #[cfg(feature = "external-buffers")]
+            #[cfg(any(feature = "sys", feature = "external-buffers"))]
```
dherman, line 213: Why is this `any` and not `all`?

crates/neon/src/sys/mod.rs
```diff
@@ -1,38 +1,106 @@
...
+//!     SysContext::from_raw(env)
+//! };
+//!
+//! let raw_string: bindings::Value = cx.string("Hello, World!").to_raw();
+//! let js_string = unsafe { JsString::from_raw(&cx, raw_string) };
+//! # }
+//! ```
+use std::{mem::MaybeUninit, sync::Once};
 
+// XXX: Minimize changes due to structure changes
```
dherman, line 70: Is `XXX` a TODO? IMO it's better hygiene to file TODOs as followup issues rather than leaving them in the code.

crates/neon/src/types_impl/buffer/mod.rs
```diff
@@ -232,7 +232,11 @@ pub struct Region<'cx, T: Binary> {
     phantom: PhantomData<T>,
 }
 
-impl<'cx, T: Binary> Region<'cx, T> {
+impl<'cx, T> Region<'cx, T>
+where
+    T: Binary,
+    JsTypedArray<T>: Value,
```
dherman, line 238: Just curious, why did this end up needing this extra `JsTypedArray<T>: Value` constraint but didn't need it before?

crates/neon/src/types_impl/mod.rs
```diff
@@ -55,14 +56,15 @@ pub use self::date::{DateError, DateErrorKind, JsDate};
 #[cfg_attr(docsrs, doc(cfg(all(feature = "napi-5", feature = "futures"))))]
 pub use self::promise::JsFuture;
 
-pub(crate) fn build<'a, T: Managed, F: FnOnce(&mut raw::Local) -> bool>(
+// TODO: Remove
```
dherman, line 59: Hygiene again, can we turn the TODO into a followup issue? You can leave a comment still but link it to the issue.

dherman (approved):
This is amazing. Just a few small suggestions and a couple clarifying questions. I don't think you need a re-review after making any tweaks.

crates/neon/src/sys/bindings/functions.rs
kjvalencik: The symbol will always exist in versions of Node-API that support it. Instead, it fails at runtime.

Since this is the `sys` crate, I wanted to make all symbols available, regardless of whether Neon uses them and hence the `any`.

Do you think we should require users to *also* enable the `external-buffers` feature for access to this `sys` API?

crates/neon/src/types_impl/buffer/mod.rs
kjvalencik: Removal of the `Managed` trait meant that in a couple of places (here included), we no longer had a blanket impl of `Value` for all `T`. Same problem here: https://github.com/neon-bindings/neon/pull/970/files#diff-62376b9ae514f1c89cc0387cf246aec628cea758623a00d3b41d4e4381828791R648

crates/neon/src/sys/bindings/functions.rs
kjvalencik: Resolving this since we determined this to be correct. We don't want users to opt-in a 2nd time since this API is already unsafe.

Issue #966, opened 2023-02-17
Port cargo-cp-artifact to Rust


dherman:
As a dogfood opportunity, port the `cargo-cp-artifact` tool to Rust using Neon. This will also serve as a useful test vehicle for creating prebuild functionality.

PR #971, opened 2023-02-28
Self-hosted `cargo-cp-artifact`

dherman:
Implements #966. Implementation notes:

- Core functionality implemented in `crates/cargo-cp-artifact` as a regular crate
- `crates/cargo-cp-artifact` also has a bin so it can be used to build the self-hosted JS package
- `pkgs/cargo-cp-artifact` has a Neon implementation that depends on `crates/cargo-cp-artifact`
- `pkgs/cargo-cp-artifact/package.json` uses `cargo run -p cargo-cp-artifact` to build itself from the monorepo


pkgs/cargo-cp-artifact/package.json
```diff
@@ -1,17 +1,21 @@
...
     "bin",
-    "src"
+    "index.node"
   ],
   "bin": {
     "cargo-cp-artifact": "bin/cargo-cp-artifact.js"
   },
   "scripts": {
-    "test": "mocha test"
+    "build": "cargo run -p cargo-cp-artifact -- -ac cargo-cp-artifact-adapter index.node -- cargo build --message-format=json-render-diagnostics",
```
dherman, line 14: This works in tree, but won't work when installed from npm. The next step is now to implement pre-builds.

kjvalencik:
@dherman Before publishing, we'll need to check to see if the Rust port is impacted by https://github.com/neon-bindings/neon/pull/972 as well.

PR #979, opened 2023-03-24
Relax Send constraints

kjvalencik:
Currently, Neon _pessimistically_ assumes that JavaScript runtimes _may_ choose to execute from different threads. The specification only defines serial execution and not necessarily single threaded execution.

In practice, this will always be one thread. V8 makes heavy usage of TLS and it would be a large lift to start doing things differently.

This PR changes Neon to _optimistically_ assume the runtime will always execute from the same thread. As a small level of protection, I added `SendWrapper` when `debug_assertions` are enabled. This will `panic` if the single-thread invariant is violated.

### DIFF
--- a/crates/neon/Cargo.toml
+++ b/crates/neon/Cargo.toml
@@ -31,6 +31,7 @@ neon-macros = { version = "=1.0.0-alpha.2", path = "../neon-macros" }
 aquamarine = { version = "0.1.11", optional = true }
 easy-cast = { version = "0.5.1", optional = true }
 doc-comment = { version = "0.3.3", optional = true }
+send_wrapper = "0.6"
 
 [dependencies.tokio]
 version = "1.24.2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -320,6 +320,7 @@ dependencies = [
  "once_cell",
  "psd",
  "semver",
+ "send_wrapper",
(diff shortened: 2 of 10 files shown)


crates/neon/Cargo.toml
```diff
@@ -31,6 +31,7 @@ neon-macros = { version = "=1.0.0-alpha.2", path = "../neon-macros" }
 aquamarine = { version = "0.1.11", optional = true }
 easy-cast = { version = "0.5.1", optional = true }
 doc-comment = { version = "0.3.3", optional = true }
+send_wrapper = "0.6"
```
complexspaces, line 34: Is this dependency unused in the current branch revision? I don't see it being imported and there's a smaller, handrolled implementation of `SendWrapper` instead in the `neon` crate.

crates/neon/src/sys/send_wrapper.rs
dherman, line 1: Naming this the same name as the third-party crate is slightly confusing, but maybe it's better than inventing another name. `debug_send_wrapper`? I don't feel too strongly about it.

dherman (approved):
LGTM! I found it slightly confusing that the debug-only version of send_wrapper is also called send_wrapper, but I don't feel too strongly about it.

PR #989, opened 2023-05-28
remove `JsFunction` type parameter

dherman:
From the list of [remaining breaking changes](https://github.com/neon-bindings/neon/wiki/Candidate-Breaking-Changes-for-1.0), this PR removes the `JsFunction` type parameter. This is safe for the following reasons:

- Neon functions are non-strict, which means when they are invoked with a non-object (e.g. with `f.call(null)` or `f.call(17)`) the primitive is replaced by an object.
- The `JsObject` type does not assume the object inherits from `Object.prototype`, so it works for any JS objects, even exotic objects like `Object.create(null)`, proxies, module objects, etc.

### DIFF
--- a/crates/neon/src/types_impl/mod.rs
+++ b/crates/neon/src/types_impl/mod.rs
@@ -16,7 +16,6 @@ pub(crate) mod utf8;
 
 use std::{
     fmt::{self, Debug},
-    marker::PhantomData,
     os::raw::c_void,
 };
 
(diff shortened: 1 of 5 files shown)


dherman:
I added a unit test that uses the global object, so it'll have a small merge conflict with #987.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1173,7 +1170,7 @@ impl<CL: Object> JsFunction<CL> {
     /// Calls this function as a constructor.
     ///
     /// **See also:** [`JsFunction::construct_with`].
-    pub fn construct<'a, 'b, C: Context<'a>, AS>(&self, cx: &mut C, args: AS) -> JsResult<'a, CL>
+    pub fn construct<'a, 'b, C: Context<'a>, AS>(&self, cx: &mut C, args: AS) -> JsResult<'a, JsObject>
```
kjvalencik, line 1173: Goodbye soundness hole!

_Technically it wasn't a soundness hole because Node-API **also** checks the type, but at least it won't unexpectedly crash._

kjvalencik:
I love the tests checking that constructions are still objects! ❤️

PR #991, opened 2023-06-05
Draft: stash

kjvalencik:
Trying out some different ways of exporting functions.

**WARNING**: This is _unsound_. There should be a HRTB when exporting. Currently this allows exporting a function that has a `Context<'static>`. However, adding the HRTB causes the `Function` trait to hold parameters with lifetimes and the code no longer compiles. I haven't figured this out.


arqgen-otavio (approved):

kjvalencik:
Closing since this is very unsafe and it's not clear if it's possible to write this in a safe API that's still ergonomic.

It may be better to focus on `cx` helpers--that already have a clear path--and make it more ergonomic with a proc macro.

PR #995, opened 2023-08-03
Root<T>: Data race allowed on T

kuzeyardabulut:
Hi,
I found a memory-safety/soundness issue in this crate while scanning Rust code for potential vulnerabilities. This PR contains a fix for the issue.

# Issue Description
`Root<T>` unconditionally implements Sync. This allows users to create data races on `T: !Sync`. Such data races can lead to undefined behavior.
https://github.com/neon-bindings/neon/blob/4c2e455a9e6814f1ba0178616d63caec7f4df317/crates/neon/src/handle/root.rs#L71-L73

### DIFF
--- a/crates/neon/src/handle/root.rs
+++ b/crates/neon/src/handle/root.rs
@@ -68,9 +68,9 @@ impl<T> std::fmt::Debug for Root<T> {
 // `Root` are intended to be `Send` and `Sync`
 // Safety: `Root` contains two types. A `NapiRef` which is `Send` and `Sync` and a
 // `PhantomData` that does not impact the safety.
-unsafe impl<T> Send for Root<T> {}
+unsafe impl<T: Send> Send for Root<T> {}
 
-unsafe impl<T> Sync for Root<T> {}
+unsafe impl<T: Sync> Sync for Root<T> {}
 
 #[cfg(feature = "napi-6")]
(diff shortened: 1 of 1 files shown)


kjvalencik:
Can you give an example of such a data race? The only thing that `Root` can *ever* hold is a [`napi_ref`](https://nodejs.org/api/n-api.html#napi_ref), which *can* be sent across threads.

`Root` can't hold any `T`, it's specifically a wrapper to allow safely holding references to JS values across threads. They can only ever be dereferenced on the JS main thread.

kuzeyardabulut:
I agree with what you said. This may not cause a direct data race. But making the above changes is useful if it doesn't break the flow of the code.

kjvalencik:
> But making the above changes is useful if it doesn't break the flow of the code.

It does because `T` is the type that will be pulled *from* the `Root` on the main thread and it is explicitly *not* `Sync`.

kjvalencik:
Closing the loop on this, after discussing and examining with a few folks:
* This is sound as `Root` may only be constructed with `T: Object` and does not contain any references to `T`
* Requiring `T: Send` and `T: Sync` is not possible because the intent of this type is a safe Send/Sync wrapper around normally !Send + !Sync references to JS values on the V8 heap.
* We *could* add a bound on `struct Root` to require `T: Value`, but this does not increase external safety (already protected by private fields) and would propagate `T` bounds unnecessarily.

PR #997, opened 2023-08-18
feat(neon): Extractors

kjvalencik:
Extract Rust data from JavaScript values and arguments.

```rust
fn add(mut cx: FunctionContext) -> JsResult<JsNumber> {
    let (a, b): (f64, f64) = cx.args()?;

    Ok(cx.number(a + b))
}
```


dherman:
Since you have to return `NeonResult` for full generality, I do agree that nested `Result`s for the overloading case would be awkward. But I wonder if we should consider a `Result<Option>` version of `args` called something like `match_args` or `args_opt`. That would allow you to do the overloading example like this:

```rust
fn combine(mut cx: FunctionContext) -> JsResult<JsValue> {
    if let Some((a, b)) = cx.args_opt()? {
        return Ok(add(cx, a, b).upcast());
    }

    let (a, b) = cx.args()?;

    Ok(concat(cx, a, b).upcast())
}
```

This is slightly more readable IMO, but it also means you're not accidentally suppressing actual JavaScript exceptions that might occur, so it's more correct.

kjvalencik:
@dherman I'm having a difficult time picturing how that ladders up from the `TryFromJs` trait. Would we have a specialized error that means "wrong type" that we could use for this? Something like a combination of `DowncastResult` and `Throw`?

kjvalencik:
```rust
pub trait TryFromJs<'cx>: Sized {
    type Error: Error + Send + Sync + 'static;

    fn try_from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Result<Self, Self::Error>>
        where C: Context<'cx>;

    fn from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Self>
        where C: Context<'cx>
    {
        Self::try_from_js(cx, v)?.or_throw(cx)
    }
}
```

PR #1012, opened 2023-12-02
Update README.md

xyloflake:
Fix for a small grammar mistake. 
(Don't kill me for the useless change lol)


kjvalencik:
Thanks for the fix!

PR #1013, opened 2023-12-08
create-neon 0.3: Create Neon 1.0 project with default feature flags

kjvalencik:


pkgs/create-neon/data/versions.json
```diff
@@ -1,5 +1,4 @@
 {
-  "neon": "1.0.0",
-  "napi": "8",
+  "neon": "1",
```
kjvalencik, line 2: This felt cleaner because `1.0.0` and `1` mean the same thing with any additional specifiers.

dherman (approved):
:shipit:

PR #1023, opened 2024-03-22
Try out an export macro

kjvalencik:


crates/neon/src/context/internal.rs
```diff
@@ -66,7 +66,13 @@ pub unsafe fn initialize_module(
     let env = Env(env);
     let exports = Handle::new_internal(JsObject::from_local(env, exports.cast()));
 
-    ModuleContext::with(env, exports, |cx| {
+    ModuleContext::with(env, exports, |mut cx| {
+        for create in crate::macro_internal::EXPORTS {
```
kjvalencik, line 70: This shouldn't be done implicitly. Instead, it should be done explicitly in the `#[neon::main]` function.

```rust
#[neon::main]
    neon::exports().register(&mut cx)?;
    Ok(())
}
```

The downside of this approach is that it's a bit more boilerplate and requires manual migration from users. The upside is that it makes the empty main serve an obvious purpose and also allows users to customize exports.

```rust
#[neon::main]
    for (k, create) in neon::exports() {
        let v = create(&mut cx)?;
        
        cx.export_value(k.to_uppercase(), v)?;
    }

    Ok(())
}
```

kjvalencik: An interesting thought is that we could potentially have it _both ways_ as long as we accept only a *single* version* of Neon can be in an addon module tree.

The rough idea:
* `#[neon::main]` is changed to work similar to `#[neon::export]`. Instead of generating `napi_register_module_v1`, it registers the function in a doc hidden global exposed by `neon`
* `neon` exports a global `napi_register_module_v1` symbol (*fun fact*: rust functions can be private and still exported as a global symbol)
* The neon's `napi_register_module_v1`calls each of the registered `main` functions
* A special case is added for *no* registered main that automatically calls `neon::exports()`

This has a nice side-effect of us no longer generating `unsafe` code in our user's crates. It's backwards compatible, even though users have existing `main` that don't export, because those users *also* won't have any `#[neon::export]`. We will need to add to the migration guide manually adding the line.

*This is _probably_ okay because I can't think of a reason that would be useful.

PR #1024, opened 2024-03-26
feat(neon): Extractors

kjvalencik:
Extractors allow quickly getting JavaScript arguments as Rust values. In addition to immediate ergonomics and performance improvements, extractors set a foundation for future proc macros.

The Neon API surface area increases by two traits:
* `TryFromJs`
* `FromArgs`

## `TryFromJs`

The `TryFromJs` trait provides a standardized way of extracting a `Rust` value from a `JsValue`. It is initially `Sealed` while we determine if changes are needed. In a future release, we will likely make it available for users to implement.

## `FromArgs`

The `FromArgs` trait is an implementation detail and will *remain* `Sealed`. It is implemented by a macro for tuples up to size 32.

## Extractors

`TryFromJs` is implemented for many standard library values (`f64`, `String`) and wrapper types (`Option`). It also has blanket implementations for `Handle` and `NeonResult`. Additionally, I introduce the concept of an "extractor". The idea is inspired by [`axum`](https://docs.rs/axum/0.7.5/axum/extract/index.html). These extractors allow for specialization where extracted types would otherwise overlap. For example, getting an `f64` from a `Date`.

Most extractors are implemented using the `sys` crate directly in order to avoid duplicate type checks. In the future, we will likely want to convert existing functions to use the `TryFromJs` implementations.

## `Json`

As found previously, serialization to JSON is nearly *always* faster than manually reading JavaScript objects. Working with JSON, thanks to `serde`, is also much easier. Community crates like `neon-serde` have demonstrated a high demand for this functionality.

The `Json` extractor leverages `serde` and `serde_json` to convert to a Rust type. The ✨ magic ✨ is that it doesn't require the value to already be stringified on the JavaScript side. It leverage's Neon's new [`LocalKey`](https://docs.rs/neon/latest/neon/thread/struct.LocalKey.html) functionality for fast access to `JSON.stringify`. `JSON.stringify` and `serde_json::from_str` are implicitly called when extracting.

### DIFF
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -157,6 +157,7 @@ use crate::{
     types::{
         boxed::{Finalize, JsBox},
         error::JsError,
+        extract::FromArgs,
         private::ValueInternal,
         Deferred, JsArray, JsArrayBuffer, JsBoolean, JsBuffer, JsFunction, JsNull, JsNumber,
         JsObject, JsPromise, JsString, JsUndefined, JsValue, StringResult, Value,
@@ -211,6 +212,36 @@ impl CallbackInfo<'_> {
             sys::call::this(env.to_raw(), self.info, &mut local);
             local
         }
+    }
+
+    pub(crate) fn argv_exact<'b, C: Context<'b>, const N: usize>(
+        &self,
+        cx: &mut C,
+    ) -> [Handle<'b, JsValue>; N] {
+        use std::ptr;
+
+        let mut argv = [JsValue::new_internal(ptr::null_mut()); N];
+        let mut argc = argv.len();
+
+        // # Safety
+        // * Node-API fills empty slots with `undefined
+        // * `Handle` and `JsValue` are transparent wrappers around a raw pointer
+        unsafe {
+            assert_eq!(
+                sys::get_cb_info(
+                    cx.env().to_raw(),
+                    self.info,
+                    &mut argc,
+                    argv.as_mut_ptr().cast(),
+                    ptr::null_mut(),
+                    ptr::null_mut(),
+                ),
+                sys::Status::Ok,
+            );
+        }
+
+        // Empty values will be filled with `undefined`
+        argv
     }
 }
 
@@ -682,6 +713,55 @@ impl<'a> FunctionContext<'a> {
     /// Produces a handle to the function's [`this`-binding](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#function_context).
     pub fn this_value(&mut self) -> Handle<'a, JsValue> {
         JsValue::new_internal(self.info.this(self))
+    }
+
+    /// Extract Rust data from the JavaScript arguments.
+    ///
+    /// This is frequently more efficient and ergonomic than getting arguments
+    /// individually. See the [`extract`](crate::types::extract) module documentation
+    /// for more examples.
+    ///
+    /// ```
+    /// # use neon::{prelude::*, types::extract::*};
+    /// fn add(mut cx: FunctionContext) -> JsResult<JsNumber> {
+    ///     let (a, b): (f64, f64) = cx.args()?;
+    ///
+    ///     Ok(cx.number(a + b))
+    /// }
+    /// ```
+    pub fn args<T>(&mut self) -> NeonResult<T>
+    where
+        T: FromArgs<'a>,
+    {
+        T::from_args(self)
+    }
+
+    /// Extract Rust data from the JavaScript arguments.
+    ///
+    /// Similar to [`FunctionContext::args`], but does not throw a JavaScript exception on errors. Useful
+    /// for function overloading.
+    ///
+    /// ```
+    /// # use neon::{prelude::*, types::extract::*};
+    /// fn combine(mut cx: FunctionContext) -> JsResult<JsValue> {
+    ///     if let Some((a, b)) = cx.args_opt::<(f64, f64)>()? {
+    ///         return Ok(cx.number(a + b).upcast());
+    ///     }
+    ///
+    ///     let (a, b): (String, String) = cx.args()?;
+    ///
+    ///     Ok(cx.string(a + &b).upcast())
+    /// }
+    /// ```
+    pub fn args_opt<T>(&mut self) -> NeonResult<Option<T>>
+    where
+        T: FromArgs<'a>,
+    {
+        T::from_args_opt(self)
+    }
+
+    pub(crate) fn argv<const N: usize>(&mut self) -> [Handle<'a, JsValue>; N] {
+        self.info.argv_exact(self)
     }
 }
 
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -224,7 +224,7 @@ impl CallbackInfo<'_> {
         let mut argc = argv.len();
 
         // # Safety
-        // * Node-API fills empty slots with `undefined
+        // * Node-API fills empty slots with `undefined`
         // * `Handle` and `JsValue` are transparent wrappers around a raw pointer
         unsafe {
             assert_eq!(
--- a/crates/neon/src/lib.rs
+++ b/crates/neon/src/lib.rs
@@ -78,8 +78,6 @@
 //! [supported]: https://github.com/neon-bindings/neon#platform-support
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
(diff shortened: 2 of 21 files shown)


crates/neon/src/context/mod.rs
```diff
@@ -683,6 +714,55 @@ impl<'a> FunctionContext<'a> {
...
+    /// }
+    /// ```
+    pub fn args_opt<T>(&mut self) -> NeonResult<Option<T>>
+    where
+        T: FromArgs<'a>,
+    {
+        T::from_args_opt(self)
+    }
+
+    pub(crate) fn argv<const N: usize>(&mut self) -> [Handle<'a, JsValue>; N] {
```
dherman, line 763: Does this impact our MSRV?

crates/neon/src/lib.rs
```diff
@@ -78,6 +78,8 @@
 //! [supported]: https://github.com/neon-bindings/neon#platform-support
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+extern crate core;
```
dherman, line 81: What's this for?

crates/neon/src/types_impl/extract/json.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::{handle::Root, thread::LocalKey};
+
+fn global_json_stringify<'cx, C>(cx: &mut C) -> JsResult<'cx, JsFunction>
+where
+    C: Context<'cx>,
+{
+    cx.global::<JsObject>("JSON")?.get(cx, "stringify")
+}
+
+#[cfg(not(feature = "napi-6"))]
```
dherman, line 22: Pedantically, this isn't exactly equivalent behaviorally since someone could modify `JSON.stringify`, but I don't think it matters.

crates/neon/src/types_impl/extract/json.rs
```diff
@@ -0,0 +1,76 @@
...
+/// Extract a value by serializing to JSON
+pub struct Json<T>(pub T);
+
+impl<'cx, T> TryFromJs<'cx> for Json<T>
+where
+    for<'de> T: serde::de::Deserialize<'de>,
+{
+    type Error = serde_json::Error;
+
+    fn try_from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Result<Self, Self::Error>>
```
dherman, line 61: This is so elegant.

crates/neon/src/types_impl/extract/mod.rs
```diff
@@ -0,0 +1,229 @@
+//! Traits and utilities for extract Rust data from JavaScript values
```
dherman, line 1:
```suggestion
//! Traits and utilities for extract Rust data from JavaScript values.
```

crates/neon/src/context/mod.rs
```diff
@@ -212,6 +213,36 @@ impl CallbackInfo<'_> {
...
+        &self,
+        cx: &mut C,
+    ) -> [Handle<'b, JsValue>; N] {
+        use std::ptr;
+
+        let mut argv = [JsValue::new_internal(ptr::null_mut()); N];
+        let mut argc = argv.len();
+
+        // # Safety
+        // * Node-API fills empty slots with `undefined
```
dherman, line 227:
```suggestion
        // * Node-API fills empty slots with `undefined`
```

crates/neon/src/types_impl/extract/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+where
+    Self: private::Sealed + Sized,
+{
+    type Error;
+
+    fn try_from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Result<Self, Self::Error>>
+    where
+        C: Context<'cx>;
+
+    fn from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Self>
```
dherman, line 129: Do we not want a default implementation that calls `self.try_from_js()` and throws an exception on `Ok(Err())`?

crates/neon/src/types_impl/extract/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    ($(#[$attrs:meta])? [$($head:ident),*], []) => {};
+
+    ($(#[$attrs:meta])? [$($head:ident),*], [$cur:ident $(, $tail:ident)*]) => {
+        from_args_impl!($(#[$attrs])? [$($head,)* $cur]);
+        from_args_expand!($(#[$attrs])? [$($head,)* $cur], [$($tail),*]);
+    };
+}
+
+macro_rules! from_args {
+    ([$($show:ident),*], [$($hide:ident),*]) => {
```
dherman, line 205: A comment here would be useful to explain that the first list is the ones with docs and the second list is `doc(hidden)`.

crates/neon/src/types_impl/extract/types.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    fn try_from_js<C>(cx: &mut C, v: Handle<'cx, JsValue>) -> NeonResult<Result<Self, Self::Error>>
+    where
+        C: Context<'cx>,
+    {
+        let mut n = 0f64;
+
+        unsafe {
+            match sys::get_value_double(cx.env().to_raw(), v.to_local(), &mut n) {
+                sys::Status::NumberExpected => return Ok(Err(TypeExpected::new())),
```
dherman, line 122: I think we've typically kept this kind of logic out of Neon proper and relegated to the runtime crate, but maybe I'm mis-remembering or you want to intentionally drop that convention. (Come to think of it, I don't know if that has a performance cost due to compiler limitations for cross-crate inlining.)

crates/neon/src/types_impl/extract/types.rs
```diff
@@ -0,0 +1,356 @@
...
+
+        Ok(Ok(Date(d)))
+    }
+
+    from_js!();
+}
+
+impl private::Sealed for Date {}
+
+impl<'cx> TryFromJs<'cx> for () {
```
dherman, line 240: What's the rationale for `()` to correspond to any JS value vs just `undefined` or `null | undefined`? I don't have a strong intuition here.

test/napi/src/js/extract.rs
```diff
@@ -0,0 +1,120 @@
+use neon::{prelude::*, types::extract::*};
+
+pub fn extract_values(mut cx: FunctionContext) -> JsResult<JsArray> {
```
dherman, line 3: Nice job testing a wide range so concisely!

dherman (changes requested):
Mostly a bunch of questions, so maybe it won't require any changes in the end, but I'm marking it as "request changes" just to talk through the comments.

crates/neon/src/types_impl/extract/mod.rs
kjvalencik: Currently, there cannot be a [useful] default implementation because there are no trait bounds on `type Error`. Currently the `TryFromJs` trait is sealed and cannot be implemented by users.

I opened an issue to stabilize `TryFromJs` and added this question. https://github.com/neon-bindings/neon/issues/1026

crates/neon/src/types_impl/extract/types.rs
kjvalencik: I added a comment explaining the rationale and also created a follow-up issue to clean-up module orgnaization.

https://github.com/neon-bindings/neon/issues/1027

crates/neon/src/lib.rs
kjvalencik: Oops. Junk added by my IDE that I missed. Removed.

crates/neon/src/types_impl/extract/json.rs
kjvalencik: ❤️ Thanks!

crates/neon/src/types_impl/extract/types.rs
kjvalencik: I added a detailed comment, but the quick version is that there's a blanket impl of `FromArgs` for `T` where `T: TryFromJs`. This makes single argument functions ergonomic and removes the "`(T)` isn't a tuple, but `(T,)`" confusion.

However, it introduces an ambiguity of whether `let () = cx.args()?` is extracting unit from the first argument or is extracting zero arguments. Infallibly extracting unit removes all impact from the ambiguity.

dherman (approved):
LGTM -- this one is super exciting, ship it!

PR #1039, opened 2024-05-08
fix(cargo-cp-artifact): Hotfix for change in cargo diagnostics naming

kjvalencik:
Fix for a change in `cargo` behavior to use the `snake_case` name in artifacts.

https://github.com/rust-lang/cargo/issues/13867

### DIFF
--- a/pkgs/cargo-cp-artifact/src/index.js
+++ b/pkgs/cargo-cp-artifact/src/index.js
@@ -80,8 +80,8 @@ function processCargoBuildLine(options, copied, line) {
   // `kind` and `filenames` zip up as key/value pairs
   kinds.forEach((kind, i) => {
     const filename = filenames[i];
-    const key = getArtifactName({ artifactType: kind, crateName: name });
-    const outputFiles = options.artifacts[key];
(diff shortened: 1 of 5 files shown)


pkgs/cargo-cp-artifact/src/index.js
```diff
@@ -100,6 +100,22 @@ function processCargoBuildLine(options, copied, line) {
...
 
+function getOutputFiles(kind, name, artifacts) {
+  const key = getArtifactName({ artifactType: kind, crateName: name });
+  const outputFiles = artifacts[key];
+
+  if (outputFiles) {
+    return { key, outputFiles };
+  }
+
+  const altKey = key.replace(/_/g, "-");
```
dherman, line 111:
```suggestion
  // Check for a snake_case version of the name, which recent versions of
  // cargo have begun transforming to:
  //     https://github.com/rust-lang/cargo/issues/13867
  const altKey = key.replace(/_/g, "-");
```

dherman (approved):
Looks good -- I made a couple presentational suggestions.

PR #1041, opened 2024-05-12
fix(create-neon): Explicit project type

dherman:
This PR changes the syntax of create-neon to require the user to explicitly specify the Neon project type, either by passing `--lib` or `--app` at the command-line, or otherwise by interactively asking them via stdin.

The reason for this is a limitation of `npm init`: the only way it will forward an argument like `--lib` to create-neon is if the user remembers to prefix it with `--`:

```
npm init neon -- --lib my-project
```

Without the extra `--`, npm silently ignores the `--lib`, which is a subtle and surprising hazard. This pitfall is easy to stumble into and difficult to debug.

(I built a demonstration package called [create-dherman](https://www.npmjs.com/package/create-dherman) that you can use to test this behavior. Test it out with `npm init dherman --lib foobar` vs `npm init dherman -- --lib foobar`, for example.)

With this PR, users still need the `--`, but if they forget it, they'll be asked the project type interactively.

### DIFF
--- a/pkgs/create-neon/data/templates/manifest/scripts.json.hbs
+++ b/pkgs/create-neon/data/templates/manifest/scripts.json.hbs
@@ -6,8 +6,7 @@
   "postcross-build": "neon dist -m /target < cross.log",
   "debug": "npm run cargo-build --",
   "build": "npm run cargo-build -- --release",
-  "cross": "npm run cross-build -- --release",{{#if packageSpec.library}}
+  "cross": "npm run cross-build -- --release"{{#if packageSpec.library}},
   "prepack": "{{#eq packageSpec.library.lang compare="ts"}}tsc && {{/eq}}neon update",
-  "version": "neon bump --binaries platforms && git add ."{{else}}
-  "prepare": "npm run build"{{/if}}
+  "version": "neon bump --binaries platforms && git add ."{{/if}}
 }
--- a/pkgs/create-neon/data/templates/manifest/scripts.json.hbs
+++ b/pkgs/create-neon/data/templates/manifest/scripts.json.hbs
@@ -6,7 +6,8 @@
   "postcross-build": "neon dist -m /target < cross.log",
   "debug": "npm run cargo-build --",
   "build": "npm run cargo-build -- --release",
-  "cross": "npm run cross-build -- --release"{{#if packageSpec.library}},
+  "cross": "npm run cross-build -- --release",{{#if packageSpec.library}}
   "prepack": "{{#eq packageSpec.library.lang compare="ts"}}tsc && {{/eq}}neon update",
-  "version": "neon bump --binaries platforms && git add ."{{/if}}
+  "version": "neon bump --binaries platforms && git add ."{{else}}
+  "prepare": "npm run build"{{/if}}
 }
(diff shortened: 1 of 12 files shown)


pkgs/create-neon/data/templates/manifest/scripts.json.hbs
```diff
@@ -6,7 +6,8 @@
   "postcross-build": "neon dist -m /target < cross.log",
   "debug": "npm run cargo-build --",
   "build": "npm run cargo-build -- --release",
-  "cross": "npm run cross-build -- --release"{{#if packageSpec.library}},
+  "cross": "npm run cross-build -- --release",{{#if packageSpec.library}}
   "prepack": "{{#eq packageSpec.library.lang compare="ts"}}tsc && {{/eq}}neon update",
-  "version": "neon bump --binaries platforms && git add ."{{/if}}
+  "version": "neon bump --binaries platforms && git add ."{{else}}
+  "prepare": "npm run build"{{/if}}
```
kjvalencik, line 12: Nit, since this is only for an app, it might be more clear to be `install`.

Although, I'm questioning my suggestion that this is a good idea. There's an odd flow where `package.json` has updated dependencies and we're going to re-build Rust just to install more packages.

pkgs/create-neon/src/cache/npm.ts
```diff
@@ -1,11 +1,16 @@
...
   readonly type: string = "npm";
 
-  constructor(org: string | null) {
-    this.org = org;
+  constructor(pkg: string, org?: string) {
+    this.org = org || NPM.inferOrg(pkg);
+  }
+
+  static inferOrg(pkg: string): string {
+    const m = pkg.match(/^@([^/]+)\/([^/]+)/);
```
kjvalencik, line 13: Nit, use `.*` as the second capture group. NPM doesn't allow slashes, so this _should_ be equivalent, but if it breaks, it will break in a bit more predictable way (grabbing the entire section rather than something in the middle).

pkgs/create-neon/src/cache/npm.ts
```diff
@@ -1,11 +1,16 @@
...
+    return "@" + (m ? m[1] : pkg);
```
kjvalencik, line 14: FYI, with fancy JS features (optional chaining and null-ish coalescing).

```js
return `@${m?.[1] ?? pkg}`;
```

kjvalencik (approved):
The templated JSON is a difficult to tell if it always results in valid JSON. In a future PR, it may be cleaner to generate it in JS and call `JSON.stringify`.

pkgs/create-neon/data/templates/manifest/scripts.json.hbs
dherman: That's a good point. Given that people can always add the `"install"` or `"prepare"` hook themselves when configuring their app, I think after reflection we should take this back out.

pkgs/create-neon/src/cache/npm.ts
dherman: ooh much nicer

kjvalencik (approved):

PR #1046, opened 2024-05-19
refactor(cargo-cp-artifact): use `copyArtifact` from `@neon-rs/artifact`

### LINKED ISSUE #1045: cargo-cp-artifact: refactor to use @neon-rs/artifact
Refactor cargo-cp-artifact to import the `copyArtifact` logic from [@neon-rs/artifact](https://www.npmjs.com/package/@neon-rs/artifact).

dherman:
Closes #1045.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -832,6 +832,11 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@neon-rs/artifact": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/@neon-rs/artifact/-/artifact-0.1.0.tgz",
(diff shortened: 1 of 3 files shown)


kjvalencik:
`cargo-cp-artifact` currently has zero dependencies. I'm not sure if adding one for ~20 LoC is worth it.

dherman:
> `cargo-cp-artifact` currently has zero dependencies. I'm not sure if adding one for ~20 LoC is worth it.

It's an internal dependency, though: we maintain all the code. It's just that we use it in two different CLI tools, so it avoids us having to maintain it in two places.

kjvalencik:
Given that we are likely to stop maintaining `cargo-cp-artifact`, my preference is not to couple the two and potentially make the `neon` CLI more difficult to maintain.

PR #1047, opened 2024-05-28
refactor(create-neon): control flow cleanup

dherman:
This PR is a big cleanup to the control flow and logic of create-neon so that it should be much more maintainable:

- much less logic resides in the root
- most of the logic is in a Creator class
- subclassing for `--app` vs `--lib` eliminates a lot of spaghetti conditionals and makes it easier to understand what operations go with what cases


kjvalencik (approved):
This clean-up is a huge improvement! Thanks!

PR #1048, opened 2024-05-30
Replace all generic contexts with a single Cx type

kjvalencik:
This allows users to take `Cx` as an argument instead of being generic over the `Context` trait. For example, this will allow us to implement `TryIntoJs` for closures.

This is technically a breaking change because several context types have been replaced by type aliases; if a user were implementing a trait over multiple of these types, it would now cause a conflict. In practice, users either didn't do this at all or implemented it generically over the `Context` trait.

We **could** keep the existing types, it just makes it harder to deprecate them in the future. For a Neon 2.0, I would want to eliminate the `Context` trait and *only* make it concrete. The special needs of function and module would become additional arguments.

### DIFF
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -180,21 +180,26 @@ use crate::types::date::{DateError, JsDate};
 #[cfg(feature = "napi-6")]
 use crate::lifecycle::InstanceData;
 
+#[doc(hidden)]
 /// An execution context of a task completion callback.
 pub type TaskContext<'cx> = Cx<'cx>;
 
+#[doc(hidden)]
 /// An execution context of a scope created by [`Context::execute_scoped()`](Context::execute_scoped).
 pub type ExecuteContext<'cx> = Cx<'cx>;
 
+#[doc(hidden)]
 /// An execution context of a scope created by [`Context::compute_scoped()`](Context::compute_scoped).
 pub type ComputeContext<'cx> = Cx<'cx>;
 
+#[doc(hidden)]
 /// A view of the JS engine in the context of a finalize method on garbage collection
 pub type FinalizeContext<'cx> = Cx<'cx>;
 
 /// An execution context constructed from a raw [`Env`](crate::sys::bindings::Env).
 #[cfg(feature = "sys")]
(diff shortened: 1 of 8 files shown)


crates/neon/src/context/mod.rs
dherman, line 1: We should add an example of writing helper functions on contexts in the module docs, right before the `Memory Management` section.

crates/neon/src/context/mod.rs
```diff
@@ -551,20 +613,37 @@ pub trait Context<'a>: ContextInternal<'a> {
...
+impl<'cx> Deref for ModuleContext<'cx> {
+    type Target = Cx<'cx>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.cx
+    }
 }
 
-impl<'a> UnwindSafe for ModuleContext<'a> {}
+impl<'cx> DerefMut for ModuleContext<'cx> {
```
dherman, line 629: Does this mean if you already have an `&mut ModuleContext` you'd have to do something like `&mut *cx` to pass it to a function that takes `Cx`?

dherman (changes requested):
I like this. I left a suggestion for a little documentation in the `neon::context` module docs.

The only thing I don't like is the collective baggage of all the different Context types. I wonder if it's worth making some of them `doc(hidden)`? Or if there's some way to structure the `neon::context` docs to make it easier to understand the minimal subset you actually need to learn?

crates/neon/src/context/mod.rs
kjvalencik: Nope! It can be passed directly and deref coercion will make it the right type. This is similar to how you can pass a `&String` to a function that takes a `&str`.

The only times you would need to be more explicit are if it's wrapped in another value (e.g., `Option<ModuleContext>` would need `as_deref_mut()`) or if we don't have a function call to cause it (e.g., a `match` on an enum of different context types).

crates/neon/src/context/mod.rs
kjvalencik: Added an example.

kjvalencik:
@dherman I marked them `#[doc(hidden)]` and removed all references to them internally. In the future, we will likely want to mark them deprecated.

dherman (approved):
Looks great, let's ship it!

PR #1055, opened 2024-07-05
feat(neon): Add tokio async runtime support

kjvalencik:
This PR adds support for exporting `async fn` and future return `fn` in `#[neon::export]`.

```rust
#[neon::export]
async fn example1(a: f64, b: f64) -> f64 {
    a + b
}

#[neon::export(async)]
fn example2(a: f64, b: f64) -> impl Future<Output = f64> {
    async move {
        a + b
    }
}
```

It is intentionally very conservative in semver guarantees. It requires registering a global future executor. The executor must implement the _private_ `Runtime` trait. Currently, only a `tokio` implementation is included. Additionally, `Future` trait bounds are strict enough to allow for multi-threaded executors.

As we see demand, we can add support for additional executors (e.g., `async-std`) and possibly even single threaded executors (`#[neon::export(?Sync)]`) or a built-in `libuv` driven executor. The eventual goal is to make the `Runtime` trait public so that users can bring their own executor.

* [x] ~~Depends on https://github.com/neon-bindings/neon/pull/1057~~ (Merged)

### DIFF
--- a/crates/neon/src/executor/mod.rs
+++ b/crates/neon/src/executor/mod.rs
@@ -1,0 +1,59 @@
+use std::{future::Future, pin::Pin};
+
+use crate::{context::Cx, thread::LocalKey};
+
+#[cfg(feature = "tokio-rt")]
+pub(crate) mod tokio;
+
+type BoxFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+
+pub(crate) static RUNTIME: LocalKey<Box<dyn Runtime>> = LocalKey::new();
+
+pub trait Runtime: Send + Sync + 'static {
+    fn spawn(&self, fut: BoxFuture);
+}
+
+/// Register a [`Future`] executor runtime globally to the addon.
+///
+/// Returns `Ok(())` if a global executor has not been set and `Err(runtime)` if it has.
+///
+/// If the `tokio` feature flag is enabled and the addon does not provide a
+/// [`#[neon::main]`](crate::main) function, a multithreaded tokio runtime will be
+/// automatically registered.
+///
+/// **Note**: Each instance of the addon will have its own runtime. It is recommended
+/// to initialize the async runtime once in a process global and share it across instances.
+///
+/// ```
+/// # #[cfg(feature = "tokio-rt-multi-thread")]
+/// # fn example() {
+/// # use neon::prelude::*;
+/// use once_cell::sync::OnceCell;
+/// use tokio::runtime::Runtime;
+///
+/// static RUNTIME: OnceCell<Runtime> = OnceCell::new();
+///
+/// #[neon::main]
+/// fn main(mut cx: ModuleContext) -> NeonResult<()> {
+///     let runtime = RUNTIME
+///         .get_or_try_init(Runtime::new)
+///         .or_else(|err| cx.throw_error(err.to_string()))?;
+///
+///     let _ = neon::set_global_executor(&mut cx, runtime);
+///
+///     Ok(())
+/// }
+/// # }
+/// ```
+pub fn set_global_executor<R>(cx: &mut Cx, runtime: R) -> Result<(), R>
+where
+    R: Runtime,
+{
+    if RUNTIME.get(cx).is_some() {
+        return Err(runtime);
+    }
+
+    RUNTIME.get_or_init(cx, || Box::new(runtime));
+
+    Ok(())
+}
--- a/crates/neon/src/lib.rs
+++ b/crates/neon/src/lib.rs
@@ -102,6 +102,9 @@ mod types_impl;
 #[cfg_attr(docsrs, doc(cfg(feature = "sys")))]
 pub mod sys;
 
+#[cfg(all(feature = "napi-6", feature = "futures"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "napi-6", feature = "futures"))))]
+pub use executor::set_global_executor;
 pub use types_docs::exports as types;
 
 #[doc(hidden)]
@@ -114,12 +117,15 @@ use crate::{context::ModuleContext, handle::Handle, result::NeonResult, types::J
 #[cfg(feature = "napi-6")]
 mod lifecycle;
 
+#[cfg(all(feature = "napi-6", feature = "futures"))]
+mod executor;
+
 #[cfg(feature = "napi-8")]
 static MODULE_TAG: once_cell::sync::Lazy<crate::sys::TypeTag> = once_cell::sync::Lazy::new(|| {
     let mut lower = [0; std::mem::size_of::<u64>()];
 
     // Generating a random module tag at runtime allows Neon builds to be reproducible. A few
-    //  alternativeswere considered:
+    //  alternatives considered:
     // * Generating a random value at build time; this reduces runtime dependencies but, breaks
     //   reproducible builds
     // * A static random value; this solves the previous issues, but does not protect against ABI
--- a/crates/neon/src/lib.rs
+++ b/crates/neon/src/lib.rs
@@ -149,18 +149,21 @@ pub struct Exports(());
 impl Exports {
     /// Export all values exported with [`neon::export`](export)
     ///
-    /// ```ignore
+    /// ```
+    /// # fn main() {
     /// # use neon::prelude::*;
     /// #[neon::main]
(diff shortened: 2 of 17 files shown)


crates/neon/src/macros.rs
```diff
@@ -122,6 +124,61 @@ pub use neon_macros::main;
...
+/// A `fn` that returns a [`Future`](std::future::Future) can be annotated with `#[neon::export(async)]`
+/// if it needs to perform some setup on the JavaScript main thread before running asynchronously.
+///
+/// ```
+/// # #[cfg(all(feature = "napi-6", feature = "futures"))]
+/// # {
+/// # use std::future::Future;
+/// # use neon::prelude::*;
+/// #[neon::export(async)]
+/// fn add(_cx: &mut FunctionContext, a: f64, b: f64) -> NeonResult<impl Future<Output = f64>> {
```
kjvalencik, line 152: @dherman This doesn't need the unused `_cx`, but I was trying to show that it _can_ use a `cx`. A few options:
* Leave it as is
* Switch to `console.log` to demonstrate using it
* Remove it
* Comment it out to show that it _could_ be used but isn't required

```suggestion
/// fn add(/* cx: &mut FunctionContext, */ a: f64, b: f64) -> NeonResult<impl Future<Output = f64>> {
```

What do you think?

crates/neon/src/macro_internal/mod.rs
```diff
@@ -17,46 +28,86 @@ pub static EXPORTS: [for<'cx> fn(&mut ModuleContext<'cx>) -> NeonResult<Export<'
 #[linkme::distributed_slice]
 pub static MAIN: [for<'cx> fn(ModuleContext<'cx>) -> NeonResult<()>];
 
-// Provides an identically named method to `NeonExportReturnJson` for easy swapping in macros
-pub trait NeonExportReturnValue<'cx> {
-    fn try_neon_export_return(self, cx: &mut Cx<'cx>) -> JsResult<'cx, JsValue>;
+// Wrapper for the value type and return type tags
+pub struct NeonMarker<Tag, Return>(PhantomData<Tag>, PhantomData<Return>);
```
kjvalencik, line 32: Possible renaming to clarify roles of these markers.

```rust
pub struct NeonMarker<Transform, ReturnType>(PhantomData<Transform>, PhantomData<ReturnType>);

pub struct NeonJsonTransform;
pub struct NeonIdentityTransform;
pub struct NeonResultReturn;
pub struct NeonValueReturn;
```

Does this improve readability?

crates/neon/src/lib.rs
```diff
@@ -143,18 +143,21 @@ pub struct Exports(());
...
     /// fn main(mut cx: ModuleContext) -> NeonResult<()> {
     ///     neon::registered().export(&mut cx)?;
     ///     Ok(())
     /// }
+    /// # }
     /// ```
     ///
     /// For more control, iterate over exports.
     ///
-    /// ```ignore
```
kjvalencik, line 157: On my laptop, this test takes 9s with a hot cache and 45s with a completely cold cache.

While that's slower than I'd like, I feel like that's fast enough to include by default. I've caught regressions in feature flag combos a few times. What do you think?

crates/neon/src/macros.rs
dherman: Honestly I feel like we should have a shorthand for the console, so you could do:

```rust
cx.console()?.log(&mut cx, "Hello from the main JavaScript thread!")?;
```

But for now why don't we just remove the `cx` argument entirely and the `NeonResult`, and we can add it back after we merge #1056, because then we could at least write:

```rust
    cx.global("console")?
        .get(&mut cx, "log")?
        .bind(&mut cx)
        .arg("Hello from the main JavaScript thread!")
        .apply()?;
```

crates/neon/src/macro_internal/futures.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub fn spawn<'cx, F, S>(cx: &mut Cx<'cx>, fut: F, settle: S) -> JsResult<'cx, JsValue>
+where
+    F: Future + Send + 'static,
+    F::Output: Send,
+    S: FnOnce(TaskContext, F::Output) -> JsResult<JsValue> + Send + 'static,
+{
+    let rt = match crate::executor::RUNTIME.get(cx) {
+        Some(rt) => rt,
+        None => return cx.throw_error("Must initialize with neon::set_global_executor"),
```
dherman, line 17: Nite: our error messages are not conventionally capitalized.

crates/neon/src/sys/bindings/mod.rs
```diff
@@ -39,7 +39,7 @@ macro_rules! napi_name {
 ///
 /// Sample input:
 ///
-/// ```ignore
+/// ```skip
```
dherman, line 42: What does `skip` do? I've never seen this one and I couldn't find it in the rustdoc book.

crates/neon/src/types_impl/extract/private.rs
```diff
@@ -35,6 +36,8 @@ impl Sealed for &str {}
 
 impl<'cx, V: Value> Sealed for Handle<'cx, V> {}
 
+impl<O: Object> Sealed for Root<O> {}
```
dherman, line 39: Is this here so that it passes the preconditions for a `Root<O>` to be able to implement `TryFromJs`?

dherman: Update: yes, now that I see the diff in try_from_js.rs, my understanding was right.

test/napi/src/lib.rs
```diff
@@ -27,6 +29,9 @@ mod js {
 
 #[neon::main]
 fn main(mut cx: ModuleContext) -> NeonResult<()> {
+    let rt = runtime(&mut cx)?;
+
+    neon::set_global_executor(&mut cx, rt).or_else(|_| cx.throw_error("Executor already set"))?;
```
dherman, line 34: nit: we don't capitalize error messages

crates/neon/src/macros.rs
dherman: Actually I think it's probably best to leave them out for good. The docs don't talk about taking a `FunctionContext` until the bottom of the page, so it's best not to jump ahead. IMO we should keep it simple with `println!` for these examples.

dherman (changes requested):
A few small changes, but nothing major. TBH I'm a little worried my lack of expertise at tokio and async means my review might be a little superficial. But the API is so easy to understand and fits so nicely with the `#[export]` macro syntax. This is a really exciting addition.

crates/neon/src/lifecycle.rs
```diff
@@ -102,7 +102,7 @@ impl LocalCell {
         }
     }
 
-    pub(crate) fn get<'cx, 'a, C>(cx: &'a mut C, id: usize) -> Option<&mut LocalCellValue>
+    pub(crate) fn get<'cx, 'a, C>(cx: &'a mut C, id: usize) -> Option<&'a mut LocalCellValue>
```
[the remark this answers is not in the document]
kjvalencik: Same as the other comment. This is part of a clippy lint fix and this was the inferred lifetime. No change.

crates/neon/src/executor/mod.rs
```diff
@@ -0,0 +1,59 @@
+use std::{future::Future, pin::Pin};
+
+use crate::{context::Cx, thread::LocalKey};
+
+#[cfg(feature = "tokio-rt")]
+pub(crate) mod tokio;
+
+type BoxFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
```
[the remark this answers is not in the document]
kjvalencik: I need the `Runtime` trait to be object-safe so that I can put it in a `Box<dyn Runtime>`; this means that the `spawn(..)` function can't be generic. Since it can't be generic, it takes a boxed dyn future.

Boxed futures need be `Pin` because of [`Future::poll`](https://doc.rust-lang.org/stable/std/future/trait.Future.html#tymethod.poll) which takes `self: Pin<&mut Self>`. Without the `Pin`, it could be moved in memory and you wouldn't be able to poll it.

This is a very common type in async Rust. The `futures` crate even includes a [type alias for it](https://docs.rs/futures-core/0.3.30/futures_core/future/type.BoxFuture.html).

dherman (approved):
🚀

PR #1056, opened 2024-07-08
feat(neon): `JsFunction::bind()` and `Object::prop()`

dherman:
This PR implements `JsFunction::bind()`, which creates a builder for calling a function using the `Try{From,Into}Js` traits.

```rust
let n = f.bind(&mut cx)
    .args((1, 2, 3))?
    .arg(4)?
    .arg_with(|cx| Ok(cx.number(5)))?
    .call::<f64>()?;
```

It also implements `Object::prop()`, which creates a builder for accessing object properties using the `Try{From,Into}Js` traits.

```rust
let x: f64 = obj
    .prop(&mut cx, "x")
    .get()?;

obj.prop(&mut cx, "y")
    .set(x)?;

let s: String = obj.prop(&mut cx, "toString")
    .bind()?
    .call()?;
```

Finally, it offers a shorthand convenience method `Object::method()`, which extracts an object property and binds it as a method:

```rust
let s: String = obj.method(&mut cx, "toString").call()?;
```

### DIFF
--- a/crates/neon/src/event/channel.rs
+++ b/crates/neon/src/event/channel.rs
@@ -75,13 +75,12 @@ type Callback = Box<dyn FnOnce(sys::Env) + Send + 'static>;
 ///         // loop. This _will_ block the event loop while executing.
 ///         channel.send(move |mut cx| {
 ///             let callback = callback.into_inner(&mut cx);
-///             let this = cx.undefined();
-///             let args = vec![
-///                 cx.null().upcast::<JsValue>(),
-///                 cx.number(result).upcast(),
-///             ];
+///             let null = cx.null();
 ///
-///             callback.call(&mut cx, this, args)?;
+///             callback
+///                 .bind(&mut cx)
+///                 .args((null, result))?
+///                 .exec()?;
 ///
 ///             Ok(())
 ///         });
--- a/crates/neon/src/event/channel.rs
+++ b/crates/neon/src/event/channel.rs
@@ -75,11 +75,10 @@ type Callback = Box<dyn FnOnce(sys::Env) + Send + 'static>;
 ///         // loop. This _will_ block the event loop while executing.
 ///         channel.send(move |mut cx| {
 ///             let callback = callback.into_inner(&mut cx);
-///             let null = cx.null();
 ///
 ///             callback
 ///                 .bind(&mut cx)
-///                 .args((null, result))?
+///                 .args(((), result))?
 ///                 .exec()?;
 ///
 ///             Ok(())
--- a/crates/neon/src/object/mod.rs
+++ b/crates/neon/src/object/mod.rs
@@ -217,6 +217,14 @@ where
 /// The trait of all object types.
 pub trait Object: Value {
     /// Create a [`PropOptions`] for accessing a property.
+    ///
+    /// # Safety
+    ///
+    /// Because `cx` is a mutable reference, Neon guarantees it
+    /// is the context with the shortest possible lifetime, so
+    /// replacing the lifetime `'self` with `'cx` cannot extend
+    /// the lifetime of the property beyond the lifetime of the
+    /// object.
     fn prop<'a, 'cx: 'a, K: PropertyKey>(
         &self,
         cx: &'a mut Cx<'cx>,
--- a/crates/neon/src/object/mod.rs
+++ b/crates/neon/src/object/mod.rs
@@ -297,7 +297,7 @@ pub trait Object: Value {
         })
     }
 
-    #[doc(hidden)]
+    #[deprecated(since = "TBD", note = "use `Object::prop()` instead")]
     fn get_opt<'a, V: Value, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
@@ -312,7 +312,7 @@ pub trait Object: Value {
         v.downcast_or_throw(cx).map(Some)
     }
 
-    #[doc(hidden)]
+    #[deprecated(since = "TBD", note = "use `Object::prop()` instead")]
     fn get_value<'a, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
@@ -323,7 +323,7 @@ pub trait Object: Value {
         })
     }
 
-    #[doc(hidden)]
+    #[deprecated(since = "TBD", note = "use `Object::prop()` instead")]
     fn get<'a, V: Value, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
@@ -368,7 +368,7 @@ pub trait Object: Value {
         }
     }
 
-    #[doc(hidden)]
+    #[deprecated(since = "TBD", note = "use `Object::prop()` instead")]
     fn set<'a, C: Context<'a>, K: PropertyKey, W: Value>(
         &self,
         cx: &mut C,
@@ -389,7 +389,7 @@ pub trait Object: Value {
         Root::new(cx, self)
     }
 
-    #[doc(hidden)]
+    #[deprecated(since = "TBD", note = "use `Object::method()` instead")]
     fn call_method_with<'a, C, K>(&self, cx: &mut C, method: K) -> NeonResult<CallOptions<'a>>
     where
         C: Context<'a>,
--- a/crates/neon/src/object/mod.rs
+++ b/crates/neon/src/object/mod.rs
@@ -38,11 +38,8 @@ use crate::{
     result::{NeonResult, Throw},
     sys::{self, raw},
     types::{
-        build,
-        extract::{TryFromJs, TryIntoJs},
-        function::{BindOptions, CallOptions},
-        utf8::Utf8,
-        JsFunction, JsUndefined, JsValue, Value,
+        build, extract::{TryFromJs, TryIntoJs}, function::{BindOptions, CallOptions}, utf8::Utf8, JsFunction, JsObject, JsUndefined, JsValue, Value,
+        private::ValueInternal
     },
 };
 
@@ -161,21 +158,19 @@ impl<'a> PropertyKey for &'a str {
 /// # Ok(cx.string(s))
 /// # }
 /// ```
-pub struct PropOptions<'a, 'cx, O, K>
+pub struct PropOptions<'a, 'cx, K>
 where
     'cx: 'a,
-    O: Object,
     K: PropertyKey,
 {
     pub(crate) cx: &'a mut Cx<'cx>,
-    pub(crate) this: Handle<'cx, O>,
+    pub(crate) this: Handle<'cx, JsObject>,
     pub(crate) key: K,
 }
 
-impl<'a, 'cx, O, K> PropOptions<'a, 'cx, O, K>
+impl<'a, 'cx, K> PropOptions<'a, 'cx, K>
 where
     'cx: 'a,
-    O: Object,
     K: PropertyKey,
 {
     /// Gets the property from the object and attempts to convert it to a Rust value.
@@ -216,8 +211,8 @@ pub trait Object: Value {
         &self,
         cx: &'a mut Cx<'cx>,
         key: K,
-    ) -> PropOptions<'a, 'cx, Self, K> {
-        let this = Handle::new_internal(unsafe { Self::from_local(cx.env(), self.to_local()) });
+    ) -> PropOptions<'a, 'cx, K> {
+        let this: Handle<'_, JsObject> = Handle::new_internal(unsafe { ValueInternal::from_local(cx.env(), self.to_local()) });
         PropOptions { cx, this, key }
     }
 
--- a/crates/neon/src/object/mod.rs
+++ b/crates/neon/src/object/mod.rs
@@ -32,14 +32,20 @@
 //! [hierarchy]: crate::types#the-javascript-type-hierarchy
 //! [symbol]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol
 
+use smallvec::smallvec;
+
 use crate::{
     context::{internal::ContextInternal, Context, Cx},
     handle::{Handle, Root},
     result::{NeonResult, Throw},
     sys::{self, raw},
     types::{
-        build, extract::{TryFromJs, TryIntoJs}, function::{BindOptions, CallOptions}, utf8::Utf8, JsFunction, JsObject, JsUndefined, JsValue, Value,
-        private::ValueInternal
+        build,
+        extract::{TryFromJs, TryIntoJs},
+        function::{BindOptions, CallOptions},
+        private::ValueInternal,
+        utf8::Utf8,
+        JsFunction, JsObject, JsUndefined, JsValue, Value,
     },
 };
 
@@ -197,10 +203,14 @@ where
     /// May throw an exception either during accessing the property or downcasting it
     /// to a function.
     pub fn bind(&'a mut self) -> NeonResult<BindOptions<'a, 'cx>> {
-        let callee: Handle<JsFunction> = self.this.get(self.cx, self.key)?;
-        let mut bind = callee.bind(self.cx);
-        bind.this(self.this)?;
-        Ok(bind)
+        let callee: Handle<JsValue> = self.this.get(self.cx, self.key)?;
+        let this = Some(self.this.upcast());
+        Ok(BindOptions {
+            cx: self.cx,
+            callee,
+            this,
+            args: smallvec![],
+        })
     }
 }
 
@@ -212,7 +222,8 @@ pub trait Object: Value {
         cx: &'a mut Cx<'cx>,
         key: K,
     ) -> PropOptions<'a, 'cx, K> {
-        let this: Handle<'_, JsObject> = Handle::new_internal(unsafe { ValueInternal::from_local(cx.env(), self.to_local()) });
+        let this: Handle<'_, JsObject> =
+            Handle::new_internal(unsafe { ValueInternal::from_local(cx.env(), self.to_local()) });
         PropOptions { cx, this, key }
     }
 
--- a/crates/neon/src/object/mod.rs
+++ b/crates/neon/src/object/mod.rs
@@ -16,16 +16,15 @@
 //!
 //! ```
 //! # use neon::prelude::*;
-//! fn set_and_check<'a>(
-//!     cx: &mut impl Context<'a>,
-//!     obj: Handle<'a, JsObject>
-//! ) -> JsResult<'a, JsValue> {
-//!     let value = cx.string("hello!");
+//! fn set_and_check<'cx>(
+//!     cx: &mut Cx<'cx>,
+//!     obj: Handle<'cx, JsObject>
+//! ) -> JsResult<'cx, JsValue> {
 //!     // set property "17" with integer shorthand
-//!     obj.set(cx, 17, value)?;
+//!     obj.prop(cx, 17).set("hello")?;
 //!     // get property "17" with string shorthand
 //!     // returns the same value ("hello!")
-//!     obj.get(cx, "17")
+//!     obj.prop(cx, "17").get()
 //! }
 //! ```
 //!
@@ -210,7 +209,6 @@ where
     }
 
     /// Gets the property from the object and attempts to convert it to a Rust value.
-    /// Equivalent to calling `R::from_js(cx, obj.get(cx)?)`.
     ///
     /// May throw an exception either during accessing the property or converting the
     /// result type.
@@ -220,7 +218,6 @@ where
     }
 
     /// Sets the property on the object to a value converted from Rust.
-    /// Equivalent to calling `obj.set(cx, v.try_into_js(cx)?)`.
     ///
     /// May throw an exception either during converting the value or setting the property.
     pub fn set<V: TryIntoJs<'cx>>(&mut self, v: V) -> NeonResult<&mut Self> {
@@ -230,7 +227,6 @@ where
     }
 
     /// Sets the property on the object to a value computed from a closure.
-    /// Equivalent to calling `obj.set(cx, f(cx).try_into_js(cx)?)`.
     ///
     /// May throw an exception either during converting the value or setting the property.
     pub fn set_with<R, F>(&mut self, f: F) -> NeonResult<&mut Self>
@@ -280,8 +276,7 @@ pub trait Object: Value {
         PropOptions { cx, this, key }
     }
 
-    /// Gets a property from a JavaScript object that may be `undefined` and
-    /// attempts to downcast the value if it existed.
+    #[doc(hidden)]
     fn get_opt<'a, V: Value, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
@@ -296,10 +291,7 @@ pub trait Object: Value {
         v.downcast_or_throw(cx).map(Some)
     }
 
-    /// Gets a property from a JavaScript object as a [`JsValue`].
-    ///
-    /// If a [`getter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)
-    /// is defined on the object, it will be called.
+    #[doc(hidden)]
     fn get_value<'a, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
@@ -310,10 +302,7 @@ pub trait Object: Value {
         })
     }
 
-    /// Gets a property from a JavaScript object and attempts to downcast as a specific type.
-    /// Equivalent to calling `obj.get_value(&mut cx)?.downcast_or_throw(&mut cx)`.
-    ///
-    /// Throws an exception if the value is a different type.
+    #[doc(hidden)]
     fn get<'a, V: Value, C: Context<'a>, K: PropertyKey>(
         &self,
         cx: &mut C,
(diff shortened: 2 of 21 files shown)


crates/neon/src/object/mod.rs
```diff
@@ -134,8 +140,77 @@ impl<'a> PropertyKey for &'a str {
...
+        let v = v.try_into_js(self.cx)?;
+        self.this.set(self.cx, self.key, v)
+    }
+
+    /// Gets the property from the object as a method and binds `this` to the object.
+    ///
+    /// May throw an exception either during accessing the property or downcasting it
+    /// to a function.
+    pub fn bind(&'a mut self) -> NeonResult<BindOptions<'a, 'cx, C>> {
+        let callee: Handle<JsFunction> = self.this.get(self.cx, self.key)?;
```
kjvalencik, line 195: Downcasting to `JsFunction` makes an extra API call that implicitly happens in Node-API when we invoke the function.

We could avoid this call by updating `BindOptions` to take a `JsValue` and updating `call` to throw a type error if `napi_status` is a type error instead of panicking. This is a bit of a performance option and also consistent with how `TryFromJs` is implemented.

If users prefer the eager check, they can use a slightly more verbose version:
```rust
- let bind_opts = o.prop(&mut cx, "myMethod").bind();
+ let bind_opts = o.prop(&mut cx, "myMethod").get::<JsFunction>().bind(&mut cx);
```

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +140,77 @@ impl<'a> PropertyKey for &'a str {
...
+/// # Ok(cx.string(s))
+/// # }
+/// ```
+pub struct PropOptions<'a, 'cx: 'a, C: Context<'cx>, O: Object, K: PropertyKey> {
+    pub(crate) cx: &'a mut C,
+    pub(crate) this: Handle<'cx, O>,
+    pub(crate) key: K,
+}
+
+impl<'a, 'cx: 'a, C: Context<'cx>, O: Object, K: PropertyKey> PropOptions<'a, 'cx, C, O, K> {
```
kjvalencik, line 170: Nit, when generic bounds get complicated, I find `where` syntax much easier to read.

```rust
impl<'a, 'cx, C, O, K> PropOptions<'a, 'cx, C, O, K>
where
    'cx: 'a,
    C: Context<'cx>,
    O: Object,
    K: PropertyKey,
{
```

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +140,77 @@ impl<'a> PropertyKey for &'a str {
...
+
 /// The trait of all object types.
 pub trait Object: Value {
+    /// Create a [`PropOptions`] for accessing a property.
+    fn prop<'a, 'cx: 'a, C: Context<'cx>, K: PropertyKey>(
+        &self,
+        cx: &'a mut C,
+        key: K,
+    ) -> PropOptions<'a, 'cx, C, Self, K> {
+        let this = Handle::new_internal(unsafe { Self::from_local(cx.env(), self.to_local()) });
```
kjvalencik, line 210: Does this need `unsafe` or could it be `self.as_value()`?

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -7,11 +7,75 @@ use crate::{
...
+pub struct BindOptions<'a, 'cx: 'a, C: Context<'cx>> {
+    pub(crate) cx: &'a mut C,
+    pub(crate) callee: Handle<'cx, JsFunction>,
+    pub(crate) this: Option<Handle<'cx, JsValue>>,
+    pub(crate) args: private::ArgsVec<'cx>,
+}
+
+impl<'a, 'cx: 'a, C: Context<'cx>> BindOptions<'a, 'cx, C> {
+    /// Set the value of `this` for the function call.
+    pub fn this<V: Value>(&mut self, this: Handle<'cx, V>) -> &mut Self {
```
kjvalencik, line 41: Can we make this use `TryIntoJs` too?

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -7,11 +7,75 @@ use crate::{
...
+        let v = a.try_into_js(self.cx)?;
+        self.args.push(v.upcast());
+        Ok(self)
+    }
+
+    /// Add an argument to the arguments list, computed from a closure.
+    pub fn arg_with<R, F>(&mut self, f: F) -> NeonResult<&mut Self>
+    where
+        R: TryIntoJs<'cx>,
+        F: FnOnce(&mut C) -> NeonResult<R>,
```
kjvalencik, line 63: `TryIntoJs` is implemented for [`Result<T, E>`](https://docs.rs/neon/1.1.0-alpha.0/neon/types/extract/trait.TryIntoJs.html#impl-TryIntoJs%3C'cx%3E-for-Result%3CT,+E%3E) and [`Throw`](https://docs.rs/neon/1.1.0-alpha.0/neon/types/extract/trait.TryIntoJs.html#impl-TryIntoJs%3C'cx%3E-for-Throw).

This means requiring the closure to return `NeonResult` is redundant. Instead, it can return `R` and that allows the user to *either* return a `NeonResult` or use an infallible function.

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -7,11 +7,75 @@ use crate::{
...
+
+    /// Add an argument to the arguments list.
+    pub fn arg<A: TryIntoJs<'cx>>(&mut self, a: A) -> NeonResult<&mut Self> {
+        let v = a.try_into_js(self.cx)?;
+        self.args.push(v.upcast());
+        Ok(self)
+    }
+
+    /// Add an argument to the arguments list, computed from a closure.
+    pub fn arg_with<R, F>(&mut self, f: F) -> NeonResult<&mut Self>
```
kjvalencik, line 60: Should there be a version of this function that returns `R: TryIntoArguments`?

I think *no* because a user should _probably_ just create this tuple before calling `.bind(..)`. It's more API noise without a valuable use case.

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -7,11 +7,75 @@ use crate::{
...
+        self.args.push(v.upcast());
+        Ok(self)
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is converted to a Rust value with `TryFromJs::from_js`.
+    pub fn apply<R: TryFromJs<'cx>>(&mut self) -> NeonResult<R> {
+        let this = self.this.unwrap_or_else(|| self.cx.undefined().upcast());
+        let v: Handle<JsValue> = self.callee.call(self.cx, this, &self.args)?;
+        R::from_js(self.cx, v)
```
kjvalencik, line 75: We haven't published `TryFromJs` yet and can still make breaking changes. This makes me wonder if we *shouldn't* make `from_js` part of the API and should _instead_ create a `impl TryFromJs for NeonResult`. 🤔 

We can flag this for follow-up.

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -111,6 +175,109 @@ impl<'a> ConstructOptions<'a> {
...
+         impl<'cx, $($tprefix: TryIntoJs<'cx> + 'cx, )* $tname1: TryIntoJs<'cx> + 'cx> TryIntoArguments<'cx> for ($($tprefix, )* $tname1, ) {}
+
+         impl_into_arguments! {
+             [ $(($tprefix, $vprefix), )* ($tname1, $vname1), ];
+             [ $($(#[$attrs])? ($tnames, $vnames), )* ];
+         }
+     };
+}
+
+impl_into_arguments! {
```
kjvalencik, line 219: The [`from_args_impl`](https://github.com/neon-bindings/neon/blob/5d16c4d0c040c7d81565673ba4b42929922e03e1/crates/neon/src/types_impl/extract/mod.rs#L229-L288) macro could be good to look at to reduce the noise here a bit.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1211,6 +1211,17 @@ impl JsFunction {
     }
 }
 
+impl JsFunction {
+    pub fn bind<'a, 'cx: 'a, C: Context<'cx>>(&self, cx: &'a mut C) -> BindOptions<'a, 'cx, C> {
+        BindOptions {
+            cx,
+            callee: Handle::new_internal(unsafe { self.clone() }),
```
kjvalencik, line 1218: I think this can be `self.as_value()` instead of unsafe..

kjvalencik (changes requested):

crates/neon/src/types_impl/mod.rs
```diff
@@ -1154,6 +1154,23 @@ impl JsFunction {
...
+    this: Handle<'b, T>,
+    args: AS
+) -> JsResult<'a, JsValue>
+where
+    T: Value,
+    AS: AsRef<[Handle<'b, JsValue>]>,
+{
+    let (argc, argv) = unsafe { prepare_call(cx, args.as_ref()) }?;
+    let env = cx.env().to_raw();
+    build(cx.env(), |out| unsafe {
```
kjvalencik, line 1169:
```rust
match status {
    sys::Status::FunctionExpected => todo!("Write code to convert to an exception"),
    sys::Status::PendingException => return Err(Throw::new()),
    status => assert_eq!(status, sys::Status::Ok),
}
```

https://github.com/neon-bindings/neon/blob/5d16c4d0c040c7d81565673ba4b42929922e03e1/crates/neon/src/sys/bindings/types.rs#L109

crates/neon/src/object/mod.rs
dherman: `as_value()` returns a `Handle<'_, JsValue>` but we need a `Handle<'_, Self>`. We don't actually have the original handle so we can't clone it… Not sure if there's another way you can think of to synthesize a new handle.

crates/neon/src/types_impl/function/mod.rs
dherman: Sure, although in practice I imagine 99% of the time people will only call this method with an object handle.

crates/neon/src/types_impl/function/mod.rs
dherman: Do you mean something like:
```rust
pub fn args_with<R, F>(&mut self, f: F) -> NeonResult<&mut Self>
where
    R: TryIntoArguments<'cx>,
    F: FnOnce(&mut Cx<'cx>) -> R,
{
    // ...
}
```
?

crates/neon/src/types_impl/function/mod.rs
kjvalencik: Nevermind, we can't actually have this. It would end up with a `NeonResult<Result<NeonResult<T>, Infallible>>`, which is really messy.

I think what I could do is make `from_js` a different trait with a blanket implementation`where <Self as TryFromJs<'cx>>::Error: TryIntoJs<'cx>`. We can discuss that outside this PR.

crates/neon/src/object/mod.rs
dherman: Resolved: remove the `O: Object` type parameter on `PropOptions` and store it as a `JsValue`.

dherman: Update: store it as a `JsObject` so we can use `get_value` and `set_value`. (In future if we extended `Object` to be implemented on all `Value` this would become `JsValue`.)

crates/neon/src/types_impl/function/mod.rs
dherman: Resolved: needs another impl for `With` on `TryIntoArguments`, and the actual implementation of `.args_with` could be to delegate to `.args(With(...))`.

crates/neon/src/types_impl/mod.rs
dherman: Turns out it's `InvalidArg`, not `FunctionExpected`, but otherwise this was perfect guidance.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1154,6 +1155,50 @@ impl JsFunction {
...
+{
+    let (argc, argv) = unsafe { prepare_call(cx, args.as_ref()) }?;
+    let env = cx.env();
+    let mut result: MaybeUninit<raw::Local> = MaybeUninit::zeroed();
+
+    let status = napi::call_function(
+        env.to_raw(),
+        this.to_local(),
+        callee,
+        argc as usize,
```
kjvalencik, line 1176: Now that the C++ backend is removed, we can update `prepare_call` to return a `usize`. We are casting this to a `u32` just to cast it back to `usize`.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1154,6 +1155,50 @@ impl JsFunction {
...
+        result.as_mut_ptr(),
+    );
+
+    match status {
+        sys::Status::InvalidArg => {
+            if !sys::tag::is_function(env.to_raw(), callee) {
+                return cx.throw_error("not a function");
+            }
+            panic!("invalid argument");
+        }
```
kjvalencik, line 1187:
```suggestion
        sys::Status::InvalidArg if !sys::tag::is_function(env.to_raw(), callee) => {
            return cx.throw_error("not a function");
        }
```

If we move the `is_function` to a match guard, it will fall through to the default arm which asserts against `sys::Status::Ok`. This is a bit more consistent with [other asserts against `sys::Status`](https://github.com/neon-bindings/neon/blob/63551b18cb509a30598c06b9ec0b983a76fc687b/crates/neon/src/types_impl/extract/try_from_js.rs#L84-L88).

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +143,90 @@ impl<'a> PropertyKey for &'a str {
...
 /// The trait of all object types.
 pub trait Object: Value {
+    /// Create a [`PropOptions`] for accessing a property.
+    fn prop<'a, 'cx: 'a, K: PropertyKey>(
+        &self,
+        cx: &'a mut Cx<'cx>,
+        key: K,
+    ) -> PropOptions<'a, 'cx, K> {
+        let this: Handle<'_, JsObject> =
+            Handle::new_internal(unsafe { ValueInternal::from_local(cx.env(), self.to_local()) });
```
kjvalencik, line 226: We are extending the lifetime `'self` to `'cx`. If I understand correctly, this is *safe* because `'cx` will *always* be the shortest possible lifetime of all handles. This is because any longer contexts will be "locked" with a `&mut self`.

In other words, this is safe because Neon invariants enforce `'self: 'cx` for all handles.

crates/neon/src/types_impl/extract/with.rs
```diff
@@ -48,4 +58,21 @@ where
     }
 }
 
+impl<'cx, F, O> TryIntoArgumentsInternal<'cx> for With<F, O>
```
kjvalencik, line 61: Nit, it makes a bit more sense to me for this implementation to be with the `TryIntoArguments` code rather than here, but if you think it's better here, I'm okay with that.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1154,6 +1155,47 @@ impl JsFunction {
     }
 }
 
+pub(crate) unsafe fn call_local<'a, 'b, C: Context<'a>, T, AS>(
+    cx: &mut C,
+    callee: raw::Local,
+    this: Handle<'b, T>,
```
kjvalencik, line 1161: Nit, it's a little odd for this to take a `Handle` when `callee` is a `raw::Local`. Can these both be handles or both `Local`?

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -3,15 +3,93 @@
...
+/// # }
+/// ```
+pub struct BindOptions<'a, 'cx: 'a> {
+    pub(crate) cx: &'a mut Cx<'cx>,
+    pub(crate) callee: Handle<'cx, JsValue>,
+    pub(crate) this: Option<Handle<'cx, JsValue>>,
+    pub(crate) args: private::ArgsVec<'cx>,
+}
+
+impl<'a, 'cx: 'a> BindOptions<'a, 'cx> {
```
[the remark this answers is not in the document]
dherman: We might want to rename `.apply()` to `.call()` so we have `.call()` and `.construct()`.

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +143,90 @@ impl<'a> PropertyKey for &'a str {
...
+    pub fn get<R: TryFromJs<'cx>>(&mut self) -> NeonResult<R> {
+        let v = self.this.get_value(self.cx, self.key)?;
+        R::from_js(self.cx, v)
+    }
+
+    /// Sets the property on the object to a value converted from Rust.
+    /// Equivalent to calling `obj.set(cx, v.try_into_js(cx)?)`.
+    ///
+    /// May throw an exception either during converting the value or setting the property.
+    pub fn set<V: TryIntoJs<'cx>>(&mut self, v: V) -> NeonResult<bool> {
```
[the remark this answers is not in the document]
kjvalencik: Without chaining:

```rust
let o = cx.empty_object();
o.prop("a").set("this is a")?;
o.prop("b").set("this is b")?;
o.prop("c").set("this is c")?;
```

crates/neon/src/object/mod.rs
dherman: I should add this to a `// SAFETY` comment in the code.

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +143,90 @@ impl<'a> PropertyKey for &'a str {
...
+            cx: self.cx,
+            callee,
+            this,
+            args: smallvec![],
+        })
+    }
+}
+
 /// The trait of all object types.
 pub trait Object: Value {
```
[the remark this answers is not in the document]
dherman: Let's make them `#[doc(hidden)]` for now and we can file a followup issue to formally deprecate them a minor version or two down the road.

crates/neon/src/types_impl/mod.rs
dherman: I moved it into a method of `ValueInternal`, so it doesn't feel as odd:

```rust
    unsafe fn try_call<'a, 'b, C: Context<'a>, T, AS>(
        &self,
        cx: &mut C,
        this: Handle<'b, T>,
        args: AS,
    ) -> JsResult<'a, JsValue>
    where
        T: Value,
        AS: AsRef<[Handle<'b, JsValue>]>,
    { /* ... */ }

// ...

unsafe { self.try_call(cx, this, args) }
```

crates/neon/src/types_impl/extract/with.rs
dherman: I moved it to `function/mod.rs` with the other impls.

crates/neon/src/object/mod.rs
dherman: Added a `/// # Safety` comment.

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +143,90 @@ impl<'a> PropertyKey for &'a str {
...
+/// obj.prop(&mut cx, "y")
+///     .set(x)?;
+///
+/// let s: String = obj.prop(&mut cx, "toString")
+///     .bind()?
+///     .apply()?;
+/// # Ok(cx.string(s))
+/// # }
+/// ```
+pub struct PropOptions<'a, 'cx, K>
```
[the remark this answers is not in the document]
dherman: I don't think we can because it wraps the `&mut Cx`.

crates/neon/src/object/mod.rs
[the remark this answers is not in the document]
dherman: I put back the type parameter for precise typing, and I called this method `.this()`. I added an example of chaining to the doc comment:
```rust
let obj = cx.empty_object()
    .prop(&mut cx, "x")
    .set(1)?
    .prop("y")
    .set(2)?
    .prop("color")
    .set("blue")?
    .this();
```

crates/neon/src/object/mod.rs
```diff
@@ -134,8 +143,90 @@ impl<'a> PropertyKey for &'a str {
...
+where
+    'cx: 'a,
+    K: PropertyKey,
+{
+    pub(crate) cx: &'a mut Cx<'cx>,
+    pub(crate) this: Handle<'cx, JsObject>,
+    pub(crate) key: K,
+}
+
+impl<'a, 'cx, K> PropOptions<'a, 'cx, K>
```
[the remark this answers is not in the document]
dherman: Added `this()` and `set_with()`.

crates/neon/src/object/mod.rs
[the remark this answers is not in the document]
dherman: I made the following methods `#[doc(hidden)]`:
- `get_opt()`
- `get()`
- `get_value()`
- `set()`
- `call_method_with()`

And I've replaced all the instances of them in the doc comments.

Next I'll do the same for `.call()`, `.construct()`, `.call_with()`, `.construct_with()`, once I've implemented the full range of replacements via `.bind()`.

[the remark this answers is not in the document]
dherman: I found a better mechanism than `#[doc(hidden)]`: Rust now allows for soft deprecation using `#[deprecated(since = "TBD")]`. This will leave a deprecation note in the API docs ([example](https://deploy-preview-1056--neon-bindings-docs.netlify.app/neon/object/trait.object#method.get)) but not generate compiler warnings. Much better!

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -3,15 +3,93 @@
...
+        F: FnOnce(&mut Cx<'cx>) -> R,
+    {
+        let v = f(self.cx).try_into_js(self.cx)?;
+        self.args.push(v.upcast());
+        Ok(self)
+    }
+
+    /// Make the function call. If the function returns without throwing, the result value
+    /// is converted to a Rust value with `TryFromJs::from_js`.
+    pub fn apply<R: TryFromJs<'cx>>(&mut self) -> NeonResult<R> {
```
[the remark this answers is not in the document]
dherman: Added.

crates/neon/src/types_impl/function/mod.rs
```diff
@@ -111,6 +189,73 @@ impl<'a> ConstructOptions<'a> {
     }
 }
 
+/// The trait for specifying values to be converted into arguments for a function call.
+/// This trait is sealed and cannot be implemented by types outside of the Neon crate.
+///
+/// **Note:** This trait is implemented for tuples of up to 32 JavaScript values,
+/// but for the sake of brevity, only tuples up to size 8 are shown in this documentation.
+pub trait TryIntoArguments<'cx>: private::TryIntoArgumentsInternal<'cx> {}
```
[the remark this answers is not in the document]
dherman: Added.

crates/neon/src/types_impl/function/mod.rs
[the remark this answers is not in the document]
dherman: Renamed `.apply()` to `.call()` and added `.construct()`.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1211,6 +1249,18 @@ impl JsFunction {
...
+            callee,
+            this: None,
+            args: smallvec![],
+        }
+    }
+}
+
 impl JsFunction {
     /// Create a [`CallOptions`](function::CallOptions) for calling this function.
     pub fn call_with<'a, C: Context<'a>>(&self, _cx: &C) -> CallOptions<'a> {
```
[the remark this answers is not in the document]
dherman: Done.

crates/neon/src/object/mod.rs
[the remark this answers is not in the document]
dherman: Replaced all of `.call()`, `.construct()`, `.call_with()`, and `.construct_with()` with `.bind()` in the doc comments.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1211,6 +1249,18 @@ impl JsFunction {
     }
 }
 
+impl JsFunction {
+    pub fn bind<'a, 'cx: 'a>(&self, cx: &'a mut Cx<'cx>) -> BindOptions<'a, 'cx> {
```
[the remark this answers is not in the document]
dherman: Added.

kjvalencik (approved):
The code and API look great! A couple small suggestions for the doc examples, but it's ready to go!

crates/neon/src/event/channel.rs
```diff
@@ -75,13 +75,12 @@ type Callback = Box<dyn FnOnce(sys::Env) + Send + 'static>;
 ///         // loop. This _will_ block the event loop while executing.
 ///         channel.send(move |mut cx| {
 ///             let callback = callback.into_inner(&mut cx);
-///             let this = cx.undefined();
-///             let args = vec![
-///                 cx.null().upcast::<JsValue>(),
-///                 cx.number(result).upcast(),
-///             ];
+///             let null = cx.null();
```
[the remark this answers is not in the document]
dherman: Nice! I'll go with `()` for now. Once you add that extractor API we can consider changing it from `()` to `Null`.

crates/neon/src/types_impl/mod.rs
```diff
@@ -1158,6 +1120,7 @@ impl JsFunction {
     /// Calls this function.
     ///
     /// **See also:** [`JsFunction::call_with`].
+    #[deprecated(since = "TBD", note = "use `JsFunction::bind` instead")]
```
kjvalencik, line 1123: I don't think this one should be deprecated. It's the low level function that doesn't have the overhead of casting and deserialization. I think we should keep it.

PR #1057, opened 2024-07-16
improve(neon): Use autoref specialization for JSON wrapping return values in exported functions

kjvalencik:
Currently, the `#[neon::export]` macro attempts to determine if the return value is a `Result` by parsing the signature and reading the name. This is only used in combination with the `json` attribute to determine how to properly wrap with the `Json` extract helper.

However, we can leverage [autoref specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md) for much more reliable handling. This is *very* important for supporting [async functions](https://github.com/neon-bindings/neon/pull/1055) because we can not easily see the output type in the signature.

This PR introduces new macro hidden trait types that can be used to extract the Rust value into a JS value. The `Json` variant is specialized for `Result<T, E>` while having a blanket implementation for `&T`. Since we have an owned `self`, autoref specialization will *prefer* our specialized type and fallback to the blanket impl. 

This works because the `Serialize` trait only needs to borrow the data. However, if owned data is needed, it's possible to extend this pattern with the technique `anyhow` uses--the specialized trait takes a reference and returns a sigil that then in turn takes the owned value.

### DIFF
--- a/crates/neon/src/macro_internal/mod.rs
+++ b/crates/neon/src/macro_internal/mod.rs
@@ -2,7 +2,12 @@
 
 pub use linkme;
 
-use crate::{context::ModuleContext, handle::Handle, result::NeonResult, types::JsValue};
+use crate::{
+    context::{Context, ModuleContext},
+    handle::Handle,
+    result::{JsResult, NeonResult},
+    types::{extract::TryIntoJs, JsValue},
+};
 
 type Export<'cx> = (&'static str, Handle<'cx, JsValue>);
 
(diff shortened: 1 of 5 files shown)


crates/neon/src/macro_internal/mod.rs
```diff
@@ -11,3 +16,55 @@ pub static EXPORTS: [for<'cx> fn(&mut ModuleContext<'cx>) -> NeonResult<Export<'
 
 #[linkme::distributed_slice]
 pub static MAIN: [for<'cx> fn(ModuleContext<'cx>) -> NeonResult<()>];
+
+pub trait NeonExportReturnValue<'cx> {
```
dherman, line 20: Let's put a comment either here or on the `impl … for &T` explaining how this leverages the Serde signature + autoref specialization.

dherman (changes requested):
Looks great, let's just leave a comment briefly explaining the technique.

crates/neon/src/macro_internal/mod.rs
kjvalencik: Done. I added some documentation describing how this works.

PR #1058, opened 2024-07-17
fix(neon): Fix a panic when borrowing an empty buffer or typed array

kjvalencik:
The borrow API currently panics if trying to borrow an empty buffer. This is because the pointer will be nil.

This PR adds guards for `length != 0`. Additionally, the ledger skips adding/removing empty buffers.


dherman (approved):
Good catch! LGTM.

PR #1059, opened 2024-07-24
feat(neon): With wrapper for TryIntoJs

kjvalencik:
Depends on https://github.com/neon-bindings/neon/pull/1048

### DIFF
--- a/crates/neon/src/types_impl/extract/with.rs
+++ b/crates/neon/src/types_impl/extract/with.rs
@@ -1,0 +1,58 @@
+use crate::{
+    context::{Context, Cx},
+    result::JsResult,
+    types::extract::TryIntoJs,
+};
+
+/// Wraps a closure that will be lazily evaluated when [`TryIntoJs::try_into_js`] is
+/// called.
+///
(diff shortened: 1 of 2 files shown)


crates/neon/src/types_impl/extract/with.rs
```diff
@@ -0,0 +1,58 @@
...
+///     let sum = nums.into_iter().sum::<f64>();
+///     let log = format!("sum took {} ms", start.elapsed().as_millis());
+///
+///     With(move |cx| -> NeonResult<_> {
+///          cx.global::<JsObject>("console")?
+///             .call_method_with(cx, "log")?
+///             .arg(cx.string(log))
+///             .exec(cx)?;
+///
+///         Ok(sum)
```
dherman, line 31: Is the indentation off here?

dherman (approved):
Looks great! I might've found one doc formatting nit, otherwise 🚢

crates/neon/src/types_impl/extract/with.rs
kjvalencik: Good eye! The `cx.global` line was off by exactly one space.

PR #1062, opened 2024-08-28
change(neon): Change TryIntoJs and TryFromJs accept Cx instead of being generic over Context

kjvalencik:
Being generic over `Context` offers no benefits over `Cx`. It is simpler to remove a generic argument.

* Every `Context` derefs to `Cx`
* `Cx` implements `Context`
* Reduces code bloat from monomorphization

We should double check that there's no other new APIs introduced in the pre-release that could benefit from this change. Unfortunately, we can't change existing APIs without a breaking change.

### DIFF
--- a/crates/neon/src/macro_internal/mod.rs
+++ b/crates/neon/src/macro_internal/mod.rs
@@ -3,7 +3,7 @@
 pub use linkme;
 
 use crate::{
-    context::{Context, ModuleContext},
+    context::{Cx, ModuleContext},
     handle::Handle,
     result::{JsResult, NeonResult},
     types::{extract::TryIntoJs, JsValue},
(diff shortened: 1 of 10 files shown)


crates/neon/src/macro_internal/mod.rs
```diff
@@ -19,19 +19,14 @@ pub static MAIN: [for<'cx> fn(ModuleContext<'cx>) -> NeonResult<()>];
 
 // Provides an identically named method to `NeonExportReturnJson` for easy swapping in macros
 pub trait NeonExportReturnValue<'cx> {
-    fn try_neon_export_return<C>(self, cx: &mut C) -> JsResult<'cx, JsValue>
-    where
-        C: Context<'cx>;
+    fn try_neon_export_return(self, cx: &mut Cx<'cx>) -> JsResult<'cx, JsValue>;
```
dherman, line 22: This is _such_ a quality of life improvement. Wow.

dherman (approved):
This looks great!

I noticed a handful of places in the API docs where there are examples of generic functions that take `C: Context<'cx>` ([1](https://github.com/neon-bindings/neon/blob/5e1b9f83032c1bc7725b9e6e6f17d1f479680002/crates/neon/src/thread/mod.rs#L16), [2](https://github.com/neon-bindings/neon/blob/5e1b9f83032c1bc7725b9e6e6f17d1f479680002/crates/neon/src/thread/mod.rs#L63), [3](http://app.slack.com/client/T0H96R4LR/D9U8LH8)). Should we change those to take a `Cx` now as well?

I also filed an issue for the website to [change the guides to use the new `Cx` type](https://github.com/neon-bindings/website/issues/258) once it gets published.

kjvalencik:
@dherman Great catch! Those are the only examples I saw that still unnecessarily used `Context`. I fixed them up. The great news is that 2 of the 3 are no longer generic at all! 🎉

PR #1064, opened 2024-09-05
Make build() fn's closure return a Result

akonradi-signal:
This is a pure refactor that makes the `build` function more idiomatic; no externally-facing types or functions are changed. The big benefit here is that it makes it possible to, in a future change, make the return type of some functions infallible, like `JsError::error`.


kjvalencik:
While I appreciate the refactor--cleaning up is always good for maintenance!--the `build` function is an artifact of when Neon supported two different backends: one that used NAN/C++ and one that used Node-API.

Since it shouldn't be used in anything new and should eventually be completely removed, I would prefer not to change it.

Thanks!

PR #1066, opened 2024-09-17
feat(neon): Add more TryIntoJs and TryFromJs implementations

kjvalencik:
While trying `#[neon::export]` out in another project, I found some trivial `TryFromJs` and `TryIntoJs` implementations that simplified code.

* `TryFromJs` for `Root<T>`. We can call `.root(cx)` when extracting
* `TryIntoJs` for `Box<T>` and `Arc<T>` smart pointers. This is especially useful if you want to return an `Arc<Vec<T>>` and convert it into a typed array.
* `TryIntoJs` for `&Vec<T>` and `&String`. We already had the slice versions, but it's nice to have these as well when combined with the smart pointers above.

I wanted to implement for `T` where `T: AsRef<U>` and `for<'a> &'a U: TryIntoJs<'cx>`, but this isn't possible since `U` wouldn't be bound by the implementation. It also conflicts with some specialized versions.

There's a bunch more that we _could_ add (e.g., `Box<str>`), but I'm trying to be a little conservative and not implement for a bunch of strange types that users don't need. These are all convenience implementations anyway. Users have `With` as an escape hatch.

### DIFF
--- a/crates/neon/src/types_impl/extract/try_into_js.rs
+++ b/crates/neon/src/types_impl/extract/try_into_js.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use crate::{
     context::{Context, Cx},
     handle::Handle,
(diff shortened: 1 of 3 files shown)


crates/neon/src/types_impl/extract/try_into_js.rs
```diff
@@ -63,6 +65,30 @@ where
...
 
+impl<'cx, T, V> TryIntoJs<'cx> for Box<T>
+where
+    for<'a> &'a T: TryIntoJs<'cx, Value = V>,
+    V: Value,
+{
+    type Value = V;
+
+    fn try_into_js(self, cx: &mut Cx<'cx>) -> JsResult<'cx, Self::Value> {
+        self.as_ref().try_into_js(cx)
```
kjvalencik, line 76: This is wrong. For `Box` we can take ownership of the value, so we shouldn't be using a reference.

dherman (approved):
This is super nice. 🚢  it!

PR #1073, opened 2024-09-27
feat(neon): Add neon::types::extract::Boxed for JsBox conversions

kjvalencik:
`JsBox<T>` can be extract in an exported function.

```rust
#[neon::export]
fn example(boxed: Handle<JsBox<String>>) {}
```

However, this only works for synchronous functions. If you need boxed data in an async function or task, it requires first cloning it in a sync function, which can be verbose.

```rust
#[neon::export]
fn example<'cx>(cx: &mut FunctionContext<'cx>, s: Handle<'cx, JsBox<String>>) -> Handle<'cx, JsPromise> {
    let s = String::clone(&s);
    
    cx.task(move || s).promise(|mut cx, s| cx.string(s))
}
```

However, we can trivially implement `TryFromJs` where `T: Clone`. It *must* be clone because `JsBox` only allows immutable references. This allows a much more ergonomic version.

```rust
#[neon::export(task)]
fn example(Boxed(s): Boxed<String>) -> String {
    s
}
```

Similarly, we can implement `TryIntoJs` for `Boxed` by calling `cx.boxed(..)`. The `TryIntoJs` implementation does not require `Clone` because we already have an owned copy.

### DIFF
--- a/crates/neon/src/types_impl/extract/boxed.rs
+++ b/crates/neon/src/types_impl/extract/boxed.rs
@@ -1,0 +1,79 @@
+use crate::{
+    context::{Context, Cx},
+    handle::Handle,
+    result::{JsResult, NeonResult, ResultExt},
+    types::{
+        extract::{private, TryFromJs, TryIntoJs, TypeExpected},
+        Finalize, JsBox, JsValue,
+    },
+};
+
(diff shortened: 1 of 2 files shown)


dherman (approved):
This is very nice, and the name is perfect, given the symmetry with `cx.boxed()`. :shipit:

PR #1074, opened 2024-09-30
Node stdlib

dherman:
Experimental API for convenient access to the Node stdlib.

Examples:

```rust
use neon::node::Node;

// ...

cx.console.log("hello, world!")?;

// ...

let versions = cx.process().versions()?;
println!("node version: {}", &versions.data().node);
```


crates/neon/src/node/console.rs
```diff
@@ -0,0 +1,84 @@
...
+        let function = ERROR.get_or_try_init(self.cx, |cx| {
+            let log: Handle<JsFunction> = console.get(cx, "error")?;
+            Ok(log.root(cx))
+        })?;
+        Ok(function.to_inner(self.cx))
+    }
+
+    // FIXME: when we land #1056, this can get simplified
+    // FIXME: msg should be a generic TryIntoJs
+    pub fn log(&mut self, msg: &str) -> NeonResult<()> {
```
kjvalencik, line 66: `TryIntoJs` will cover logging things that aren't string, but `console.log(..)` is also variadic. Thoughts on how we cover that? Could we use the `TryIntoArguments` trait?

It would be interesting to explore what a log _macro_ would look like. 🤔

dherman: Here’s what my thinking was, but tell me what you think:

I was thinking, you can get the most general behavior from the JS values, but the goal here is convenient and idiomatic Rust syntax for the most common functionality. So I figure the highest value is getting the `cx.console().log(x)` syntax, and the variadic use case is less important.

PR #1075, opened 2024-10-01
feat(neon-macros): Add `Channel` context for async functions and `this` support

kjvalencik:
Additionally, improves error messaging when incorrect types are provided. UI tests, via trybuild, are added to prevent regressions on errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,26 @@
+# Rust
 target
+rls*.log
+
+# Created by `trybuild` ui tests
 wip
 **/*~
+
+# Node
 **/node_modules
-**/.DS_Store
+npm-debug.log
+
+# JS build
 **/build
+**/dist
+cli/lib
+test/cli/lib
+
+# Neon build
 **/index.node
 **/artifacts.json
-cli/lib
-**/dist
 pkgs/create-neon/create-neon-test-project
 pkgs/create-neon/create-neon-manual-test-project
-test/cli/lib
-npm-debug.log
-rls*.log
+
+# System
+**/.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 target
+wip
 **/*~
 **/node_modules
 **/.DS_Store
--- a/crates/neon-macros/src/export/function/meta.rs
+++ b/crates/neon-macros/src/export/function/meta.rs
@@ -4,6 +4,7 @@ pub(crate) struct Meta {
     pub(super) name: Option<syn::LitStr>,
     pub(super) json: bool,
     pub(super) context: bool,
+    pub(super) this: bool,
 }
 
 #[derive(Default)]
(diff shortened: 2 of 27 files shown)


test/ui/tests/fail/need-borrowed-context.stderr
```diff
@@ -0,0 +1,35 @@
+error: Context must be a `&mut` reference.
```
dherman, line 1: How brittle will these patterns be from one version of the compiler to another?

test/ui/tests/fail/wrong-context.stderr
```diff
@@ -0,0 +1,77 @@
+error: Expected `&mut Cx` instead of `Channel`.
+ --> tests/fail/wrong-context.rs:2:22
+  |
+2 | fn sync_channel(_ch: Channel) {}
+  |                      ^^^^^^^
+
+error: Expected `&mut Cx` instead of a `Channel` reference.
```
dherman, line 7: I wonder if we should consider creating Neon compiler error codes similar to rustc and tsc. Just something we could think about for a future PR.

crates/neon-macros/src/export/function/mod.rs
```diff
@@ -130,53 +145,244 @@ pub(super) fn export(meta: meta::Meta, input: syn::ItemFn) -> proc_macro::TokenS
...
-        _ => return None,
+    match (has_context, has_this) {
+        (true, true) => n - 2,
+        (false, false) => n,
+        _ => n - 1,
+    }
+}
+
+// Generate the context extraction and argument for the function
+fn context_parse(
```
dherman, line 160: Some of this parsing and checking logic feels a bit intermingled. That might just be a side effect of the code review UI, not sure.

A couple ideas you could consider for modularizing it a bit:

- Separating parsing logic into a helper module.
- Make the parsers just return an error type with the data needed for the validation, and separate logic for taking the error information and producing the syntax error. Not sure if this doesn't end up changing much.

Just some ideas.

dherman (approved):
This is awesome, and the tests are impressively thorough. `Boxed(this)` is especially powerful for custom native classes.

I left a couple detailed suggestions but they're up to you.

.gitignore
```diff
@@ -1,4 +1,5 @@
 target
+wip
```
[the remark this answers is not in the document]
kjvalencik: It's important. `trybuild` writes to this directory with the `stderr` output.

crates/neon-macros/src/export/function/meta.rs
```diff
@@ -28,33 +29,29 @@ impl Meta {
...
     }
 
-    fn force_context(&mut self, meta: syn::meta::ParseNestedMeta) -> syn::Result<()> {
-        match self.kind {
-            Kind::Normal | Kind::AsyncFn => {}
-            Kind::Async => return Err(meta.error(super::ASYNC_CX_ERROR)),
-            Kind::Task => return Err(meta.error(super::TASK_CX_ERROR)),
-        }
-
+    fn force_context(&mut self, _meta: syn::meta::ParseNestedMeta) -> syn::Result<()> {
```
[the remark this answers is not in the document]
kjvalencik: Some of these are fallible. I kept the infallible ones as returning `Result<()>` to make the match code consistent.

test/ui/tests/fail/wrong-context.stderr
kjvalencik: Oooh, I like that idea. That would give us something to easily cross-reference to get more details. I'll file that for future improvement.

crates/neon-macros/src/export/function/mod.rs
kjvalencik: I hear you, but I'm concerned that separating the logic _feels_ better but actually makes the code more difficult to understand and change.

These methods only have a *two* correct branches:
* Optional argument not provided
* Optional argument present

All the other branches are for custom error messages on common mistakes. If we separate the logic, we end up with a bunch of unique variants that only serve to map to an error message. Adding or improving error messages becomes a three step process of add enum variant, parse enum variant, match enum variant instead of a single step.

.gitignore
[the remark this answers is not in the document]
dherman: Just a short one would be good, thanks.

crates/neon-macros/src/export/function/meta.rs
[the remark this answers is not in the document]
dherman: Yeah, I realized after I reviewed the others. Up to you.

test/ui/tests/fail/need-borrowed-context.stderr
dherman: Agreed!

PR #1076, opened 2024-10-02
fix(neon): Require `TryFromJs::Error` to implement `TryIntoJs`

kjvalencik:
Additionally:
* Replace `TryFromJs::from_js` implementations with a default
* Add a `json::Error` wrapper to hide implementation details

This will be import for implementing `TryFromJs` on `either::Either`. Without this requirement, we would either need a `cx.try_catch(..)` or to discard the `Left` error in the `TryFromJs::from_js` implementation.


crates/neon/src/types_impl/extract/mod.rs
```diff
@@ -99,24 +99,27 @@
...
 
 #[cfg(feature = "serde")]
 #[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
 pub use self::json::Json;
 
-mod error;
 #[cfg(feature = "serde")]
-mod json;
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
+pub mod json;
```
kjvalencik, line 120: This is public now. It felt better to expose the JSON error wrapper as `json::Error` than to move it up a level as `JsonError`.

I kept the `Json` extractor double exposed at `extract::Json` so that it can live next to the other extractors. Most users won't need to dig into the deeper module for the error.

crates/neon/src/types_impl/extract/error.rs
```diff
@@ -1,13 +1,60 @@
...
+    }
+}
+
+impl<T: Value> private::Sealed for TypeExpected<T> {}
+
+impl<'cx> TryIntoJs<'cx> for Infallible {
+    type Value = JsValue;
+
+    fn try_into_js(self, _: &mut Cx<'cx>) -> JsResult<'cx, Self::Value> {
+        unreachable!()
```
kjvalencik, line 52: I'll be very glad when the `never` / `!` type stabilizes, but for now this is the best we can do to make the trait bounds happy. It's impossible to construct `Infallible` in safe rust so this should never happen.

crates/neon/src/types_impl/extract/json.rs
```diff
@@ -73,17 +91,15 @@ impl<'cx, T> TryFromJs<'cx> for Json<T>
 where
     for<'de> T: serde::de::Deserialize<'de>,
 {
-    type Error = serde_json::Error;
```
kjvalencik, line 76: This was a mistake. It tightly couples the implementation to `serde_json`. In the future, if Node-API gets optimized object construction, we change the implementation to direct transcoding. Hiding away the error details gives us the flexibility.

test/napi/src/js/objects.rs
```diff
@@ -113,12 +113,12 @@ pub fn call_methods_with_prop(mut cx: FunctionContext) -> JsResult<JsString> {
...
         .arg("Wonder Woman")?
-        .call()?;
+        .exec()?;
     obj.prop(&mut cx, "toString").bind()?.call()
 }
 
 pub fn call_non_method_with_prop(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     let obj: Handle<JsObject> = cx.argument::<JsObject>(0)?;
-    obj.prop(&mut cx, "number").bind()?.call()?;
+    obj.prop(&mut cx, "number").bind()?.exec()?;
```
kjvalencik, line 122: These are some fly-by lint fixes. `.call()` requires a return generic. Since it's never used, there's an ambiguity where the compiler treats `()` and `!` as equivalent. This will be a compiler error in 2025 edition.

dherman (approved):
Nice improvements! Great catch on hiding the Serde implementation.

PR #1078, opened 2024-10-03
feat(neon): Implement TryIntoJs/TryFromJs for either::Either

kjvalencik:
This implements the extract traits for `either::Either`. I chose *not* to feature flag it because `either` is a very small crate with zero dependencies.

The `Ok` implementation paths were very clear. However, there isn't an _obviously_ correct implementation for a failed `TryFromJs`.

We have two errors, both `L::Error` and `R::Error`. This PR does the following:
* Creates a new `Error` that prints the type name of `L` and `R` saying that they failed
* Attaches `left` and `right` keys to the error object with their `TryIntoJs` implementation

Another option would be to return only *one* of the errors. Unfortunately, we can't attach one error to the other because it _might not_ be an `Object`.

What's the best thing to do for this?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,9 +157,9 @@ checksum = "10936778145f3bea71fd9bf61332cce28c28e96a380714f7ab34838b80733fd6"
 
 [[package]]
 name = "either"
-version = "1.9.0"
+version = "1.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


dherman:
> What's the best thing to do for this?

I agree with what you did. Two errors occurred, so neither should be dropped; the caller needs to decide what to do and should have full access to both. It kind of reminds me of how exception cause chaining preserves a history for introspection.

PR #1080, opened 2024-10-11
Add neon::types::JsMap

### LINKED ISSUE #1072: Expose Map
The [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) object type seems to be missing from the bindings.

Given it requires the use of `new` to be instantiated (just like e.g. Promise), I guess it should be exposed in those bindings.

I'm currently trying to create an instance of `Map` with the bindings, but it seems rather complex since using new requires to to use the napi unsafe api... is there something I've missed here ?

My current solution is to use the eval function to create the Map, which is a bit wasteful ^^
```rust
        let new_map_code = cx.string("(function() { return new Map(); })");
        let js_map = neon::reflect::eval(cx, new_map_code)?.downcast_or_throw::<JsObject, _>(cx)?;
```

So would you accept a PR to add Map in the bindings ?

touilleMan:
close #1072 

This PR is not done yet, but I'd like some feedback on the implementation:

- `Map.groupBy` is currently broken (this is the added test that is not passing), the issue comes from a `failed to downcast any to object` error when trying to downcast `Map.groupBy` into a `JsFunction`. I guess this is related to the fact groupBy is a static method, but I don't know how to deal with it (maybe entirely bypassing `JsFunction` and using lower level code instead ?).
- What should be the policy for return types ? For instance when a method is expected (according the MDN documentation) to return `undefined` should we check the actual returned value is `undefined` then return an empty tuple to stick with the Rust way, or should we just avoid any guesswork and return a `JsValue` no matter what ?
- I'm unsure if and how to implement `Map[Symbol.species]` and `Map.prototype[Symbol.iterator]()`

(FYI I've put TODO about those points in the code)


crates/neon/src/sys/bindings/functions.rs
```diff
@@ -51,6 +51,9 @@ mod napi1 {
 
             fn close_handle_scope(env: Env, scope: HandleScope) -> Status;
 
+            fn instanceof(env: Env, object: Value, constructor: Value, result: *mut bool)
```
kjvalencik, line 54: This is awesome! Thanks for adding it. I think we should add it as a method on the `Value` trait.

```rust
trait Value {
    fn instance_of<V: Value>(&self, cx: &mut Cx, ctor: &V) -> bool {
        /* ... */
    }
}
```

test/napi/src/lib.rs
```diff
@@ -233,6 +234,24 @@ fn main(mut cx: ModuleContext) -> NeonResult<()> {
     cx.export_function("return_js_array_with_string", return_js_array_with_string)?;
     cx.export_function("read_js_array", read_js_array)?;
 
+    cx.export_function("return_js_map", return_js_map)?;
```
kjvalencik, line 237: Can we use the `#[neon::export]` macro for all of the new tests?

crates/neon/src/sys/tag.rs
```diff
@@ -137,3 +139,37 @@ pub unsafe fn check_object_type_tag(env: Env, object: Local, tag: &super::TypeTa
 pub unsafe fn is_bigint(env: Env, val: Local) -> bool {
     is_type(env, val, napi::ValueType::BigInt)
 }
+
+pub unsafe fn is_map(env: Env, val: Local) -> bool {
+    let global_object = unsafe {
+        let mut out: super::raw::Local = std::mem::zeroed();
```
kjvalencik, line 145: Can we use `MaybeUnit::unit()` instead? The existing patterns that are `mem::zeroed()` are an old style and less desirable.

crates/neon/src/sys/tag.rs
```diff
@@ -137,3 +139,37 @@ pub unsafe fn check_object_type_tag(env: Env, object: Local, tag: &super::TypeTa
 pub unsafe fn is_bigint(env: Env, val: Local) -> bool {
     is_type(env, val, napi::ValueType::BigInt)
 }
+
+pub unsafe fn is_map(env: Env, val: Local) -> bool {
```
kjvalencik, line 143: In my opinion, there isn't a great reason to put this in `sys`. A lot of the `sys` patterns like `sys::tag` go back to when Neon supported a C++ backend (prior to Node-API).

The current conventions are to put this code alongside the safe implementation. This also makes it easier to use _other_ safe wrappers instead of falling back to unsafe. Less unsafe code is needed. Unfortunately, the `ValueInternal` trait uses pointers, so _some_ unsafe is needed, but only for getting a `Context` and `Handle`.

crates/neon/src/sys/tag.rs
```diff
@@ -137,3 +139,37 @@ pub unsafe fn check_object_type_tag(env: Env, object: Local, tag: &super::TypeTa
...
+        let mut out: super::raw::Local = std::mem::zeroed();
+        let literal = b"Map";
+        assert_eq!(
+            super::string::new(&mut out, env, literal.as_ptr(), literal.len() as _),
+            true
+        );
+        out
+    };
+
+    let map_constructor = {
```
kjvalencik, line 160: The steps to get this global can be somewhat expensive. We can make wrappers that cache the pointer in a `LocalKeyRoot<_>>`.

Take a look at the `Json` example for how it caches `JSON.stringify` and `JSON.parse`.

https://github.com/neon-bindings/neon/blob/1efcd4ea0890dfe55a5d4cee9661aaf232539406/crates/neon/src/types_impl/extract/json.rs#L46-L53

crates/neon/src/types_impl/map.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn new<'cx, C>(cx: &mut C) -> NeonResult<Handle<'cx, Self>>
+    where
+        C: Context<'cx>,
+    {
+        let map = cx
+            .global::<JsFunction>("Map")?
+            .construct_with(cx)
+            .apply::<JsObject, _>(cx)?;
+
+        Ok(map.downcast_or_throw(cx)?)
```
kjvalencik, line 32: FYI, downcasting *is* necessary. Since it's possible to override `global.Map`, we don't know that it's actually the built-in. If someone were to override it, the constructor *might* return a different type.

```js
global.Map = function Map() {
    return new Array();
};

new Map() instanceof Map
// false
```

crates/neon/src/types_impl/map.rs
```diff
@@ -0,0 +1,163 @@
...
+    // 1. Stick to the JS and return the `undefined` (this is what we do now)
+    // 2. Check we get an `undefined` and return `Ok(())`
+    // 3. Just discard the return value and return `Ok(())`
+    // Solutions 2 & 3 are more user-friendly, but make more assumptions (though it
+    // should be fine given `Map` is not expected to be overridden ?).
+    pub fn clear<'a, C: Context<'a>>(&self, cx: &mut C) -> NeonResult<Handle<'a, JsUndefined>> {
+        Object::call_method_with(self, cx, "clear")?.apply(cx)
+    }
+
+    pub fn delete<'a, C: Context<'a>, K: Value>(
```
kjvalencik, line 50: If we are going to have wrappers for the built in methods, we should try to make them feel more Rust-y. Currently, this requires calling like:

```rust
ket k = cx.string("key");
map.delete(&mut cx, k)?;
```

Ideally, I could call:

```rust
map.delete(&mut cx, "key")?;
```

My suggestion is to accept `K: TryIntoJs<'cx>` instead of a handle. Additionally, with the addition of `Cx` there isn't a compelling reason to be generic over `Context`.

```rust
impl JsMap {
    pub fn delete<'cx, K>(cx: &mut Cx<'cx>, k: K) -> NeonResult<bool>
    where
        K: TryIntoJs<'cx>,
    {
        self.prop(&mut cx, "delete").bind()?.arg(k)?.call()
    }
}
```

crates/neon/src/types_impl/map.rs
```diff
@@ -0,0 +1,163 @@
...
+            .apply::<JsObject, _>(cx)?;
+
+        Ok(map.downcast_or_throw(cx)?)
+    }
+
+    pub fn size<'a, C: Context<'a>>(&self, cx: &mut C) -> NeonResult<Handle<'a, JsNumber>> {
+        Object::get(self, cx, "size")
+    }
+
+    // TODO: is the return type important here ?
```
kjvalencik, line 39: It's not important here. I would just return `NeonResult<()>` and skip any type checks.

crates/neon/src/types_impl/map.rs
```diff
@@ -0,0 +1,163 @@
...
+    {
+        let map = cx
+            .global::<JsFunction>("Map")?
+            .construct_with(cx)
+            .apply::<JsObject, _>(cx)?;
+
+        Ok(map.downcast_or_throw(cx)?)
+    }
+
+    pub fn size<'a, C: Context<'a>>(&self, cx: &mut C) -> NeonResult<Handle<'a, JsNumber>> {
```
kjvalencik, line 35: Similar to other methods, I think it's more valuable to immediately extract the value. The user will almost always need the `f64` if they are calling this and it's more ergonomic in addition to being one fewer Node-API call.

```rust
impl JsMap {
    pub fn size<'cx>(&self, cx: &mut Cx<'cx>) -> NeonResult<f64> {
        self.prop("size").get()
    }
}
```

crates/neon/src/types_impl/map.rs
```diff
@@ -0,0 +1,163 @@
...
+        // TODO: This is broken and leads to a `failed to downcast any to object` error
+        // when trying to downcast `Map.groupBy` into a `JsFunction`...
+        cx.global::<JsObject>("Map")?
+            .call_method_with(cx, "groupBy")?
+            .arg(elements)
+            .arg(cb)
+            .apply(cx)
+    }
+
+    // TODO: should we implementd those as well ?
```
kjvalencik, line 130: We would need to implement `Symbol` first so they can be skipped for now.

kjvalencik (commented):
Thanks so much for this contribution @touilleMan! It'll be really great to have this type in here.

In terms of the design, I would like to have more ergonomic wrappers (i.e., that take Rust types). Unfortunately, this makes the design space a lot more complicated.

What is the minimum set of features you need for your use case? Maybe we could initially start with a smaller version of `Map` that doesn't expose everything and we could iterate in future PRs.

crates/neon/src/types_impl/map.rs
touilleMan: So you mean `Cx` is the new way of doing things and `C: Context<'a>` the old one ?
I was wondering why there was those two and wasn't sure if `C: Context<'a>` wasn't present for a specific need regarding genericity...

kjvalencik: `Context` is the old way and `Cx` is the new way. `Cx` was added to support places where we *can't* be generic because of limitations of GAT (e.g., `Fn*` traits).

Sometimes it's still useful to be generic over `Context` if you hand the `Context` back to user controlled code because `FunctionContext` has additional capabilities. However, if the implementation is fully generic over `Context`, `Cx` makes for a much simpler function signature.

From a developer experience perspective, they are equivalent because all contexts implement `RefMut<Cx>`.

test/napi/src/lib.rs
touilleMan: Yes ! that's much cleaner this way \o/

crates/neon/src/sys/tag.rs
kjvalencik: I opened https://github.com/neon-bindings/neon/pull/1081 to allow implementing `ValueInternal::is_typeof` without any unsafe.

kjvalencik:
> My implementation of sys::tag::is_map is very low level and clumsy, I feel like their should be a cleaner way to implemented it ^^

Unfortunately, `Map` isn't in Node-API so it does require the clunky grabbing `Map` off of the global object and instance checking. This is true about many of the types that are implemented is globals and not built into the language.

I do think we can clean it up a bit by implementing with the safe API wrappers and implementing `LocalKey<Root<_>>` cached lookups.

crates/neon/src/sys/bindings/functions.rs
kjvalencik: Would you mind opening a separate PR to implement this? It's a great feature and we can get it merged pretty quick.

I'm happy to pick it up if you prefer.

touilleMan: Done, I've also added a dedicated test

crates/neon/src/sys/tag.rs
touilleMan: So you mean I should entirely remove `sys::tag::is_map` and instead put it content directly inside `impl private::ValueInternal for JsMap`'s `is_typeof` ?

kjvalencik: Yep, exactly!

touilleMan:
@kjvalencik Thanks for you precise and quick feedback, it's a very motivating to have such nice interactions when submitting a PR :tada: 

I've pushed a new comment that should have addressed all your comments (don't hesitate to tell me if you think more changes are needed ^^)

(Once the PR is ready for merging, I will squash the commits together to keep things clean)

> What is the minimum set of features you need for your use case? Maybe we could initially start with a smaller version of Map that doesn't expose everything and we could iterate in future PRs.

I'm doing simple conversion between [Rust `HashMap` and Javascript `Map`](https://github.com/Scille/parsec-cloud/blob/6c1d3be71b7f440fb6ef7b5977afb3ea91ab3fd5/bindings/generator/templates/binding_electron_meths.rs.j2#L324-L332). So my needs are really minimal: just being able to create a map and get/set items.

> In terms of the design, I would like to have more ergonomic wrappers (i.e., that take Rust types). Unfortunately, this makes the design space a lot more complicated.

What do you have in mind ? Having `keys`/`values` methods returning a wrapper over the javascript iterator object ?

touilleMan:
Also regarding `Map.groupBy`, I've added the following debug code in my `JSMap::group_by` method:

```rust
        let xx = cx.string("console.log('============~~~~', Map.groupBy)");
        crate::reflect::eval(cx, xx)?;
```

And this prints `undefined`, so I guess this explain why `JSMap::group_by` has this `failed to downcast any to object` error.
So I guess the issue comes from the version of Node I'm using (i.e. 18.17.0) which is too old to have `Map.groupBy` ([as it seems a rather recent addition to the standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy#browser_compatibility))

What should be done then ? I would say doing `self.method(cx, "groupBy")?` in the codebase should detect that the method is not available and return the correct not found error, however it seems to accept the `undefined` value and the error only occurs later when a downcast is tried (which leads to the more cryptic error I was initial talking about...)

touilleMan:
@kjvalencik friendly ping ;-)

PR #1081, opened 2024-10-11
chore(neon): Improve ValueInternal safety

kjvalencik:
`ValueInternal::is_typeof` and `ValueInternal::downcast` accept raw `Env`, but can trivially accept `Cx` instead.

This will allow for global type wrappers (e.g., https://github.com/neon-bindings/neon/pull/1080) that can type check without any unsafe code.

### DIFF
--- a/crates/neon/src/types_impl/private.rs
+++ b/crates/neon/src/types_impl/private.rs
@@ -1,7 +1,10 @@
 use std::{ffi::c_void, mem::MaybeUninit};
 
 use crate::{
-    context::{internal::Env, Context},
+    context::{
+        internal::{ContextInternal, Env},
+        Context, Cx,
+    },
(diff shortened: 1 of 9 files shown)


crates/neon/src/types_impl/private.rs
```diff
@@ -31,13 +34,13 @@ pub(crate) unsafe fn prepare_call<'a, 'b, C: Context<'a>>(
 pub trait ValueInternal: TransparentNoCopyWrapper + 'static {
```
kjvalencik, line 34: `ValueInternal::from_local` could accept safe wrappers, but it would need to remain `unsafe` since it doesn't perform type checking.

dherman (approved):
Nice cleanup 👍

PR #1084, opened 2024-11-22
feat(neon-macros): Convert snake_case to camelCase when exporting functions

kjvalencik:
Rust uses a `snake_case` convention for functions while JavaScript `camelCase`. This changes makes the default behavior of `neon::export` convert to `camelCase` to follow JavaScript naming conventions.

If the user does not want this behavior the `name = ".."` override may be used to explicitly provide a name.

No mapping is provided for `const` or `static` exported globals since both Rust and JS follow a `YELLING_SNAKE_CASE` convention.

### DIFF
--- a/crates/neon-macros/src/export/function/mod.rs
+++ b/crates/neon-macros/src/export/function/mod.rs
@@ -111,7 +111,10 @@ pub(super) fn export(meta: meta::Meta, input: syn::ItemFn) -> proc_macro::TokenS
     let export_name = meta
         .name
         .map(|name| quote::quote!(#name))
-        .unwrap_or_else(|| quote::quote!(stringify!(#name)));
+        .unwrap_or_else(|| {
+            let name = to_camel_case(&name.to_string());
+            quote::quote!(#name)
+        });
 
     // Generate the function that is registered to create the function on addon initialization.
     // Braces are included to prevent names from polluting user code.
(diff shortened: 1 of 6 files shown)


crates/neon-macros/src/export/function/mod.rs
```diff
@@ -145,6 +148,28 @@ pub(super) fn export(meta: meta::Meta, input: syn::ItemFn) -> proc_macro::TokenS
     .into()
 }
 
+fn to_camel_case(name: &str) -> String {
+    let mut should_upper = false;
+    let mut out = String::with_capacity(name.len());
```
kjvalencik, line 153: N.B.: This is a _reasonable_ heuristic to avoid allocations. In most cases it will over allocate by a few characters and in rare circumstances where `to_uppercase()` produces longer chars it will need to grow (TBH, I'm not even sure if this is possible with valid Rust identifiers).

test/napi/src/js/export.rs
```diff
@@ -27,7 +27,7 @@ fn simple_add(a: f64, b: f64) -> f64 {
 }
 
 #[neon::export(name = "renamedAdd")]
-fn renamed_add(a: f64, b: f64) -> f64 {
+fn rs_renamed_add(a: f64, b: f64) -> f64 {
```
kjvalencik, line 30: I added `rs_` prefixes here because the tests were useless after this change (the override matched the default).

crates/neon-macros/src/export/function/mod.rs
```diff
@@ -145,6 +148,28 @@ pub(super) fn export(meta: meta::Meta, input: syn::ItemFn) -> proc_macro::TokenS
     .into()
 }
 
+fn to_camel_case(name: &str) -> String {
```
dherman, line 151: We could use the [regex](https://crates.io/crates/regex) crate for this. Granted, we're pretty conservative about taking dependencies, but it's a well-maintained crate, and it's just a compile-time dependency. ([example playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=670615e3baa025f239d6c4c77729f2f1))

```rust
impl Replacer for CamelReplacer {
    fn replace_append(&mut self, caps: &Captures<'_>, dst: &mut String) {
        dst.push_str(&caps[1].to_uppercase());
    }
}

let capture_the_snake = Regex::new(r"_(\p{Lowercase_Letter})").unwrap();
...
capture_the_snake.replace_all(name, CamelReplacer)
```

Even more straightforwardly, there's the [convert_case](https://crates.io/crates/convert_case) crate:

```rust
name.to_case(Case::Camel)
```

dherman: Yeah I think it makes the most sense to use `convert_case`. It's a popular library, used by the AWS SDK and even by napi.rs for the same purpose as what we're doing here. And it'll handle edge cases (e.g. what if the identifier starts with an underscore) that we might not think of.

crates/neon-macros/src/export/function/mod.rs
```diff
@@ -386,3 +411,19 @@ fn check_this(opts: &meta::Meta, sig: &syn::Signature, has_context: bool) -> boo
         _ => false,
     }
 }
+
+#[cfg(test)]
```
dherman, line 415: You can drop all these tests if we use `convert_case`.

dherman (changes requested):
I like this, I think it's a better default. For the implementation, I'm proposing that we use [convert_case](https://crates.io/crates/convert_case). See the comments inline for my rationale.

crates/neon-macros/src/export/function/mod.rs
kjvalencik: I considered this and checked our existing tree to see if we already had any dependencies that did it. Since we did not, I opted not to use a dependency since it's only 15 lines of code.

You make a good point about edge cases. This implementation considers them and handles the properly. However, the definition of "proper" is a judgement call. For example, I chose to remove all underscores, but another library may choose to keep leading underscores or duplicates.

It might be worth using a library, not to save code, but to have confidence that we are consistent with the ecosystem. I do have one concern with using a library. This conversion is part of our **public** API and a change is a semver breaking change. Using a crate means we need to be careful when bumping dependency versions since a major version could change the conversion.

Do you think that semver risk is worth using a crate?

dherman:
We discussed this on Slack and came to the following conclusions:
- `convert_case` doesn't handle things like leading and trailing underscores
- Using an external crate is a bit risky since this is part of the public behavior of Neon
- Even using regex would be a bit complex so it ends up being fine to implement this by hand
- We want to be pretty conservative about what conventions we recognize; if they have nonstandard names we will just bail and not transform the name

In the end this means:
- Preserve leading and trailing underscores
- Only capitalize the first character after a non-leading underscore
- Bail if we see two underscores in a row in between words or a capital letter

dherman (approved):
This looks great. Ship it!

PR #1085, opened 2024-11-22
chore(neon): fix clippy lints

dherman:
This PR fixes the Clippy lints exposed by the awesome refactoring done in #1065.

### DIFF
--- a/crates/neon/src/sys/async_work.rs
+++ b/crates/neon/src/sys/async_work.rs
@@ -74,7 +74,7 @@ pub unsafe fn schedule<I, O, D>(
         Ok(()) => {}
         status => {
             // If queueing failed, delete the work to prevent a leak
-            napi::delete_async_work(env, *work).unwrap();
+            let _ = napi::delete_async_work(env, *work);
             status.unwrap()
         }
     }
@@ -151,7 +151,7 @@ unsafe extern "C" fn call_complete<I, O, D>(env: Env, status: napi::Status, data
     } = *Box::<Data<I, O, D>>::from_raw(data.cast());
 
     if napi::delete_async_work(env, work).is_err() {
-        panic!("failed to delete async work");
+        panic!("Failed to delete async work");
     }
 
     BOUNDARY.catch_failure(env, None, move |env| {
--- a/crates/neon/src/sys/async_work.rs
+++ b/crates/neon/src/sys/async_work.rs
@@ -150,9 +150,7 @@ unsafe extern "C" fn call_complete<I, O, D>(env: Env, status: napi::Status, data
         ..
     } = *Box::<Data<I, O, D>>::from_raw(data.cast());
 
-    if napi::delete_async_work(env, work).is_err() {
-        panic!("Failed to delete async work");
-    }
+    debug_assert_eq!(napi::delete_async_work(env, work), Ok(()));
 
     BOUNDARY.catch_failure(env, None, move |env| {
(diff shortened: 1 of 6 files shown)


crates/neon/src/sys/async_work.rs
```diff
@@ -74,7 +74,7 @@ pub unsafe fn schedule<I, O, D>(
         Ok(()) => {}
         status => {
             // If queueing failed, delete the work to prevent a leak
-            napi::delete_async_work(env, *work);
+            napi::delete_async_work(env, *work).unwrap();
```
kjvalencik, line 77: Since this call isn't directly from the user and we are unwrapping a line later, I think it's best to ignore this return value.

We could consider adding a log warning, but I think it's okay for this to be best effort.

crates/neon/src/sys/object.rs
```diff
@@ -7,7 +7,7 @@ use super::{
 
 /// Mutates the `out` argument to refer to a `napi_value` containing a newly created JavaScript Object.
 pub unsafe fn new(out: &mut Local, env: Env) {
-    napi::create_object(env, out as *mut _);
+    assert_eq!(napi::create_object(env, out as *mut _), Ok(()));
```
kjvalencik, line 10: Why not `unwrap`?

crates/neon/src/sys/tsfn.rs
```diff
@@ -182,9 +182,12 @@ impl<T> Drop for ThreadsafeFunction<T> {
         }
 
         unsafe {
-            napi::release_threadsafe_function(
-                self.tsfn.0,
-                napi::ThreadsafeFunctionReleaseMode::Release,
+            assert_eq!(
```
kjvalencik, line 185: This is internal neon code and only prevents a memory leak. I think it would be better to ignore this error, possibly warn when a log, rather than panic.

crates/neon/src/sys/object.rs
```diff
@@ -113,20 +112,14 @@ pub unsafe fn set_string(
...
+        return false;
     }
 
-    match napi::set_property(env, object, key_val.assume_init(), val) {
-        Err(_) => {
-            *out = false;
-            return false;
-        }
-        Ok(()) => (),
+    if napi::set_property(env, object, key_val.assume_init(), val).is_err() {
```
kjvalencik, line 120: I think a lot of these might be subtly incorrect. The legacy backend assumed exception in many places when `false` was returned, but that might not be the case.

The correct thing to do might be to return `false` on exception and otherwise panic.

kjvalencik (changes requested):
@dherman, it's awesome to get these fixed! I think a couple of the memory leak releases shouldn't panic.

Maybe a good compromise is to panic in debug mode and ignore in release mode?

Additionally, I haven't traced through, but the `return false` on any error lines seem incorrect. Those should probably only return false on exception.

crates/neon/src/sys/async_work.rs
```diff
@@ -150,7 +150,9 @@ unsafe extern "C" fn call_complete<I, O, D>(env: Env, status: napi::Status, data
         ..
     } = *Box::<Data<I, O, D>>::from_raw(data.cast());
 
-    napi::delete_async_work(env, work);
+    if napi::delete_async_work(env, work).is_err() {
```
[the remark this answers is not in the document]
kjvalencik: I could go either way. We've already unboxed the closure, so we aren't going to leak memory by panicking here. However, it might be a bit more consistent to debug assert.

PR #1086, opened 2024-12-17
fix(neon): Relax lifetime constraints on `With` and provide helper for forcing HRTB with JS values

kjvalencik:
`With` is primarily intended as a helper for providing a **thunk** executed on the main thread after an asynchronous operation. However, it's not currently possible to return JavaScript values due to unnecessarily restrictive lifetimes.

```rust
// DOES NOT COMPILE
use neon::types::extract::{TryIntoJs, With};

#[neon::export(task)]
fn example() -> impl for<'cx> TryIntoJs<'cx> {
    With(|cx| cx.number(42.0))
}
```

These changes relax thoe bounds, but unfortunately causes Rust to no longer correctly infer the correct HRTB. As a workaround, an `extract::with` helper is provided.

```rust
use neon::types::extract::{self, TryIntoJs};

// Compiles :)
#[neon::export(task)]
fn example() -> impl for<'cx> TryIntoJs<'cx> {
    extract::with(|cx| Ok(cx.number(42.0)))
}
```

I could not find a way to make a helper that works with *both* `JsResult` and Rust data *without* a lifetime. Since someone returning Rust data always has the option to call `.try_into_js(cx)`, I chose to keep the API simple by *only* providing the one helper.

## Open Questions

* [ ] Is it possible to implement a helper that supports generic return types of `O: TryIntoJs<'cx>` that aren't `'static`?
* [ ] Should we make `With` private? The `with` type would return an opaque `impl for<'cx> TryIntoJs<'cx>`. This _could_ simplify the API.

### DIFF
--- a/crates/neon/src/macros.rs
+++ b/crates/neon/src/macros.rs
@@ -201,15 +201,15 @@ pub use neon_macros::main;
 /// ```
 /// # #[cfg(all(feature = "napi-6", feature = "futures"))]
 /// # {
-/// # use neon::types::extract::{TryIntoJs, With};
+/// # use neon::types::extract::{self, TryIntoJs};
 /// #[neon::export]
 /// async fn add(a: f64, b: f64) -> impl for<'cx> TryIntoJs<'cx> {
 ///     let sum = a + b;
 ///
-///     With(move |_cx| {
+///     extract::with(move |cx| {
 ///         println!("Hello from the JavaScript main thread!");
 ///
-///         sum
+///         sum.try_into_js(cx)
 ///     })
 /// }
 /// # }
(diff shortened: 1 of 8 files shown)


dherman:
>  Is it possible to implement a helper that supports generic return types of `O: TryIntoJs<'cx>` that aren't `'static`?

I've pinged @nikomatsakis to see if he can take a minute to see if we've correctly understood the lifetime issues.

> Should we make `With` private? The with type would return an opaque `impl for<'cx> TryIntoJs<'cx>`. This could simplify the API.

I like this idea. It also feels more future-proof.

dherman (approved):
I think this is the right call. Implementation looks good, no notes :)

PR #1089, opened 2025-01-22
feat(neon): Add extractor for TypedArray

kjvalencik:
This removes the `TryIntoJs` implementations for `Arc` and `Box` to free them up for implicitly `Boxed` implementations.

In order to maintain the ability to return borrowed buffers (e.g., `Arc<Vec<u8>>`), I've added extractors for typed arrays. It requires slightly more type noise (e.g., `Uint8Array<Arc<Vec<u8>>` instead of `Arc<Vec<u8>>`), but it is more powerful because it will work across all types that are `AsRef<[T]>`.

The implementation in this PR opts for bespoke `struct` extractors for each of the typed array types (e.g., `Uint8Array`, `Int8Array`). The benefits of this approach are the simplicity of implementation and the ability to directly use them as destructor patterns and constructors. The most significant downside is that there's no trait to unify them for generic buffer functions.

If we wanted to include a trait for handling these in the future, we could introduce a `TypedArray` extractor and accept that this implementation is distinct from the one in this PR. No breaking change would be necessary.

## Example Usage

```rust
use bytes::BytesMut;
use neon::types::extract::Uint8Array;

#[neon::export]
pub fn concat_bytes(
    Uint8Array(mut buf): Uint8Array<BytesMut>,
    data: Vec<u8>,
) -> Uint8Array<BytesMut> {
    buf.put_slice(&data);

    Uint8Array(buf)
}
```

### DIFF
--- a/crates/neon/src/types_impl/extract/buffer.rs
+++ b/crates/neon/src/types_impl/extract/buffer.rs
@@ -1,0 +1,221 @@
+use crate::{
+    context::Cx,
+    handle::Handle,
+    result::{JsResult, NeonResult},
+    types::{
+        buffer::{Binary, TypedArray},
+        extract::{private, TryFromJs, TryIntoJs, TypeExpected},
+        JsArrayBuffer, JsBigInt64Array, JsBigUint64Array, JsBuffer, JsFloat32Array, JsFloat64Array,
+        JsInt16Array, JsInt32Array, JsInt8Array, JsTypedArray, JsUint16Array, JsUint32Array,
+        JsUint8Array, JsValue, Value,
+    },
+};
+
+/// Wrapper for converting between bytes and [`JsArrayBuffer`](JsArrayBuffer)
+pub struct ArrayBuffer<B>(pub B);
+
+impl<'cx, B> TryFromJs<'cx> for ArrayBuffer<B>
+where
+    for<'b> B: From<&'b [u8]>,
+{
+    type Error = TypeExpected<JsBuffer>;
+
+    fn try_from_js(
+        cx: &mut Cx<'cx>,
+        v: Handle<'cx, JsValue>,
+    ) -> NeonResult<Result<Self, Self::Error>> {
+        let v = match v.downcast::<JsArrayBuffer, _>(cx) {
+            Ok(v) => v,
+            Err(_) => return Ok(Err(Self::Error::new())),
+        };
+
+        Ok(Ok(ArrayBuffer(B::from(v.as_slice(cx)))))
+    }
+}
+
+impl<'cx, B> TryIntoJs<'cx> for ArrayBuffer<B>
+where
+    B: AsRef<[u8]>,
+{
+    type Value = JsArrayBuffer;
+
+    fn try_into_js(self, cx: &mut Cx<'cx>) -> JsResult<'cx, Self::Value> {
+        JsArrayBuffer::from_slice(cx, self.0.as_ref())
+    }
+}
+
+impl<B> private::Sealed for ArrayBuffer<B> {}
+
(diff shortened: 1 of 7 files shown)


crates/neon/src/types_impl/extract/buffer.rs
```diff
@@ -0,0 +1,240 @@
...
+impl<T> private::Sealed for &[T]
+where
+    JsTypedArray<T>: Value,
+    T: Binary,
+{
+}
+
+macro_rules! typed_array {
+    ($js:ident, $name:ident, $type:ty) => {
+        #[doc = concat!(
```
dherman, line 182: Such attention to detail! Thanks for making the effort to make the docs good.

crates/neon/src/types_impl/extract/buffer.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl<T, const N: usize> private::Sealed for [T; N]
+where
+    JsTypedArray<T>: Value,
+    T: Binary,
+{
+}
+
+impl<'cx, T> TryIntoJs<'cx> for &[T]
```
dherman, line 161: This may be a corner case but should we also think about impls for `std::cell:Ref` as well? It's very similar to `&` but comes up when you dynamically borrow from a standard cell, and if we don't implement it the orphan rule will prevent anyone else from being able to.

dherman (approved):
I love this design, it's a huge ergonomics win. Let's add just a couple unit tests to have some examples in repo.

crates/neon/src/types_impl/extract/buffer.rs
dherman: Oh never mind, that's covered in the wrappers via `AsRef`.

crates/neon/src/types_impl/extract/buffer.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl<T> private::Sealed for Vec<T>
+where
+    JsTypedArray<T>: Value,
+    T: Binary,
+{
+}
+
+impl<'cx, T> TryIntoJs<'cx> for Box<[T]>
```
dherman, line 123: (Just discussed in person: let's drop the Box special cases here. Boxed slice is not as common as Vec and &[].)

test/napi/src/js/extract.rs
```diff
@@ -22,9 +22,9 @@ pub fn extract_values(mut cx: FunctionContext) -> JsResult<JsArray> {
         String,
         Date,
         Handle<JsValue>,
-        ArrayBuffer,
+        ArrayBuffer<Vec<u8>>,
```
kjvalencik, line 25: Need a test or two for the typed array extractors. We could use `String` or `Cow` as an example that couldn't otherwise use the shorthand. A more common example is `bytes::Bytes`, but that would require an additional dev dependency.

crates/neon/src/types_impl/extract/buffer.rs
kjvalencik: Yep! It will work with anything that is `AsRef<[T]>` for `TryIntoJs` and `for<'a> From<&'a [T]>` for `TryFromJs`. This covers `Ref` as well as other useful types like `bytes::Bytes`.

Keep in mind that in _many_ cases you wouldn't actually be able to use `Ref` directly because of the lifetime, but there are some niche scenarios where it would work.

crates/neon/src/types_impl/extract/buffer.rs
kjvalencik: Removed.

test/napi/src/js/extract.rs
kjvalencik: Added.

PR #1091, opened 2025-02-01
Type extractors for common container types

dherman:
Implement type extractors for common container types in the standard library, so that they don't required explicit `Boxed` wrapping with `JsBox` in userland.

Implemented so far:
- `RefCell`
  - `impl TryIntoJs for RefCell`
  - `impl TryFromJs for &RefCell`
  - `impl TryFromJs for Ref`
  - `impl TryFromJs for RefMut`
- `Rc`
  - `impl TryIntoJs for Rc`
  - `impl TryFromJs for Rc`
- `Arc`
  - `impl TryIntoJs for Arc`
  - `impl TryFromJs for Arc`

### DIFF
--- a/crates/neon/src/types_impl/extract/container.rs
+++ b/crates/neon/src/types_impl/extract/container.rs
@@ -1,0 +1,42 @@
+use std::{any::Any, cell::RefCell};
+
+use crate::{
+    context::Cx, handle::Handle, result::{JsResult, NeonResult}, types::{extract::{TryFromJs, TryIntoJs}, JsBox, JsValue}
+};
+
+use super::error::RustTypeExpected;
+
+pub trait Container {
+    fn container_name() -> &'static str;
+}
+
+impl<T> Container for RefCell<T> {
+    fn container_name() -> &'static str {
+        "std::cell::RefCell"
+    }
+}
+
+impl<'cx, T: Container + 'static> TryFromJs<'cx> for &'cx RefCell<T> {
+    type Error = RustTypeExpected<RefCell<T>>;
+
+    fn try_from_js(
+        cx: &mut Cx<'cx>,
+        v: Handle<'cx, JsValue>,
+    ) -> NeonResult<Result<Self, Self::Error>> {
+        match v.downcast::<JsBox<RefCell<T>>, _>(cx) {
+            Ok(v) => Ok(Ok(JsBox::deref(&v))),
+            Err(_) => Ok(Err(RustTypeExpected::new())),
+        }
+    }
+}
+
+impl<'cx, T> TryIntoJs<'cx> for RefCell<T>
+where
+    T: Any + 'static,
+{
+    type Value = JsBox<RefCell<T>>;
+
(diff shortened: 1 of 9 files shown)


crates/neon/src/types_impl/extract/container.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<T> Container for Arc<T> {
+    fn container_name() -> &'static str {
+        "std::sync::Arc"
+    }
+}
+
+impl<'cx, T: 'static> TryFromJs<'cx> for &'cx RefCell<T> {
+    type Error = RustTypeExpected<RefCell<T>>;
```
kjvalencik, line 42: What do you think of using `TypeExpected<JsBox<T>>` to reduce the API surface area a bit? Since these types are still `JsBox` under the hood, I think it's okay to expose the error as that.

dherman: That may be a good idea also so people can understand what the actual runtime representation is. (I think it would be `TypeExpected<JsBox<RefCell<T>>>`, right?)

dherman: @kjvalencik A related design call I'd love your input on: right now the dynamic borrow errors are explicitly handled with `try_borrow` and `try_borrow_mut` and turned into a Neon-specific error type called `RefCellError`. An alternative is we could just let them panic and keep the API surface to just `TypeExpected`. Do you have a preference?

dherman: I also dislike the fact that `RefCellError::WrongType` overlaps with `TypeExpected`.

dherman: Yeah I really don't see much tangible benefit to the extra types. The panic message is basically the same. I'll remove it for now and if you disagree we can talk about putting the type back.

kjvalencik: Yes, `TypeExpected<JsBox<RefCell<T>>>` would be the type I would expect. I don't like letting it panic. We'll catch the unwind and convert it to an exception, but it might not be clear.

I think it's okay to let throw an exception. If you don't want this behavior, the author can go to `&'cx RefCell` first and then borrow. If they are going directly to `Ref` or `RefMut`, they are already opting into some shorthand syntax.

However, I think it's better for the `try_from_js` to call the `try_` methods on `RefCell` and on error, manually throw with `cx.throw_error(..)` and a clear error message.

dherman: Oh I think I see what you're suggesting: instead of creating a new Error type, just explicitly throw an error with a custom message directly in the `TryFromJs` implementation. That's reasonable. I'm not sure it makes a huge difference but I'm OK with it.

dherman: OK I pushed this middle ground -- it uses `try_borrow()` and `try_borrow_mut()` but explicitly calls `cx.throw_error()` instead of creating a custom type, so it can use `TypeExpected` for the other case. This feels like a good choice to me but see what you think.

kjvalencik (approved):

PR #1096, opened 2025-03-24
breaking: Remove the `FromArgs` impl on `T: TryFromJs` and add `cx.arg()` and `cx.arg_opt()`.

kjvalencik:
A blanket impl was added to avoid duplicate methods or a confusing `let (n,) = cx.args()?;` syntax for unary functions. However, this does not work well with the `#[neon::export]` macro where functions with zero arguments end up unnecessarily deserializing `undefined`.

Removing the blanket `impl` and replacing it with a specialized version for `()` allows the proc macros to be more efficient without unneeded complexity.

In order to make these changes more ergonomic, I added `Context::arg` and `Context::arg_opt` for unary functions.

**Attention**: This is a _breaking_ change, but the API has not landed in a stable version yet.

In the future, we could unify these APIs if Rust ships specialization.

### DIFF
--- a/crates/neon/src/types_impl/extract/mod.rs
+++ b/crates/neon/src/types_impl/extract/mod.rs
@@ -183,27 +183,19 @@ pub struct Date(pub f64);
 /// the sake of brevity, only tuples up to size 8 are shown in this documentation.
 pub trait FromArgs<'cx>: private::FromArgsInternal<'cx> {}
 
-// Convenience implementation for single arguments instead of needing a single element tuple
-impl<'cx, T> private::FromArgsInternal<'cx> for T
-where
-    T: TryFromJs<'cx>,
-{
-    fn from_args(cx: &mut FunctionContext<'cx>) -> NeonResult<Self> {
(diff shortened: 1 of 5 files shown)


crates/neon/src/types_impl/extract/mod.rs
```diff
@@ -184,26 +184,17 @@ pub struct Date(pub f64);
 pub trait FromArgs<'cx>: private::FromArgsInternal<'cx> {}
 
 // Convenience implementation for single arguments instead of needing a single element tuple
```
dherman, line 186: Let's update this comment to explain that this serves as the base case for `from_args!`, and as an optimization avoids the unnecessary Node-API calls of the n > 0 cases.

dherman (approved):
This is arguably a better API since it avoids the ambiguity of `let () = cx.args()?;` being either the nullary or unary case, and `let x = cx.arg()?;` feels like a reasonable way to explicitly do the unary case and avoid Rust's awkward 1-tuple syntax.

I just left one suggestion to clarify the comment next to the base case impl.

PR #1105, opened 2025-05-23
Fix pre-napi-5 build

jrose-signal:
Regressions in the 1.1 release, fortunately easy to fix!

```
cargo check --no-default-features -p neon
```


kjvalencik:
@jrose-signal can you share the project that broke with these changes? I tested `libsignal` prior to release to double check that it still compiled, but I might have been missing a flag.

If possible, I would like to prevent us from creating similar regressions in the future. Thanks!

PR #1106, opened 2025-05-23
Fix pre-napi-5 build

kjvalencik:
Includes https://github.com/neon-bindings/neon/pull/1105 and some additional fixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,7 +376,7 @@ dependencies = [
 
 [[package]]
 name = "neon"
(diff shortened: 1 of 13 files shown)


kjvalencik:
FYI, the coverage is just from a change in feature flag and not real code. After merging, the CI failure should go away.

kjvalencik:
FYI, I disabled code coverage status checks because they were the defaults and causing failures for weird reasons (it thinks my cfg flag is a code change that has zero coverage).

PR #1111, opened 2025-07-16
feat(neon): Export LazyLock globals

kjvalencik:
This changes the macro to call `try_into_js` on a reference and leverage autoref specialization and autoderef to make it work. Since the macro only works on `static` and `const`, it already only worked on `Copy` types which continues to work.

This makes it more flexible so that you can also export types where their reference implements `TryIntoJs`. I also added a `TryIntoJs` implementation for `LazyLock<T>` where `&T: TryIntoJs`. It only makes sense to implement it for a reference since getting an owned value would require clone.

The combination of these features allows you to export things like `LazyLock<Vec<u8>>`.


test/napi/lib/export.js
```diff
@@ -18,6 +18,14 @@ function globals() {
     assert.deepStrictEqual(addon.MESSAGES, ["hello", "neon"]);
     assert.deepStrictEqual(addon.renamedMessages, ["hello", "neon"]);
   });
+
+  it("LazyLock", () => {
+    // Test that LazyLock<String> is exported as a string
+    assert.strictEqual(typeof addon.LAZY_LOCK_HELLO, "string");
```
kjvalencik, line 24: This test was vibe coded with Copilot in Agent mode. This assertion is redundant with the next assertion.

PR #1113, opened 2025-07-22
fix(neon): Fix test on Rust nightly

kjvalencik:
The error message in Rust changed from `already borrowed: BorrowMutError` to `RefCell already borrowed`.


dherman (approved):
lol, LGTM

PR #1117, opened 2025-09-02
[WIP] Class macro

dherman:
Example:

```rust
#[derive(Debug, Clone)]
pub struct Point {
    x: u32,
    y: u32,
}

#[neon::class]
impl Point {
    pub fn new(x: u32, y: u32) -> Self {
        Self { x, y }
    }

    pub fn x(&self) -> u32 {
        self.x
    }

    pub fn y(&self) -> u32 {
        self.y
    }

    pub fn distance(&self, other: Instance<Self>) -> f64 {
        let dx = (self.x as i32 - other.x() as i32).pow(2);
        let dy = (self.y as i32 - other.y() as i32).pow(2);
        ((dx + dy) as f64).sqrt()
    }
}

impl Finalize for Point {
    fn finalize<'cx, C: Context<'cx>>(self, _cx: &mut C) { }
}
```


dherman:
Notes
- Move the `Finalize` implementation into the syntax and offer a default implementation.
- No need to make the context argument optional, since it's really the only point of `Finalize` as opposed to `Drop`
- No need for a constructor renaming attribute or finalize renaming attribute to allow multiple classes, since newtype should be sufficient for now, but we could add this compatibly in the future.

crates/neon/src/types_impl/extract/error.rs
```diff
@@ -11,7 +11,42 @@ use crate::{
 
 type BoxError = Box<dyn error::Error + Send + Sync + 'static>;
 
-/// Error returned when a JavaScript value is not the type expected
+/// Error returned when a JavaScript value is not an instance of the expected class.
```
kjvalencik: This isn't accurate. If it's not the expected class, it's a panic. This is only the error if it's not an object at all.

PR #1127, opened 2025-12-12
fix(neon): class methods have proper `.name` property

dherman:
Build class prototype methods with `JsFunction::with_name()` so that they display more idiomatically with JavaScript reflection operations like `.name` and `util.inspect()`.


test/napi/lib/class.js
```diff
@@ -44,6 +44,46 @@ describe("classes", function () {
     );
   });
 
+  it("class methods have normal-looking function names", function () {
+    const util = require("util");
+    const Message = addon.Message;
+    const message = new Message("test");
+
+    assert.strictEqual(message.read.name, "read");
```
kjvalencik: Can we add a test for something that should be camel cased?

kjvalencik (approved):

PR #1139, opened 2026-05-11
fix(neon): (soundness) prevent handle leaks from execute_scoped / compute_scoped

kjvalencik:
## Summary

`execute_scoped` / `compute_scoped` let the caller pick the inner lifetime
`'b`, including `'b = 'a`, so a closure could `leaked = Some(inner.number(42))`
into outer state and dereference it after the scope dropped. UB.

Fix: HRTB on the closure plus a new `ScopedCx<'outer, 'inner>` whose
`PhantomData<&'inner &'outer ()>` carries the implied `'outer: 'inner`
bound. HRTB universalizes `'b` so nothing parameterized by it escapes;
the phantom bound keeps `recompute_scoped`-style pass-through compiling
via `Handle` covariance.

`ScopedCx::new` is `unsafe fn` with a documented contract — `Cx::new` is
deliberately untouched.

## Breaking

Closure parameter is now `ScopedCx<'a, 'b>`, not `Cx<'b>`. `ExecuteContext`
/ `ComputeContext` aliases gain a second lifetime. Most callers compile
unchanged via `Deref<Target = Cx<'inner>>`.

## Tests

- `fail/execute-scoped-leak.rs`, `fail/compute-scoped-leak.rs` — leaks via
  captured `&mut` no longer compile (E0521).
- `fail/scoped-cx-escape.rs` — `ScopedCx` itself can't be returned.
- `pass/scoped-cx.rs` — outer-handle pass-through, fresh-handle escape,
  and using outer handles inside the scope all still compile.
- `napi-tests` builds, exercising `recompute_scoped` at runtime.

### DIFF
--- a/crates/neon/src/context/mod.rs
+++ b/crates/neon/src/context/mod.rs
@@ -204,13 +204,11 @@ use crate::lifecycle::InstanceData;
 /// An execution context of a task completion callback.
 pub type TaskContext<'cx> = Cx<'cx>;
 
-#[doc(hidden)]
 /// An execution context of a scope created by [`Context::execute_scoped()`](Context::execute_scoped).
-pub type ExecuteContext<'cx> = Cx<'cx>;
+pub type ExecuteContext<'outer, 'inner> = ScopedCx<'outer, 'inner>;
 
-#[doc(hidden)]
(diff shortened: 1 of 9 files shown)


dherman (approved):
What a good catch, thank you. This PR looks great. Thank you for the very clear comments and good unit tests.

